PR #53, opened 2026-01-07
feat(apollo-react): add edge toolbar [MST-5311]

BenGSchulz:
- Edge updates
 - Move ExecutionStatusContext to shared hooks to also consume it for edges.
 - Add ValidationStatusContext to function similarly.
 - Extract reusable elements from NodeToolbar and add new EdgeToolbar all under new Toolbar folder.
 - Add EdgeToolbar related hooks to handle positioning by following mouse along path.
 - Bring over loop path logic and add into SequenceEdge.
   - Still missing loop-specific edge maintenance logic that exists in Maestro.
 - Update AddNodeManager logic to allow splitting existing edge with a new node.
   - Still missing repositioning logic when there is no room for new node.

https://github.com/user-attachments/assets/e711f4a5-9112-4a79-ac9d-3439ab645ee9

Screenshot 2025-12-19 at 8 46 52 PM


packages/apollo-react/src/canvas/components/AddNodePanel/AddNodeManager.tsx
```diff
@@ -61,31 +62,60 @@ export const AddNodeManager: React.FC<AddNodeManagerProps> = ({
...
+              ? previewEdge.targetHandle || 'input'
+              : previewEdge.sourceHandle || 'output',
+            previewEdgeId: previewEdge.id,
+          };
         });
+        setConnectionInfo(connections);
         setIsOpen(true);
+        restoreEdgesRef.current = previewNode.data.originalEdge
+          ? [previewNode.data.originalEdge as Edge]
+          : null;
```
rahul-sharma-uipath: this logic is becoming a bit unreadable -- can you extract this into a function?

packages/apollo-react/src/canvas/components/Toolbar/EdgeToolbar/useEdgeToolbarPositioning.ts
```diff
@@ -0,0 +1,219 @@
...
+  pathElementRef,
+  isEnabled,
+  targetPosition,
+}: UseEdgeToolbarPositioningProps): EdgeToolbarPositioning {
+  const reactFlow = useReactFlow();
+  const [positionData, setPositionData] = useState<EdgeToolbarPositionData | null>(null);
+  const rafIdRef = useRef<number | null>(null);
+
+  // Reset position and cleanup when disabled
+  useEffect(() => {
```
rahul-sharma-uipath, line 143: nit: think about how we might move this out of an effect

packages/apollo-react/src/canvas/hooks/ExecutionStatusContext.tsx
```diff
@@ -0,0 +1,38 @@
...
+  edgeId: string,
+  targetNodeId: string
+): ExecutionState | undefined => {
+  const context = useContext(ExecutionStatusContext);
+  const [state, setState] = useState<ExecutionState | undefined>();
+
+  useEffect(() => {
+    const executionState = context.getEdgeExecutionState(edgeId, targetNodeId);
+    setState(executionState);
+  }, [edgeId, targetNodeId, context]);
```
rahul-sharma-uipath, line 35: similar nit here -- ideally we can find a way to do this without an effect

rahul-sharma-uipath (approved):
minor nits -- otherwise lgtm

PR #57, opened 2026-01-07
feat(apollo-react): extract task icon types as component [MST-5923]

SreedharAvvari:
## Description
Introduced new component `TaskIcon` which accepts task type & size as props.
**Light mode:**
&nbsp;
image
&nbsp;
image

&nbsp;
**Dark mode:**
&nbsp;
image
&nbsp;
image


packages/apollo-react/src/canvas/icons/HumanIcon.tsx
```diff
@@ -0,0 +1,8 @@
+export const HumanIcon = ({ w = 24, h = 24 }: { w?: number | string; h?: number | string }) => (
```
kittyyueli, line 1: take a look at how we are doing icons in PO.Frontend :)

packages/apollo-react/src/canvas/components/TaskIcon/TaskIcon.tsx
```diff
@@ -0,0 +1,39 @@
...
+    case TaskItemTypeValues.ApiAutomation:
+      return <ApiProject w={iconSize} h={iconSize} />;
+    case TaskItemTypeValues.User:
+      return <HumanIcon w={iconSize} h={iconSize} />;
+    case TaskItemTypeValues.AgenticProcess:
+      return <BusinessProcessProject w={iconSize} h={iconSize} />;
+  }
+};
+
+export const TaskIcon = ({ type, size = 'md' }: TaskIconProps) => {
```
kittyyueli, line 30: can we split the props for an icon size and container size? I think there is a case where the background is larger than the icon inside it~ For example the icons on the task vs the toolbox has different ratio of icon size to background size: 

Image

maybe also have a circle version as well since I see it in the Figma with the gradients

kittyyueli (commented):
Also consume in the stage node :) thanks!! This looks really close~

packages/apollo-react/src/canvas/components/TaskIcon/TaskIcon.tsx
SreedharAvvari: I thought consumption will be cleaner if we have size variants (small | medium | large). Otherwise, we have to pass the sizes at each place we consume the icons.
The above 2 cases are already covered
- icons in stage task node - small variant with 24px container, 16px icon
- icons in toolbox - medium variant with 32px container, 24px icon

Let me know if we need something else here.

> maybe also have a circle version as well since I see it in the Figma with the gradients


Unable to find the circular version 🤔 is it this one
image

SreedharAvvari:
> Also consume in the stage node :) thanks!! This looks really close~

Ah, missed this one. Will make the changes. Thanks

packages/apollo-react/src/canvas/icons/HumanIcon.tsx
SreedharAvvari: I have followed the arrow function convention from other icons in `packages/apollo-react/src/canvas/icons` and PO.frontend has icons defined as function declarations. Not sure which is preferred 🤔 

`packages/apollo-react/src/canvas/icons`
&nbsp;
image

`PO.frontend`
&nbsp;
image

SreedharAvvari:
> Also consume in the stage node :) thanks!! This looks really close~

Checked this and found out stageNode is getting icons as prop which is passed from PO.frontend. So I think I might have to make the changes in PO.frontend once this gets merged.

```
interface StageTaskItem {
    id: string;
    label: string;
    icon?: React.ReactElement;
}
```

kittyyueli (approved):

PR #59, opened 2026-01-08
feat(apollo-vertex): persist shell sidebar state in local storage [AGVSOL-793]

alincadariu:


apps/apollo-vertex/registry/use-local-storage/use-local-storage.ts
```diff
@@ -0,0 +1,42 @@
+import { useCallback, useSyncExternalStore } from "react";
+
+export const SAME_CONTEXT_STORAGE_EVENT_TYPE = "same-context-storage-event-type";
+
+export function useLocalStorage<T>(
```
ruudandriessen, line 5: This doesn't work - you only made it work in our app - but the useLocalStorage hook isn't part of the registry, so if anyone actually uses the sidebar, it'll get a runtime error

ruudandriessen (changes requested):
Idea: As a follow up - make sure we disallow anything from the registry to use code outside of it 👀

apps/apollo-vertex/registry/use-local-storage/use-local-storage.ts
alincadariu: @ruudandriessen oops, moved it into registry. created ticket https://uipath.atlassian.net/browse/AGVSOL-838 for disallow follow up

ruudandriessen, line 5: shouldn't we also add this as a dependency of the sidebar component?

ruudandriessen (changes requested):

apps/apollo-vertex/registry/use-local-storage/use-local-storage.ts
alincadariu: @ruudandriessen we should, and we should also add this hook in the registry as a `registry:hook`. pushed the fixup

ruudandriessen (approved):

PR #66, opened 2026-01-13
fix(apollo-react): Updated StageNode Task functionality [MST-5922]

ayushya:
Demo:

https://github.com/user-attachments/assets/ba1d8cfe-5088-40cc-8152-a220f7afeb0c

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/DraggableTask.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/DraggableTask.tsx
@@ -4,14 +4,12 @@ import { FontVariantToken, Padding, Spacing } from '@uipath/apollo-core';
 import { Column, Row } from '@uipath/apollo-react/canvas/layouts';
 import {
   ApBadge,
-  ApIcon,
   ApTooltip,
   ApTypography,
   BadgeSize,
   type StatusTypes,
 } from '@uipath/apollo-react/material';
-import { memo, useCallback, useMemo } from 'react';
-
+import { memo, useCallback, useMemo, useState } from 'react';
 import { ExecutionStatusIcon } from '../ExecutionStatusIcon';
 import type { DraggableTaskProps, TaskContentProps } from './DraggableTask.types';
 import {
@@ -20,12 +18,12 @@ import {
   StageTaskDragPlaceholder,
   StageTaskDragPlaceholderWrapper,
   StageTaskIcon,
-  StageTaskRemoveButton,
   StageTaskRetryDuration,
   StageTaskWrapper,
 } from './StageNode.styles';
 import type { StageTaskExecution } from './StageNode.types';
 import { TaskContextMenu } from './TaskContextMenu';
+import { TaskMenu } from './TaskMenu';
 
 const ProcessNodeIcon = () => (
   <svg viewBox="0 0 24 24" fill="none" stroke="currentColor" strokeWidth="1.5">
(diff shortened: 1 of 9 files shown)


packages/apollo-react/src/canvas/components/StageNode/DraggableTask.tsx
```diff
@@ -123,18 +129,54 @@ const DraggableTaskComponent = ({
...
       onTaskClick(e, task.id);
     },
-    [onTaskClick, task.id]
+    [isMenuOpen, isContextMenuVisible, handleMenuClose, onTaskClick, task.id]
+  );
+
+  const handleMenuClick = useCallback((e: React.MouseEvent) => {
+    e.stopPropagation();
+    e.preventDefault();
+    setIsMenuOpen(true);
```
kittyyueli: nit: 
```suggestion
    setIsMenuOpen(open => !open);
```

packages/apollo-react/src/canvas/components/StageNode/DraggableTask.tsx
```diff
@@ -182,19 +224,45 @@ const DraggableTaskComponent = ({
...
+            menuItems={transformedMenuItems}
+            anchorEl={menuAnchorRef.current}
+            onClose={handleMenuClose}
+            anchorOrigin={{
+              vertical: 'bottom',
+              horizontal: 'right',
+            }}
+            transformOrigin={{
+              vertical: 'top',
+              horizontal: 'right',
```
kittyyueli: You can move this into its own component as well with the event handlers, the ref and the open state

packages/apollo-react/src/canvas/components/StageNode/transformMenuItems.ts
kittyyueli, line 1: can you move this into the StageNodeTaskUtilities file?

kittyyueli: and add some tests~ thanks!

packages/apollo-react/src/canvas/components/StageNode/StageNode.tsx
```diff
@@ -243,34 +243,24 @@ const StageNodeComponent = (props: StageNodeProps) => {
...
+      taskIndex: number,
       tasksLength: number,
       taskGroupLength: number,
       isAboveParallel: boolean,
       isBelowParallel: boolean
     ) =>
       getContextMenuItems(
-        taskStateReference.isParallel,
-        taskStateReference.groupIndex,
+        isParallel,
```
kittyyueli, line 263: do you know why it was using the taskStateRef before vs using the actual state itself?

kittyyueli:
lets have the 3 dots menu be visible if the menu is open :)

packages/apollo-react/src/canvas/components/StageNode/StageNode.tsx
ayushya: The change from taskStateRef to passing parameters directly improves the code in a few ways:

Avoids stale closures: The old approach stored task state in a ref that was updated when the context menu opened. The new approach passes the actual parameters when generating menu items, ensuring we always have the correct values.

More explicit and clearer: You can see exactly which task's context menu is being generated by looking at the call site, rather than having to trace back to where taskStateReference was set.

Reduces state complexity: One less piece of state to manage and keep synchronized.

Better with the new menu architecture: Since we now have both a context menu (right-click) and an inline menu button (3-dots), passing parameters directly makes it easier to generate the correct menu items for each task without shared state.

The old pattern was likely used when there was only a single context menu that needed to track which task triggered it. With the refactored approach, each task component can independently generate its menu items.

PR #67, opened 2026-01-14
chore(repo): add yalc temp linking fix [MST-6092]

CalinaCristian:
Adds [yalc](https://github.com/wclr/yalc) as a temp solution 

## DEMO

https://github.com/user-attachments/assets/ba841fa0-13d5-4b3d-91e8-47e2b02b4e28


README.md
```diff
@@ -195,6 +240,62 @@ Please read [CLAUDE.md](./CLAUDE.md) for detailed architecture documentation and
...
 
+## 🔗 Local Development Linking
+
+Use these commands to link Apollo packages for local development in external projects.
+
+### Quick Start
+
+**Install yalc globally (one-time setup):**
+```bash
+npm install -g yalc
```
0xr3ngar, line 251: wait why do I have to download this globally? What if I don't want to? We should add it in the root package.json as a dependency and use it only in here.

turbo.json
```diff
@@ -7,9 +7,12 @@
       "outputs": ["dist/**", ".next/**", "build/**", "storybook-static/**"]
     },
     "dev": {
-      "dependsOn": ["^build"],
```
0xr3ngar, line 10: why remove the build in the dev?

0xr3ngar (changes requested):
why do we need this yalc? can't we just use pnpm link? directly?

PR #73, opened 2026-01-16
fix(apollo-react): stage icons alignment [6185]

ayush578:
alignment of status icon with stage title
removed sla/escalation icons
remove status text and provide it as a tooltip to status icon.

before

https://github.com/user-attachments/assets/32fb581a-aaa8-444c-8406-35cf66718881

after

https://github.com/user-attachments/assets/c1891a02-3e60-49af-99bd-dab08b64e42a


kittyyueli (commented):
The timer icon and the escalation icon should be removed

kittyyueli (approved):

PR #79, opened 2026-01-16
feat(apollo-react): migrated ap-accordion and ApButton accept custom Styles PLT-94897

Miyuki-L:
https://github.com/user-attachments/assets/08d12ae6-94fa-4bfc-94e5-76345bb1e854

- Migrated ap-accordion from apollo-design-system
- Removed some of the props used to provide custom stylings as consumers can style them now a restriction from web components. 

- Kept Label props for Font Weight and Color - for easier stylings as they are hard to reach with style and sx props 
- Added props to position the left and right endings of the divider as sx Props passed to summary would not work
- Added disableDivider prop to disable divider 
    - Mui also has the problem where if its previous sibiling is a `div` and the Accordion is the first accordion it doesn't register it as a first and shows the divider.  
 - removed `expanded` prop as it had the same purpose as `defaultExpanded` and its purpose was more `defaultExpanded` prop instead

- ApButton switch to use tokens and also accept custom styles

### DIFF
--- a/packages/apollo-react/src/material/components/ap-accordion/ApAccordion.tsx
+++ b/packages/apollo-react/src/material/components/ap-accordion/ApAccordion.tsx
@@ -1,0 +1,149 @@
+import Accordion from '@mui/material/Accordion';
+import AccordionDetails from '@mui/material/AccordionDetails';
+import AccordionSummary from '@mui/material/AccordionSummary';
+import { styled } from '@mui/material/styles';
+import { FontVariantToken, PadL } from '@uipath/apollo-core';
+import token from '@uipath/apollo-core';
+import { ChevronDown } from '@uipath/apollo-react/icons';
+import { ApTypography } from '../ap-typography';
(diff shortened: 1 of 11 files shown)


CalinaCristian (dismissed):

packages/apollo-react/src/material/components/ap-accordion/ApAccordion.tsx
```diff
@@ -0,0 +1,150 @@
...
+    dividerLeftPosition,
+    dividerRightPosition,
+    detailsSx,
+    detailsProps,
+
+    ...accordionProps
+  } = props;
+
+  return (
+    <StyledAccordion defaultExpanded={expanded || defaultExpanded} {...accordionProps}>
```
CalinaCristian: Ai suggestion seems correct here, they do the same thing

CalinaCristian (approved):

PR #83, opened 2026-01-19
chore: formatting fixes

ruudandriessen:
Properly run biome based on the workspace configuration.

Run the formatter on the registry, such that the consuming repo's don't get as many deltas if they do want to run the formatter on all files as well


apps/apollo-vertex/biome.json
```diff
@@ -1,6 +1,6 @@
 {
-    "root": false,
-    "$schema": "https://biomejs.dev/schemas/2.3.8/schema.json",
```
0xr3ngar, line 3: update your local biome binary 🥺 🙏  so we don't have to bump down

0xr3ngar (changes requested):

apps/apollo-vertex/biome.json
ruudandriessen: pnpm dedupes package versions and the root one is 2.3.6

ruudandriessen: So we'd have to update all of the repo

0xr3ngar (approved):

PR #85, opened 2026-01-20
feat(apollo-react): add isreadOnly attribute in stage node props [MST-6193]

ayush578:
Prior logic affects the handling of stage handles in a wrong way.
So I created a isReadOnly attribute to hide the handles in a right way.
Also used this variable wherever needed.

https://github.com/user-attachments/assets/9da9c2ea-b2e5-432b-9719-e2ef340d6942

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNode.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNode.tsx
@@ -84,6 +84,7 @@ const StageNodeComponent = (props: StageNodeProps) => {
   const taskIds = useMemo(() => flatTasks.map((task) => task.id), [flatTasks]);
 
   const isException = stageDetails?.isException;
+  const isReadOnly = !!stageDetails?.isReadOnly;
   const icon = stageDetails?.icon;
   const selectedTasks = stageDetails?.selectedTasks;
   const defaultContent = stageDetails?.defaultContent || 'Add first task';
@@ -96,7 +97,7 @@ const StageNodeComponent = (props: StageNodeProps) => {
     [onTaskGroupModification]
   );
 
-  const isStageTitleEditable = !!onStageTitleChange;
+  const isStageTitleEditable = !!onStageTitleChange && !isReadOnly;
 
   const [isHovered, setIsHovered] = useState(false);
   const [label, setLabel] = useState(props.stageDetails.label);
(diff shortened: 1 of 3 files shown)


packages/apollo-react/src/canvas/components/StageNode/StageNode.tsx
```diff
@@ -84,6 +84,7 @@ const StageNodeComponent = (props: StageNodeProps) => {
   const taskIds = useMemo(() => flatTasks.map((task) => task.id), [flatTasks]);
 
   const isException = stageDetails?.isException;
+  const isExecution = stageDetails?.isExecution || false;
```
ayushya: Is this different than readonly? Can we use it?

ayushya:
The PR title mentions isDesigning, but probably was renamed. Please update the PR title

packages/apollo-react/src/canvas/components/StageNode/StageNode.tsx
SreedharAvvari: I agree, the naming can be more generic (~readonly) since we want to hide handles in other scenarios as well apart from execution

ayush578: As far i know, It is not different from readOnly.
Its just one usage is bcoz of we are in execution.
Alignment of status icon is dependent of whether we are in execution mode or design mode

ayushya: Lets use it (isReadonly/readonly) whichever seems more appropriate

packages/apollo-react/src/canvas/components/StageNode/StageNode.tsx
```diff
@@ -96,7 +97,7 @@ const StageNodeComponent = (props: StageNodeProps) => {
     [onTaskGroupModification]
   );
 
-  const isStageTitleEditable = !!onStageTitleChange;
+  const isStageTitleEditable = !!onStageTitleChange  && !isExecution;
```
ayushya: We have Extra spacing here, this should have been handled by the linter. Can you check what happened here? 
```suggestion
  const isStageTitleEditable = !!onStageTitleChange && !isExecution;
```

packages/apollo-react/src/canvas/components/StageNode/StageNode.tsx
ayush578: done

packages/apollo-react/src/canvas/components/StageNode/StageNode.tsx
```diff
@@ -84,6 +84,7 @@ const StageNodeComponent = (props: StageNodeProps) => {
   const taskIds = useMemo(() => flatTasks.map((task) => task.id), [flatTasks]);
 
   const isException = stageDetails?.isException;
+  const isReadOnly = stageDetails?.isReadOnly || false;
```
uipath-utkarsh: wont it be `false` already if not true?  or are we looking for `undefined` / `null`?


maybe a better way is `!!stageDetails?.isReadOnly`

ayush578: okay

kittyyueli (approved):

PR #94, opened 2026-01-21
feat(apollo-react): add category nesting and constraint filtering for node options [MST-6070]

BenGSchulz:
- Adds category nesting via `parentId` in CategoryManifest.
- Updates NodeTypeRegistry to work off of a pre-built caches so accessing node/category manifests and particular nested properties like handles/constraints is quicker/easier. It will also add a tree structure to be able to work with for displaying categorized nodes in the add node panel and support operations like filtering and flattening.
- Migrates the constraint-validator from flow-workbench and use it for filtering options for add node panel based on connection constraints.
- Moves the registry and supporting files into new core directory because it felt out-of-place in the BaseNode directory.

https://github.com/user-attachments/assets/8f5e80f8-a195-44d0-9ef0-b8194a56b8a1

https://github.com/user-attachments/assets/d1b59ba9-069f-407b-a0e7-4526785670eb


packages/apollo-react/src/canvas/components/AddNodePanel/AddNodePanel.tsx
```diff
@@ -12,23 +20,29 @@ export function AddNodePanel({
   items,
   title,
   loading,
+  previewNodeConnectionInfo,
```
rahul-sharma-uipath: does this need to be a prop?

packages/apollo-react/src/canvas/core/NodeTypeRegistry.ts
```diff
@@ -0,0 +1,459 @@
...
+ */
+export class NodeTypeRegistry {
+  categoryById = new Map<string, CategoryManifest>();
+  nodeByType = new Map<string, NodeManifest>();
+  nodeHandles = new Map<string, NodeHandles>();
+  categoriesByParent = new Map<string | undefined, CategoryManifest[]>();
+  nodesByCategory = new Map<string | undefined, NodeManifest[]>();
+  categoryAncestors = new Map<string, string[]>();
+  categoryDescendants = new Map<string, string[]>();
+  categoryTree: CategoryTree | null = null;
```
rahul-sharma-uipath: private?

packages/apollo-react/src/canvas/core/NodeTypeRegistry.ts
```diff
@@ -0,0 +1,459 @@
...
+          icon: node.display.icon ?? 'box',
+        });
+      }
+    }
+
+    const workflowManifest: WorkflowManifest = {
+      version: '0',
+      categories: Array.from(categoryMap.values()),
+      nodes,
+    };
```
rahul-sharma-uipath, line 63: currently hte /nodes api explicitly returns categories - we would need a way to ensure you can leverage that still rather than extracting categories from nodes

packages/apollo-react/src/canvas/utils/constraint-validator.ts
```diff
@@ -0,0 +1,248 @@
...
+    return false;
+  }
+
+  return true;
+}
+
+/**
+ * Check if connection is allowed by HandleTarget constraints
+ */
+function checkHandleTargetConstraint(
```
rahul-sharma-uipath, line 80: i'm thinking we should have a one pager for the constraints system - will be helpful documentation even for customers

packages/apollo-react/src/canvas/core/NodeTypeRegistry.ts
BenGSchulz: This method is there for backwards compatibility / making sure this isn't a breaking change. The `registerManifest` method will accept the full manifest including the categories from `/nodes`.

packages/apollo-react/src/canvas/utils/constraint-validator.ts
BenGSchulz: Yeah makes sense, I can work on that.

packages/apollo-react/src/canvas/components/AddNodePanel/AddNodePanel.tsx
BenGSchulz: I suppose I could also use the `usePreviewNode` hook in this component.

rahul-sharma-uipath (approved):

PR #99, opened 2026-01-22
Port theme provider to framework-agnostic implementation

ruudandriessen:
## Summary
Replaced the Next.js-dependent theme provider with a standalone implementation that manages light/dark/system theme state independently. This allows the theme provider to be used outside of Next.js while maintaining CSS variable override functionality for custom themes.

## Changes
- Added theme state management (light/dark/system) with localStorage persistence
- Exported `useTheme` hook for consuming components
- Maintained CSS variable override functionality for custom theme configs
- Updated theme-toggle to use the local hook instead of next-themes

### DIFF
--- a/apps/apollo-vertex/registry/theme-provider/theme-provider.tsx
+++ b/apps/apollo-vertex/registry/theme-provider/theme-provider.tsx
@@ -1,5 +1,7 @@
-import type React from "react";
-import { useCallback, useEffect } from "react";
+import type { ReactNode } from "react";
+import { createContext, useContext, useEffect, useState } from "react";
+
+export type Theme = "light" | "dark" | "system";
 
 export interface ThemeConfig {
   light?: {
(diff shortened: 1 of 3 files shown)


apps/apollo-vertex/registry/theme-provider/theme-provider.tsx
```diff
@@ -111,95 +123,116 @@ const cssVarMap: Record<string, string> = {
...
-const getStoredTheme = (storageKey: string): Theme => {
-  if (typeof window === "undefined") return "system";
-  const stored = localStorage.getItem(storageKey);
-  if (stored === "light" || stored === "dark" || stored === "system") {
-    return stored;
-  }
-  return "system";
-};
+function isValidTheme(value: string | null): value is Theme {
+  return value === "light" || value === "dark" || value === "system";
```
0xr3ngar: I would've 

```
    if (value === null) return false;
    return ["light", "dark", "system"].includes(value);
```

0xr3ngar (approved):

apps/apollo-vertex/registry/theme-provider/theme-provider.tsx
0xr3ngar: I would also cleanup the comments cuz they are visual clutter imo - maybe extract the effects into different hooks with better naming 🤔

0xr3ngar (commented):
check comment ☝️

apps/apollo-vertex/registry/theme-provider/theme-provider.tsx
ruudandriessen: Moved the util functions out of the components entirely

ruudandriessen: Not sure if I wanna go around extracting all the util functions unless we have another spot where we use them TBH. Those effects are basically the component - the rest is just the provider and that's it

ruudandriessen:
I already merged this in a separate pr - idk why i opened two 🤣

PR #100, opened 2026-01-22
feat(apollo-react): added support for Widget mode in ApChat component

maninder-uipath:
Screenshot 2026-01-22 at 3 30 22 PM

### DIFF
--- a/packages/apollo-react/src/material/components/ap-chat/service/ChatModel.ts
+++ b/packages/apollo-react/src/material/components/ap-chat/service/ChatModel.ts
@@ -8,6 +8,7 @@ export enum AutopilotChatMode {
   SideBySide = 'side-by-side',
   FullScreen = 'full-screen',
   Embedded = 'embedded',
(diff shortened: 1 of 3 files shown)


packages/apollo-react/src/material/components/ap-chat/service/ChatModel.ts
```diff
@@ -8,6 +8,7 @@ export enum AutopilotChatMode {
   SideBySide = 'side-by-side',
   FullScreen = 'full-screen',
   Embedded = 'embedded',
+  Widget = 'widget'
```
CalinaCristian, line 11: What's the difference between this and embedded?

CalinaCristian (changes requested):
I would not add a new mode. what is widget vs embedded?

maninder-uipath:
Closing this PR because Embedded mode provides same experience as the new Widget mode that I was introducing.

PR #108, opened 2026-01-22
feat(apollo-vertex): allow consumer to provide company logo [AGVSOL-911]

alincadariu:


apps/apollo-vertex/registry/shell/internal/company.tsx
```diff
@@ -31,7 +37,15 @@ export const Company = ({ companyName, productName }: CompanyProps) => {
...
     >
-      <Box className="w-4 h-4 text-primary" />
+      {companyLogo ? (
+        <img
+          src={companyLogo.url}
+          alt={companyLogo.alt}
+          className="w-8 h-8 text-primary"
+        />
+      ) : (
+        <Box className="w-4 h-4 text-primary" />
```
0xr3ngar, line 47: I would put the alt text here, instead of a default box 🤔 

shouldn't we make both mandatory and pass Box from the top inside of the template

and idk if it should also be an image, what if it is an svg or a component, lets use ReactNode or something

ruudandriessen (approved):

apps/apollo-vertex/registry/shell/internal/shell-layout.tsx
```diff
@@ -1,19 +1,26 @@
 import type { PropsWithChildren } from "react";
+import type { CompanyLogo } from "../shell";
 import { Sidebar } from "./sidebar";
 
 interface ShellLayoutProps {
   companyName: string;
   productName: string;
+  companyLogo?: CompanyLogo;
```
0xr3ngar, line 8: remove optional

0xr3ngar:
maaan you merged it before I reviewed it 😢 @ruudandriessen

PR #118, opened 2026-01-23
chore(apollo-react): expose manifest utils

BenGSchulz:
- Exporting utils so they can be used in flow-workbench and resolving some conflicts.

### DIFF
--- a/packages/apollo-react/src/canvas/utils/icon-registry.tsx
+++ b/packages/apollo-react/src/canvas/utils/icon-registry.tsx
@@ -6,9 +6,9 @@
  * Supports both Lucide icons (via DynamicIcon) and UIPath custom icons.
  */
 
-import { Icons } from '@uipath/apollo-react/canvas';
+import * as Icons from '../icons';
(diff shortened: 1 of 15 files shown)


packages/apollo-react/src/canvas/utils/icon-registry.tsx
```diff
@@ -6,9 +6,9 @@
  * Supports both Lucide icons (via DynamicIcon) and UIPath custom icons.
  */
 
-import { Icons } from '@uipath/apollo-react/canvas';
+import * as Icons from '../icons';
```
CalinaCristian, line 9: we do have alias shortcuts for those, and can add more, but we can keep relative as well.

CalinaCristian (approved):

packages/apollo-react/src/canvas/utils/icon-registry.tsx
BenGSchulz: Needed to make relative to avoid circular dependencies

PR #122, opened 2026-01-26
feat(apollo-vertex): add registry deps CI check [AGVSOL-962]

alincadariu:
## Summary

Add CI validation for apollo-vertex registry.json dependencies on every PR. The new `validate-registry-deps.ts` script ensures:
- Local registryDependencies reference existing components
- External scoped packages (@uipath/*) are recognized as external registry refs
- All npm dependencies exist in package.json

## Test plan

- Validation runs via `pnpm lint:registry-deps --affected` in PR checks
- Script detects missing local components and npm packages
- Supports external registry references with @scope prefixes

### DIFF
--- a/apps/apollo-vertex/scripts/validate-registry-deps.ts
+++ b/apps/apollo-vertex/scripts/validate-registry-deps.ts
@@ -1,0 +1,122 @@
+/**
+ * Validates that all dependencies in registry.json are properly declared.
+ *
+ * Checks:
+ * 1. All `registryDependencies` that are local (non-scoped) must exist in the registry
+ *    - Local: "button", "tooltip" → must exist as registry component names
+ *    - External: "@uipath/use-local-storage" → skipped (external registry references)
+ * 2. All `dependencies` reference packages in package.json (dependencies or devDependencies)
+ */
+
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+interface RegistryItem {
+  name: string;
+  dependencies?: string[];
+  registryDependencies?: string[];
+}
+
+interface Registry {
+  items: RegistryItem[];
+}
+
+interface PackageJson {
+  dependencies?: Record<string, string>;
+  devDependencies?: Record<string, string>;
+}
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const registryPath = join(__dirname, "../registry.json");
+const packageJsonPath = join(__dirname, "../package.json");
+
+function readJsonFile<T>(path: string, name: string): T {
+  let content: string;
+  try {
+    content = readFileSync(path, "utf-8");
+  } catch (error) {
+    console.error(`Failed to read ${name} at ${path}:`, (error as Error).message);
+    process.exit(1);
+  }
+
+  try {
+    return JSON.parse(content);
+  } catch (error) {
+    console.error(`Failed to parse ${name}:`, (error as Error).message);
+    process.exit(1);
+  }
+}
+
+function extractPackageName(dep: string): string {
+  // Handle version specifiers like "react-day-picker@latest" or "recharts@2.15.4"
+  const atIndex = dep.lastIndexOf("@");
+  // Check if @ is not the first character (scoped package like @radix-ui/react-slot)
+  if (atIndex > 0) {
+    return dep.substring(0, atIndex);
+  }
+  return dep;
+}
+
+function isScopedPackage(dep: string): boolean {
+  return dep.startsWith("@");
+}
+
+function main(): void {
+  const registry = readJsonFile<Registry>(registryPath, "registry.json");
+  const packageJson = readJsonFile<PackageJson>(packageJsonPath, "package.json");
+
+  const registryComponentNames = new Set(registry.items.map((item) => item.name));
+  const installedPackages = new Set([
+    ...Object.keys(packageJson.dependencies ?? {}),
+    ...Object.keys(packageJson.devDependencies ?? {}),
+  ]);
+
+  const errors: string[] = [];
+
+  for (const item of registry.items) {
+    // Check registryDependencies
+    if (item.registryDependencies) {
+      for (const dep of item.registryDependencies) {
+        if (isScopedPackage(dep)) {
+          // Scoped packages (e.g., @uipath/use-local-storage) are external registry
+          // references handled by shadcn CLI - skip validation
+          continue;
+        }
+        // Non-scoped names should be local registry components
+        if (!registryComponentNames.has(dep)) {
+          errors.push(
+            `Component "${item.name}" has registryDependency "${dep}" which does not exist in the registry`,
+          );
+        }
+      }
+    }
+
+    // Check dependencies (npm packages)
+    if (item.dependencies) {
+      for (const dep of item.dependencies) {
+        const packageName = extractPackageName(dep);
+        if (!installedPackages.has(packageName)) {
+          errors.push(
+            `Component "${item.name}" has dependency "${packageName}" which is not in package.json`,
+          );
+        }
+      }
+    }
+  }
+
+  if (errors.length > 0) {
+    console.error("Registry dependency validation failed:\n");
+    for (const error of errors) {
+      console.error(`  ✗ ${error}`);
+    }
+    console.error(`\n${errors.length} error(s) found.`);
+    process.exit(1);
+  }
+
+  console.log(
+    `✓ Registry dependency validation passed (${registry.items.length} components checked)`,
+  );
+}
+
+main();
--- a/apps/apollo-vertex/scripts/validate-registry-deps.ts
+++ b/apps/apollo-vertex/scripts/validate-registry-deps.ts
@@ -2,10 +2,12 @@
  * Validates that all dependencies in registry.json are properly declared.
  *
  * Checks:
- * 1. All `registryDependencies` that are local (non-scoped) must exist in the registry
+ * 1. All `registryDependencies` that are local must exist in the registry
+ *    Per shadcn docs, registryDependencies supports 3 formats:
  *    - Local: "button", "tooltip" → must exist as registry component names
- *    - External: "@uipath/use-local-storage" → skipped (external registry references)
- * 2. All `dependencies` reference packages in package.json (dependencies or devDependencies)
+ *    - Namespaced: "@uipath/use-local-storage" → skipped (external registry)
+ *    - Remote URL: "https://example.com/r/item.json" → skipped (external)
+ * 2. All `dependencies` reference packages in package.json
  */
 
 import { readFileSync } from "node:fs";
@@ -62,6 +64,10 @@ function isScopedPackage(dep: string): boolean {
   return dep.startsWith("@");
 }
 
+function isRemoteUrl(dep: string): boolean {
+  return dep.startsWith("http://") || dep.startsWith("https://");
+}
+
 function main(): void {
   const registry = readJsonFile<Registry>(registryPath, "registry.json");
   const packageJson = readJsonFile<PackageJson>(packageJsonPath, "package.json");
@@ -76,14 +82,17 @@ function main(): void {
 
   for (const item of registry.items) {
     // Check registryDependencies
+    // Supports 3 formats per shadcn docs:
+    // 1. Local items: "button", "input" → must exist in registry
+    // 2. Namespaced items: "@acme/input-form" → external registry refs, skip
+    // 3. Remote URLs: "https://example.com/r/item.json" → external, skip
     if (item.registryDependencies) {
       for (const dep of item.registryDependencies) {
-        if (isScopedPackage(dep)) {
-          // Scoped packages (e.g., @uipath/use-local-storage) are external registry
-          // references handled by shadcn CLI - skip validation
+        if (isScopedPackage(dep) || isRemoteUrl(dep)) {
+          // External registry references handled by shadcn CLI - skip validation
           continue;
         }
-        // Non-scoped names should be local registry components
+        // Non-scoped, non-URL names should be local registry components
         if (!registryComponentNames.has(dep)) {
           errors.push(
             `Component "${item.name}" has registryDependency "${dep}" which does not exist in the registry`,
--- a/apps/apollo-vertex/scripts/validate-registry-deps.ts
+++ b/apps/apollo-vertex/scripts/validate-registry-deps.ts
@@ -2,23 +2,31 @@
  * Validates that all dependencies in registry.json are properly declared.
  *
  * Checks:
- * 1. All declared registryDependencies must exist in the registry
- *    Per shadcn docs, registryDependencies supports these formats:
- *    - Local: "button", "tooltip" → must exist as registry component names
- *    - Namespaced @uipath: "@uipath/use-local-storage" → local, strip prefix and validate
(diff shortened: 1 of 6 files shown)


apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
+/**
```
ruudandriessen, line 1: This file is waaaay to large - lets split it up

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+ * Validates that registry.json is properly configured.
+ *
+ * Structure validation:
+ * 1. Every component must have a name
+ * 2. Every component must have files array with at least one file
+ * 3. Each file must have path and type properties
+ * 4. No duplicate component names
+ * 5. File paths must start with "registry/"
```
ruudandriessen, line 9: Split it into these utils?

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+
+// Packages that are expected to exist (not declared per-component)
+const IMPLICIT_PACKAGES = new Set([
+  "react",
+  "react-dom",
+  "next",
+  "next/link",
+  "next/image",
+  "next/navigation",
+  "next/font",
```
ruudandriessen, line 37: These really should not be implicit - we don't have next at all for these components

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+  const uiMatch = importPath.match(/^@\/components\/ui\/([^/]+)/);
+  if (uiMatch) {
+    return uiMatch[1];
+  }
+
+  // @/registry/X/* → "X"
+  const registryMatch = importPath.match(/^@\/registry\/([^/]+)/);
+  if (registryMatch) {
+    return registryMatch[1];
+  }
```
ruudandriessen, line 155: we should never import from registry - that's actually invalid - because there's no "registry" in the consumer project

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+
+  // @/components/X/* (non-ui) → "X"
+  const componentsMatch = importPath.match(/^@\/components\/([^/]+)/);
+  if (componentsMatch && componentsMatch[1] !== "ui") {
+    return componentsMatch[1];
+  }
+
+  // @/hooks/X → "X"
+  const hooksMatch = importPath.match(/^@\/hooks\/([^/]+)/);
```
ruudandriessen, line 164: actually.. how does this work with shadcn - I believe you can configure this path in the consumers `components.json` 🤔  what if we put `hooks: "@/beepboop/path"` there?

ruudandriessen: Separate issue from this PR probably though

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+function parseImports(filePath: string): Set<string> {
+  const imports = new Set<string>();
+
+  if (!existsSync(filePath)) {
+    return imports;
+  }
+
+  const content = readFileSync(filePath, "utf-8");
+
+  const importRegex = /import\s+(?:[\s\S]*?)\s+from\s+["']([^"']+)["']/g;
```
ruudandriessen, line 181: we use quite a bunch of regexes - which feels flaky - can we use an AST instead? May make it a bit more robust

ruudandriessen: guess it can be a follow up as well

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+  );
+
+  const errors: string[] = [];
+
+  // Validate registry structure
+  if (!registry.items || !Array.isArray(registry.items)) {
+    errors.push("registry.json must have an 'items' array");
+    console.error("Registry validation failed:\n");
+    console.error(`  ✗ ${errors[0]}`);
+    process.exit(1);
```
ruudandriessen, line 282: this feels pretty silly, just console log the error instead of putting it in an error array?

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+  for (let i = 0; i < registry.items.length; i++) {
+    const item = registry.items[i];
+    if (item.name) {
+      if (seenNames.has(item.name)) {
+        errors.push(
+          `Duplicate component name "${item.name}" at indices ${seenNames.get(item.name)} and ${i}`,
+        );
+      } else {
+        seenNames.set(item.name, i);
+      }
```
ruudandriessen, line 296: prefer early exits

ruudandriessen (changes requested):

apps/apollo-vertex/scripts/validate-registry-deps.ts
0xr3ngar: +1

apps/apollo-vertex/scripts/validate-registry-deps.ts
0xr3ngar: +1

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+ */
+
+import { existsSync, readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const LOCAL_REGISTRY_SCOPE = "@uipath/";
+
+// Packages that are expected to exist (not declared per-component)
+const IMPLICIT_PACKAGES = new Set([
```
0xr3ngar, line 30: :wat: why is this a Set? also ruud has a point we do not have next for the components

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+const VALID_FILE_TYPES = [
+  "registry:ui",
+  "registry:hook",
+  "registry:lib",
+  "registry:component",
+  "registry:page",
+  "registry:file",
+  "registry:block",
+  "registry:style",
+];
```
0xr3ngar, line 49: instead of doing this, can't we just do `registry:*` because now if we add a folder or something we have to add it manually in here?

0xr3ngar: also registry:file 🤔 do we even have that?

edit: seems like we dont?
Image

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+  type?: string;
+  target?: string;
+}
+
+interface RegistryItem {
+  name?: string;
+  type?: string;
+  files?: RegistryFile[];
+  dependencies?: string[];
+  registryDependencies?: string[];
```
0xr3ngar, line 62: why is everything undefined :wat:

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+  "registry:page",
+  "registry:file",
+  "registry:block",
+  "registry:style",
+];
+
+interface RegistryFile {
+  path?: string;
+  type?: string;
+  target?: string;
```
0xr3ngar, line 54: why is everything undefined

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+  registryDependencies?: string[];
+}
+
+interface Registry {
+  items?: RegistryItem[];
+}
+
+interface PackageJson {
+  dependencies?: Record<string, string>;
+  devDependencies?: Record<string, string>;
```
0xr3ngar, line 71: why undefined?

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+function readJsonFile<T>(path: string, name: string): T {
+  let content: string;
+  try {
+    content = readFileSync(path, "utf-8");
+  } catch (error) {
+    console.error(
+      `Failed to read ${name} at ${path}:`,
+      (error as Error).message,
+    );
+    process.exit(1);
```
0xr3ngar, line 87: should a util kill the entire process? 

also shouldn't the caller of readJsonFile() handle for errors? 
```ts
 const content = readFileSync(path, "utf-8"); // Let fs throw ENOENT if missing
  return JSON.parse(content); // Let JSON.parse throw SyntaxError
```
and then when we call this we wrap a try catch ? or even better we have a util inside of the registry that does something like what https://www.effect.website does

```ts
const catchError = <T>(promise: Promise<T>): Promise<[undefined, T]> | [Error] => {
    return promise.then((data: T) => ([undefined, data])).catch(error => ([error]))
} 

// usgae
const [error, data] = await catchError(someFunc(blabla));
if (error) process.exit(1).... // or handle the error accordingly
// ... use data just fine 
```
then we have some clean code instead of the ugly try catches everywhere 🗿

apps/apollo-vertex/scripts/validate-registry-deps.ts
0xr3ngar: ☝️
the consumer project has components/ui/....
or heck it is even custom defined, we have to scan their shadcn components.json to find out their aliases 

https://ui.shadcn.com/docs/components-json#aliases, it is not a given that the import statement will be `/components/ui`

apps/apollo-vertex/scripts/validate-registry-deps.ts
0xr3ngar: aaaa I literally left the same comment before I read yours XD

apps/apollo-vertex/scripts/validate-registry-deps.ts
0xr3ngar: ruud does have a point, regexes are ❓. I think the way to go is with AST as well

apps/apollo-vertex/scripts/validate-registry-deps.ts
0xr3ngar: ☝️ also with my suggestion to create the catchError util we can store things here

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+
+  const registryComponentNames = new Set(
+    registry.items
+      .filter((item) => item.name)
+      .map((item) => item.name as string),
+  );
+  const installedPackages = new Set([
+    ...Object.keys(packageJson.dependencies ?? {}),
+    ...Object.keys(packageJson.devDependencies ?? {}),
+  ]);
```
0xr3ngar, line 308: why do we need a set? deps and dev deps are different lol

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+  for (let i = 0; i < registry.items.length; i++) {
+    const item = registry.items[i];
+    const itemId = item.name ? `"${item.name}"` : `at index ${i}`;
+
+    // Check required name
+    if (
+      !item.name ||
+      typeof item.name !== "string" ||
+      item.name.trim() === ""
+    ) {
```
0xr3ngar, line 319: ❓ if how can item.name be different than a string? when it is defined as a string | undefined?

Image

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+      errors.push(`Component ${itemId} is missing required 'name' property`);
+      continue; // Can't do further validation without name
+    }
+
+    // Check required files array
+    if (!item.files || !Array.isArray(item.files)) {
```
0xr3ngar, line 325: why do we need to check array.isArray() on a type that is defined as an array?

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+    for (let j = 0; j < item.files.length; j++) {
+      const file = item.files[j];
+
+      // Check required path
+      if (!file.path || typeof file.path !== "string") {
+        errors.push(
+          `Component "${item.name}" file at index ${j} is missing required 'path' property`,
+        );
+        continue;
+      }
```
0xr3ngar, line 347: same thing it's the same check the type is string | undefined and you're doing the same check here

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+
+      // Check path starts with registry/
+      if (!file.path.startsWith("registry/")) {
+        errors.push(
+          `Component "${item.name}" file "${file.path}" must be in the registry/ folder`,
+        );
+      }
+
+      // Check required type
+      if (!file.type || typeof file.type !== "string") {
```
0xr3ngar, line 357: same check, why? type is string | undefined

apps/apollo-vertex/scripts/validate-registry-deps.ts
```diff
@@ -0,0 +1,473 @@
...
+          const packageName = extractPackageName(dep);
+          if (!installedPackages.has(packageName)) {
+            errors.push(
+              `Component "${item.name}" declares dependency "${packageName}" which is not in package.json`,
+            );
+          }
+        }
+      }
+    }
+  }
```
0xr3ngar, line 457: dude I am not reviewing that. extract it into utils, make it more readable so I can follow the flow of the data.

0xr3ngar (commented):
Please take the time to properly review and improve your code before submitting or committing.
This level of quality is not acceptable, even if it is just a script for a CI check. How you do one thing is how you do everything.

I don’t care whether you write it by hand, use AI, or mix both - the only thing that matters is the final result

I’m no longer going to spend time reviewing code that was clearly “vibed together” just because it "works"

- Remove unnecessary `Set`, `Map`, intermediate collections, etc. - use the simplest structure that does the job cleanly
- Remove explicit return types from functions and let TypeScript infer them (they add noise here)  
- Delete pointless comments that just restate what the code obviously does, for example:  
  ```ts
  // Check required name
  if (!item.name || typeof item.name !== "string" || item.name.trim() === "") { ... }
  ```  
  I can already see what the if is doing - the comment is redundant
- Extract duplicated or complex logic into clear, well-named helper functions  
- Drastically reduce nesting - aim for code that’s easy to read at a glance, not 5 levels deep  
  (The only reason this is semi-readable right now is the 2-space indent; with 4 or 8 it would be unreadable)

Please deliver a clean, intentional version. Thanks.

alincadariu:
closing it in favor of https://github.com/UiPath/apollo-ui/pull/122

PR #124, opened 2026-01-26
fix(apollo-react): delete vertical-rectangle type and support switch with square [MST-6216]

KodudulaAshishUiPath:
### This PR

- Removes vertical-rectangle node shape and supports styling for it as a square node 

 ### Demo

https://github.com/user-attachments/assets/81d52d81-e95d-49f6-897c-9828f1face3c

### DIFF
--- a/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.styles.ts
+++ b/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.styles.ts
@@ -196,9 +196,9 @@ export const BaseIconWrapper = styled.div<{
   height: ${({ height, width, shape }) => {
     // Use 7/8 scaling for a vertical rectangle, and use default 3/4 scaling for other shapes
     const scaleFactor = height ? height / 96 : 1;
-    return height !== width && shape === 'vertical-rectangle'
+    return height !== width && shape !== 'rectangle' // True for only a expandable node
       ? `${84 * scaleFactor}px`
-      : `${72 * scaleFactor}px`
+      : `${72 * scaleFactor}px`;
   }};
   display: flex;
   align-items: center;
--- a/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
+++ b/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
@@ -135,12 +135,11 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
 
     const leftRightHandles = Math.max(leftHandles, rightHandles);
 
-    if (leftRightHandles <= 3) return height;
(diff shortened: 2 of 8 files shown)


packages/apollo-react/src/canvas/components/BaseNode/BaseNode.styles.ts
```diff
@@ -193,8 +193,8 @@ export const BaseIconWrapper = styled.div<{
   }};
  height: ${({ height, width, shape }) => {
    const scaleFactor = height ? height / 96 : 1;
-   return height !== width && shape === 'vertical-rectangle'
-     ? `${84 * scaleFactor}px` // Using 7/8 scaling for a vertical rectangle
+   return height !== width && shape !== 'rectangle'
```
rahul-sharma-uipath: what will happen for circles?

akraminakib:
I do not think rectangle nodes should look like that? See [this PR](https://github.com/UiPath/apollo-ui/pull/125)

Can we coordinate to make sure there's no visual regressions for rectangle nodes?

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.styles.ts
```diff
@@ -193,8 +193,8 @@ export const BaseIconWrapper = styled.div<{
...
+     ? `${84 * scaleFactor}px` // Using 7/8 scaling for a switch node type
```
BenGSchulz: nit: nodes other than switch could have different height scale factor.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
```diff
@@ -270,7 +266,7 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
     showNotches,
     nodeWidth: width,
     nodeHeight: height,
-    isExpandable: displayShape === 'vertical-rectangle',
+    isExpandable: width !== height && displayShape !== 'rectangle',
```
BenGSchulz: Could you explore what it would be like to not have to use `isExpandable` at all? It seems like we could just ensure that the handle spacing / alignment is handled consistently for any number of handles.

However, we may need to explore what it means for circle nodes to have more handles, or if we even want to allow that.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.styles.ts
KodudulaAshishUiPath: For circles, height === width so scaling factor would be 3/4 as expected

snuziale (approved):

BenGSchulz (approved):

PR #126, opened 2026-01-27
feat(apollo-react): Ap-Popover component implementation PLT-95438

Miyuki-L:
- Implemented Ap-Popover using Mui Popover
- Original popover in apollo-design-system was a pure stencil component and more complex than necessary to migrate. 

https://github.com/user-attachments/assets/d61022cb-7fb6-4c29-ab2d-e2d1d00f61b7

### DIFF
--- a/apps/react-playground/src/pages/components/PopoverShowcase.tsx
+++ b/apps/react-playground/src/pages/components/PopoverShowcase.tsx
@@ -1,0 +1,282 @@
+import {
+	ApAccordion,
+	ApButton,
+	ApPopover,
+} from "@uipath/apollo-react/material/components";
+import styled from "styled-components";
+import { CodeBlock } from "../../components/CodeBlock";
+
+import {
+	PageContainer,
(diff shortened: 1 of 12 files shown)


apps/react-playground/src/pages/components/PopoverShowcase.tsx
```diff
@@ -0,0 +1,217 @@
...
+					>
+						<div style={{ padding: "16px", maxWidth: "200px" }}>
+							This is a basic popover example positioned below and to the left
+							of the anchor point.
+						</div>
+					</ApPopover>
+
+					<Label>Code Example:</Label>
+					<ApAccordion label="Uncontrolled Component" disableDivider>
+						<CodeBlock>
```
CalinaCristian: should we use something like `react-syntax-highlighter` here?

CalinaCristian (approved):

CalinaCristian (approved):

PR #133, opened 2026-01-28
feat(apollo-vertex): add language switcher [AGVSOL-1055, AGVSOL-1045]

alincadariu:
The goal: consumer would only need to provide the translation files.

This PR adds:
- Locale provider
- type safety for translation keys
- i18n initialization
- language switcher

Screenshot 2026-01-28 at 15 25 13


apps/apollo-vertex/app/vertex-components/language-toggle/page.mdx
```diff
@@ -0,0 +1,123 @@
...
+    .detail;
+  i18n.changeLanguage(selectedLanguageId);
+  document.documentElement.lang = selectedLanguageId;
+};
+
+useEffect(() => {
+  document.addEventListener("languageChanged", handleLanguageChanged);
+  return () =>
+    document.removeEventListener("languageChanged", handleLanguageChanged);
+}, [handleLanguageChanged]);
```
0xr3ngar: should use the new useEffectEvent hook https://react.dev/reference/react/useEffectEvent

apps/apollo-vertex/registry/shell/internal/language-toggle.tsx
```diff
@@ -0,0 +1,58 @@
...
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuTrigger,
+} from "@/components/ui/dropdown-menu";
+
+interface LanguageToggleProps {
+  supportedLocales: {
+    code: string;
+    name: string;
```
0xr3ngar: hmm shouldn't this name be translated? 🤔 

Because in German, Germany is `Deutschland` but in English is `German` and in Japanese is `ドイツ語` and now we kinda hardcode the names?

0xr3ngar: also I'm thinking should we also use the types from process mining, so we can have type safe translation keys 🤔  https://github.com/UiPath/ProcessMining-Frontend/blob/main/client/src/models/TranslationKey.ts

0xr3ngar (changes requested):

apps/apollo-vertex/registry/language-toggle/language-toggle.tsx
```diff
@@ -0,0 +1,58 @@
...
+      }),
+    );
+  }, []);
+
+  return (
+    <DropdownMenu>
+      <DropdownMenuTrigger asChild>
+        <Button variant="outline" size="icon">
+          <Globe className="h-[1.2rem] w-[1.2rem]" />
+          <span className="sr-only">Switch language</span>
```
ruudandriessen: shouldn't this also be translated?

apps/apollo-vertex/app/vertex-components/language-toggle/page.mdx
```diff
@@ -0,0 +1,123 @@
+import { LanguageToggle } from '@/registry/language-toggle/language-toggle'
+
+# Language Toggle
+
+A dropdown menu component that allows users to toggle between different language preferences.
+
+<div className="p-4 border rounded-lg mt-4">
```
ruudandriessen: Why do we need docs for this, it's part of the shell no?

apps/apollo-vertex/registry.json
```diff
@@ -664,6 +665,20 @@
           "type": "registry:ui"
         }
       ]
+    },
+    {
+      "name": "language-toggle",
```
ruudandriessen: Wouldn't we just make this part of the shell, so no need to add this as another entry point in the shell?

ruudandriessen (changes requested):

apps/apollo-vertex/registry.json
alincadariu: @ruudandriessen what is the difference then between `theme-toggle` and this one? I'd either make both part of the shell or expose both as individual components

apps/apollo-vertex/app/vertex-components/language-toggle/page.mdx
alincadariu: @ruudandriessen see my comment below

apps/apollo-vertex/registry.json
ruudandriessen: Oh theme toggle indeed also shouldn't be exposed 😛

0xr3ngar (approved):

angeloaltamiranom (approved):

apps/apollo-vertex/registry/shell/internal/locale-provider.tsx
```diff
@@ -0,0 +1,23 @@
+import { type PropsWithChildren, useEffect, useEffectEvent } from "react";
+import { useTranslation } from "react-i18next";
+import "@/lib/i18n";
```
ruudandriessen: imports with side effects should be avoided

ruudandriessen (changes requested):

apps/apollo-vertex/registry/shell/internal/locale-provider.tsx
alincadariu: @ruudandriessen done

ruudandriessen (approved):

CalinaCristian:
### Current Stack

- `main`

PR #137, opened 2026-01-29
ci: add registry component validation check [AGVSOL-962]

alincadariu:
example fail which commit https://github.com/UiPath/apollo-ui/pull/137/commits/d50af1a6d82cb47a77cc934e19648b187443975b fixes
Screenshot 2026-02-02 at 18 59 11


.github/scripts/test-registry/configure-registry.ts
```diff
@@ -0,0 +1,35 @@
...
+  const configPath = './components.json';
+
+  if (!registryUrl) {
+    console.error('REGISTRY_URL environment variable is required');
+    process.exit(1);
+  }
+
+  const config: ComponentsConfig = JSON.parse(readFileSync(configPath, 'utf-8'));
+
+  config.registries = config.registries || {};
```
0xr3ngar: ?? instead of ||

.github/scripts/test-registry/configure-registry.ts
```diff
@@ -0,0 +1,35 @@
...
+function main(): void {
+  const registryUrl = process.env.REGISTRY_URL;
+  const configPath = './components.json';
+
+  if (!registryUrl) {
+    console.error('REGISTRY_URL environment variable is required');
+    process.exit(1);
+  }
+
+  const config: ComponentsConfig = JSON.parse(readFileSync(configPath, 'utf-8'));
```
0xr3ngar: these can fail, wrap both into try catches

let configFileString;
try{} catch (error) { nice error }

let jsonConfig;

try {} catch (error) { nice error }

.github/scripts/test-registry/configure-registry.ts
```diff
@@ -0,0 +1,35 @@
...
+    string,
+    {
+      url: string;
+      params?: Record<string, string>;
+    }
+  >;
+  [key: string]: unknown;
+}
+
+function main(): void {
```
0xr3ngar: Why the : void

.github/scripts/test-registry/configure-registry.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  const config: ComponentsConfig = JSON.parse(readFileSync(configPath, 'utf-8'));
+
+  config.registries = config.registries || {};
+  config.registries['@uipath'] = {
+    url: `${registryUrl}/r/{name}.json`,
+  };
+
+  writeFileSync(configPath, JSON.stringify(config, null, 2));
```
0xr3ngar: this can also fail, try catch

.github/scripts/test-registry/test-registry.ts
```diff
@@ -0,0 +1,52 @@
...
+  } catch (error) {
+    const output =
+      error instanceof Error && 'stdout' in error
+        ? String((error as { stdout: string }).stdout)
+        : String(error);
+    return { component, success: false, output };
+  }
+}
+
+function main(): void {
```
0xr3ngar: why the : void

.github/scripts/test-registry/test-registry.ts
```diff
@@ -0,0 +1,52 @@
...
+    const output = execSync(`pnpm exec shadcn add "@uipath/${component}"`, {
+      encoding: 'utf-8',
+      stdio: ['pipe', 'pipe', 'pipe'],
+    });
+    return { component, success: true, output };
+  } catch (error) {
+    const output =
+      error instanceof Error && 'stdout' in error
+        ? String((error as { stdout: string }).stdout)
+        : String(error);
```
0xr3ngar: shouldn't we capture both stdout and stderr 🤔

.github/scripts/test-registry/test-registry.ts
```diff
@@ -0,0 +1,52 @@
...
+
+function main(): void {
+  const registryPath = process.env.REGISTRY_PATH;
+
+  if (!registryPath) {
+    console.error('REGISTRY_PATH environment variable is required');
+    process.exit(1);
+  }
+
+  const components = JSON.parse(readFileSync(registryPath, 'utf-8')).items;
```
0xr3ngar: same as above these 2 can fail wrap in try catch

.github/scripts/test-registry/test-registry.ts
```diff
@@ -0,0 +1,52 @@
...
+  for (const item of components) {
+    const result = testComponent(item.name);
+
+    if (result.success) {
+      console.log(`Passed: ${item.name}`);
+    } else {
+      console.log(`FAILED: ${item.name}`);
+      console.log('Error output:');
+      console.log(result.output);
+      process.exit(1);
```
0xr3ngar: won't this exit on first component error? Shouldn't we store all errors into an array outside and then print all errors after the for loop?

.github/workflows/test-registry.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Build registry
+        run: pnpm --filter apollo-vertex registry:build
+
+      - name: Start dev server
+        run: |
+          pnpm --filter apollo-vertex dev &
+
+      - name: Create shadcn app
+        run: cd /tmp && pnpm dlx shadcn@latest create minimal-app --preset "https://ui.shadcn.com/init?base=radix&style=vega&baseColor=neutral&theme=neutral&iconLibrary=lucide&font=inter&menuAccent=subtle&menuColor=default&radius=default&template=next&rtl=false" --template next
```
0xr3ngar: what is pnpm dlx? shoudn't this be pnpx?

0xr3ngar (changes requested):

.github/workflows/test-registry.yml
alincadariu: same thing https://pnpm.io/cli/dlx

0xr3ngar: aaa oki 👌

0xr3ngar (approved):

.github/workflows/apollo-vertex-registry-check.yml
```diff
@@ -0,0 +1,63 @@
...
+  contents: read
+
+jobs:
+  test-registry:
+    name: Test Registry Components
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
```
ruudandriessen, line 26: why don't we just use v4?

ruudandriessen (changes requested):

.github/workflows/apollo-vertex-registry-check.yml
alincadariu: @ruudandriessen https://github.com/UiPath/apollo-ui/pull/154

.github/scripts/test-registry/test-registry.ts
```diff
@@ -0,0 +1,76 @@
...
+    registry = JSON.parse(registryString);
+  } catch (error) {
+    console.error(`Failed to parse registry.json: ${error}`);
+    process.exit(1);
+  }
+
+  const errors: Array<{ component: string; output: string }> = [];
+
+  for (const item of registry.items) {
+    const result = testComponent(item.name);
```
ruudandriessen: we don't test each of them independently now right? We just add them one by one? What if:

* We first install component A
* Then install component B, which just missed a dependency on either A or a dependency of A
* Test of B will now pass, but if only B was installed, it would fail.

ruudandriessen (changes requested):

.github/scripts/test-registry/test-registry.ts
alincadariu: @ruudandriessen I don't completely get it.

if I take the example from the description:
- We first install component A (use-local-storage)
- Then install component B (theme-toggle), which had an incorrect configured dependency on Component A (use-local-storage)
- Test of B (theme-toggle) failed.

CalinaCristian:
### Current Stack

- `main`
  - #137 `alincadariu/phoenix` **← THIS PR**

ruudandriessen (approved):

PR #141, opened 2026-01-30
fix(apollo-react): initialize base node height only after reactFlow measure [MST-6329]

KodudulaAshishUiPath:
### Description

This PR

- Initialises originalHeightRef in BaseNode only after ReactFlow measures it dynamically. This prevents rendering issues when no height and width are passed to BaseNode. 

### Demo

Before:

Screenshot 2026-01-30 at 11 59 30

After:

Screenshot 2026-01-30 at 11 59 14

### DIFF
--- a/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
+++ b/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
@@ -44,7 +44,6 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
   const [isHovered, setIsHovered] = useState(false);
   const [isFocused, setIsFocused] = useState(false);
 
-  const isInitialMountRef = useRef(true);
   const originalHeightRef = useRef<number | undefined>(undefined);
 
   // Get execution status from external source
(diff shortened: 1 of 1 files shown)


packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
```diff
@@ -140,9 +139,9 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
 
   // Sync computed height to node when it changes
   useEffect(() => {
-    if (isInitialMountRef.current) {
```
rahul-sharma-uipath, line 143: why remove this though

KodudulaAshishUiPath: Earlier we were using it to track original height after first render, but we cannot be sure that height would be calculated after first render. ReactFlow measures it after some 2-3 renders, till then height would be 0(if not provided in manifest), so updating the originalHeightRef only when ref is undefined and height is 0

snuziale (approved):

CalinaCristian:
### Current Stack

- `main`

PR #167, opened 2026-02-05
fix(apollo-react): filter invisible handles in calcuation of spacing [MST-6413]

KodudulaAshishUiPath:
### Description

This PR

- Filters out handles with `visible:true` before calculation of spacing on the node
- Adds a story to demonstrate the same

### Demo

Before:

https://github.com/user-attachments/assets/e661fad2-0dc2-43ad-970f-af793db90e6c

After:

https://github.com/user-attachments/assets/4cf8285e-cdc4-4f50-8f70-456020f1ae78

### DIFF
--- a/packages/apollo-react/src/canvas/components/ButtonHandle/ButtonHandle.tsx
+++ b/packages/apollo-react/src/canvas/components/ButtonHandle/ButtonHandle.tsx
@@ -256,12 +256,15 @@ const ButtonHandlesBase = ({
     selected: boolean;
   }) => boolean;
 }) => {
-  const total = handles.length;
   const finalSelected = shouldShowAddButtonFn({ showAddButton, selected });
+
+  // Filter to only visible handles for spacing calculations
+  const visibleHandles = handles.filter((h) => visible && (h.visible ?? true));
+  const total = visibleHandles.length;
 
   return (
     <>
-      {handles.map((handle, index) => (
+      {visibleHandles.map((handle, index) => (
         <ButtonHandle
           key={handle.id}
           id={handle.id}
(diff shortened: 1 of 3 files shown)


packages/apollo-react/src/canvas/components/ButtonHandle/ButtonHandle.tsx
```diff
@@ -256,36 +256,47 @@ const ButtonHandlesBase = ({
...
-          showButton={finalSelected && visible && handle.showButton}
-          color={handle.color}
-          onAction={handle.onAction}
-          showNotches={showNotches}
-          customPositionAndOffsets={customPositionAndOffsets}
-          nodeWidth={nodeWidth}
-          nodeHeight={nodeHeight}
-        />
-      ))}
+      {handles.map((handle, originalIndex) => {
```
BenGSchulz: What would happen if we had an edge connected to the invisible handle? Would it look like it was connected to a visible handle positioning-wise?

If we don't support connections to invisible handles, then I would be confused as to why we include them in the final handle group in the first place.

KodudulaAshishUiPath: Yeah I'm also thinking of some use case where they would be needed. Rahul confirmed that we wouldn't be allowing edges from invisible handles, but I'm not sure if we could delete them entirely

BenGSchulz: Maybe we can keep them but use separate arrays rather than nesting the slice + filter inside the `handles.map`? So at top-level of component we split handles -> visibleHandles and invisibleHandles.

I know the `<ButtonHandle>` would need to be duplicated, but seems like we could actually reduce the code by spreading the handle item into the props rather than listing them all out.

Another solution could be to precalculate a mapping of index to visibleIndex outside of the `handles.map`

CalinaCristian:
### Current Stack

- `main`
  - #167 `fix/MST-6413` **← THIS PR**

BenGSchulz (approved):

PR #172, opened 2026-02-05
feat(apollo-react): make empty stage node text clickable [MST-6401]

kittyyueli:
## Description
Updated the StageNode component to display the empty state content as a clickable link instead of plain text when task addition is available and the node is not read-only.

## Demo
image

## Key Changes
- Added `ApLink` import to the material components
- Replaced static `ApTypography` with conditional rendering:
  - When `onTaskAdd` or `onAddTaskFromToolbox` callbacks exist and the node is not read-only, the empty state text is now rendered as an `ApLink` that triggers `handleTaskAddClick`
  - Otherwise, it falls back to the original `ApTypography` display
- This improves UX by making the empty state more discoverable and actionable

## Implementation Details
- The link uses the same `FontVariantToken.fontSizeS` variant for visual consistency
- The conditional logic checks for both task addition callbacks and read-only status to determine interactivity
- The change maintains backward compatibility for read-only nodes and nodes without task addition handlers

https://claude.ai/code/session_01Lvobb2cFYqMc8QB2WmwGz9

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNode.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNode.tsx
@@ -19,7 +19,13 @@ import {
 import { FontVariantToken, Padding, Spacing } from '@uipath/apollo-core';
 import { Column, Row } from '@uipath/apollo-react/canvas/layouts';
 import { Position, useStore, useViewport } from '@uipath/apollo-react/canvas/xyflow/react';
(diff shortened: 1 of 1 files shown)


alextran1 (approved):

Miyuki-L (approved):

CalinaCristian:
### Current Stack

- `main`

PR #181, opened 2026-02-07
fix(apollo-react): handle dark themed icon backgrounds on nodes [MST-6275]

BenGSchulz:
- Fixes icon background color handling for dark theme.
  - Previously did not use the `colorDark` on category manifest, now will use when dark theme is used.
  - Previously had no dark variant for `iconBackground` in node manifest, this adds `iconBackgroundDark` and uses it when dark theme is used.

https://github.com/user-attachments/assets/1b4692f6-9ca7-4191-bc44-12e5e6ba1edd

### DIFF
--- a/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
+++ b/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
@@ -46,6 +46,7 @@ const BaseCanvasInnerComponent = <NodeType extends Node = Node, EdgeType extends
     backgroundVariant = BASE_CANVAS_DEFAULTS.background.variant,
     backgroundGap = BASE_CANVAS_DEFAULTS.background.gap,
     backgroundSize = BASE_CANVAS_DEFAULTS.background.size,
+    isDarkMode,
 
     // Configuration
     minZoom = BASE_CANVAS_DEFAULTS.zoom.min,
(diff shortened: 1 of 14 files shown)


CalinaCristian (approved):

packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
```diff
@@ -101,6 +103,11 @@ const BaseCanvasInnerComponent = <NodeType extends Node = Node, EdgeType extends
   const isInteractive = mode !== 'readonly';
   const isDesignMode = mode === 'design';
 
+  // Check MUI theme for dark mode if not explicitly set by prop
+  const muiTheme = useTheme();
```
rahul-sharma-uipath: i don't understand, doesn't this tie the canvas to having a material theme provider?

BenGSchulz: It won't require the theme, as you can still just provide the prop.

I used this as a quick workaround for getting the theming to work in the storybook (which uses mui) because otherwise will need to change a bunch of plumbing to use the prop. I can remove though if you think is a blocker.

CalinaCristian:
### Current Stack

- `main`
  - #181 `fix/node-icon-themeing` **← THIS PR**

rahul-sharma-uipath (approved):

PR #205, opened 2026-02-11
fix(apollo-react): increase icon size for URL based icons [MST-6486]

KodudulaAshishUiPath:
## Description

This PR

- Allows fallback to CSS iconSize in BaseNode if no height or width is given

## Demo
Before:
Screenshot 2026-02-11 at 18 03 57
After:
Screenshot 2026-02-11 at 18 03 02

### DIFF
--- a/packages/apollo-react/src/canvas/utils/icon-registry.tsx
+++ b/packages/apollo-react/src/canvas/utils/icon-registry.tsx
@@ -73,8 +73,8 @@ export function getIcon(iconId: string): IconComponent {
         src={iconId}
         alt="icon"
         style={{
-          width: w ?? 24,
-          height: h ?? 24,
+          width: w,
+          height: h,
         }}
       />
     );


packages/apollo-react/src/canvas/utils/icon-registry.tsx
```diff
@@ -69,8 +69,8 @@ export function getIcon(iconId: string): IconComponent {
         src={iconId}
         alt="icon"
         style={{
-          width: w ?? 24,
-          height: h ?? 24,
+          width: w,
+          height: h,
```
rahul-sharma-uipath, line 77: i still think we should have a fallback, but we should align on a larger size for the fallback

KodudulaAshishUiPath: We can have a fallback of 40 to get the same size as other SVGs. But then we would be controlling the sizes at 2 different places, in CSS for nodes like Agents, HITL. For connector nodes in this file. I think it would be better if we could keep both in CSS

rahul-sharma-uipath: so then can we only rely on CSS for this with some sort of `data-size` attribute? 

e.g. `<svg [data-size]="small"` or `<svg [data=size]="medium"`

and map these to CSS sizes?

KodudulaAshishUiPath: We can do that but we would face the same issue, some images are <svg/> and some are  We need to include `data-size` for both of those. I think it would be better with the current setup of using the consumer sizes if they are passed else falling back to 40 for both svg, img in CSS.

rahul-sharma-uipath (approved):

PR #214, opened 2026-02-12
feat(apollo-vertex): update status tokens [AGVSOL-1019]

alincadariu:


apps/apollo-vertex/app/globals.css
```diff
@@ -46,6 +46,12 @@
 
   --color-destructive: var(--destructive);
 
+  --color-success: var(--success);
```
ruudandriessen, line 49: how will this work for the consumers?

ruudandriessen: I guess we can add part of this in the template? Then I think we also need to compare this in the adoption tracker?

ruudandriessen (approved):

apps/apollo-vertex/app/globals.css
alincadariu: @ruudandriessen i'm thinking we might need to set up a registry theme 🤔  https://ui.shadcn.com/docs/registry/examples#registrytheme

ruudandriessen: ooooo I didn't see this feature before - that does sound like it makes sense

frankkluijtmans (approved):

apps/apollo-vertex/app/globals.css
alincadariu: @ruudandriessen created https://uipath.atlassian.net/browse/AGVSOL-1201

alincadariu: @ruudandriessen https://github.com/UiPath/apollo-ui/pull/216

PR #216, opened 2026-02-12
feat(apollo-vertex): setup registry theme [AGVSOL-1201]

alincadariu:


apps/apollo-vertex/scripts/generate-theme-css.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+const sections: string[] = [
+  "",
+  renderBlock("@theme inline", cssVars.theme),
+  "",
+  renderBlock(":root", cssVars.light),
+  "",
+  renderBlock(".dark", cssVars.dark),
+  "",
```
0xr3ngar: nit: these "" are not needed since in the renderBlock you already return a \n character

0xr3ngar (approved):

PR #218, opened 2026-02-12
feat(apollo-react): allow passing in error cta to agent flow nodes [AG-700]

akraminakib:
https://uipath.atlassian.net/browse/AG-700

- Allows consumers of agent flow to pass error cta action metadata (label & icon) on a per resource level
- Provides a way to give a callback which gets invoked when a cta action is clicked from the toolbar. This will allow consumers to route to the right logic from checks on their end

https://github.com/user-attachments/assets/3b005a62-5403-4697-a1b9-79578af53f46


packages/apollo-react/src/canvas/components/AgentCanvas/nodes/ResourceNode.tsx
```diff
@@ -236,9 +251,12 @@ export const ResourceNode = memo(
         onAction: () => {},
       };
 
-      const actions: ToolbarAction[] = [removeAction];
+      // Only show error CTA if `errorAction` is provided in data
+      // Consumers need to conditionally pass this prop if they want to show the error CTA
+      const actions: ToolbarAction[] = [removeAction, ...(errorAction ? [errorCtaAction] : [])];
+
       const overflowActions: ToolbarAction[] = [
-        ...(data.type !== 'memorySpace' ? [toggleBreakpointAction] : []),
```
akraminakib, line 241: In SW form editor, memory spaces now accept breakpoints, so I'm removing this logic here which inhibited that functionality on the canvas

swathiJayav (approved):

PR #222, opened 2026-02-13
feat(apollo-wind file-upload): add per-file error display and external errors support to FileUpload

maninder-uipath:
https://uipath.atlassian.net/browse/PLT-97253

Also enforced type check. Added a new story for external errors prop
Screenshot 2026-02-13 at 3 39 53 PM

### DIFF
--- a/packages/apollo-wind/src/components/ui/file-upload.stories.tsx
+++ b/packages/apollo-wind/src/components/ui/file-upload.stories.tsx
@@ -116,3 +116,49 @@ export const WithMaxSize = {
     );
   },
 };
+
+export const WithExternalErrors = {
+  args: {},
+  render: () => {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -102,7 +102,8 @@
       "body-parser": "^2.2.1",
       "lodash": "^4.17.23",
       "lodash-es": "^4.17.23",
-      "caniuse-lite": "1.0.30001756"
+      "caniuse-lite": "1.0.30001756",
+      "zod": "^4.3.5"
```
CalinaCristian: Why is this an override here? It should just be the installed version across the repo not an override 🤔

maninder-uipath: Reverted

CalinaCristian (approved):

PR #223, opened 2026-02-13
fix(apollo-vertex): remove next from registry components

alincadariu:


apps/apollo-vertex/registry/shell/shell-minimal-nav-item.tsx
```diff
@@ -1,7 +1,6 @@
 "use client";
 
-import Link from "next/link";
-import { usePathname } from "next/navigation";
+import { Link } from "@radix-ui/react-navigation-menu";
```
ruudandriessen, line 3: this should be the nav item of the consumer no? So this won't work?

ruudandriessen (changes requested):

apps/apollo-vertex/registry/shell/shell-minimal-nav-item.tsx
alincadariu: @ruudandriessen in what sense this won't work?
this is just a link component from radix ui 🤔

ruudandriessen: Yeah but the nav item should be a router link - like how it was before from next, but in this case, likely from something like tanstack router

alincadariu: @ruudandriessen ah you are right.

alincadariu: @ruudandriessen should we just leave it up to the consumer to fix this for now?
I don't see a nice way of handling this 🤔

ruudandriessen: Yes I think it's either that or push for tanstack router tbh

PR #227, opened 2026-02-14
fix(apollo-react): cleanup node manifests [PLT-000]

rahul-sharma-uipath:
simplify manifest export

image

now, we have 2 types of exports
`NodeInstance` and `NodeManifest`

the `Instance` denotes all instances of the node that are ultimately plumbed to node `data`
the `Manifest` denotes all the manifest definitions, or all the items that will end up in the node registry

also remove `WorkflowManifest` export -> `apollo-ui` has no stance on how users structure their manifest response, so long as they provide the list of category manifests and node manifests 

finally, introduce `inputDefaults`, `drillable`, and `debug` as top-level keys in the manifest
`inputDefaults` -> intended to specify the default values in the inputDefinition
`drillable` -> intended to specify whether a node can be drilled into with a child canvas
`debug` -> intended to specify the runtime for when users click on "run single node"


packages/apollo-react/src/canvas/schema/node-definition/node-manifest.ts
```diff
@@ -81,12 +89,18 @@ export const nodeManifestSchema = z.object({
...
+
   /** Input definition for the node, record of string to any */
   inputDefinition: z.record(z.string(), z.any()).optional(),
 
   /** Output definition for the node, record of string to any */
   outputDefinition: z.record(z.string(), z.any()).optional(),
 
+  /** Debug configuration for the node */
+  debug: nodeDebugManifestSchema.optional(),
+
```
snuziale, line 106: can you add drillable for me:

Image

snuziale (approved):

packages/apollo-react/src/canvas/schema/node-definition/node-manifest.ts
rahul-sharma-uipath: added

PR #234, opened 2026-02-16
fix(apollo-react): nest dynamic handle data under inputs to match resolver context

snuziale:
## Summary
- Fix BaseNode handle resolution to support per-instance handleConfigurations in node data (Priority 2), enabling edge stories and runtime handle overrides without a manifest
- Fix three stale/incorrect useMemo dependency arrays in BaseNode: display missed data changes, handleConfigurations missed id, and smartHandleElements used connectedHandleIds.has (always Set.prototype.has — never changes)
- Fix createNode storybook utility to not pollute node data with undefined values for optional fields
- Add DynamicHandles story demonstrating repeat expressions and templated handle labels

image


CalinaCristian (approved):
Will check why vercel deployment failed

PR #236, opened 2026-02-17
fix(apollo-vertex): use custom auth [AGVSOL-904]

alincadariu:
Screenshot 2026-02-17 at 13 45 48


frankkluijtmans (approved):
I can't really test it in the vertex UI since it's not configured with a real client ID, but it looks like this follows the same auth/refresh logic from invoice processing which looks to be stable 👌

PR #244, opened 2026-02-18
fix(apollo-react): handles in flow [MST-6514]

ayush578:
add button was hidden due to text label for handles.
limitize the text length when add button needs to be shown

https://github.com/user-attachments/assets/1343f8a7-08f4-4272-bc6d-c3916bd9edd4


KodudulaAshishUiPath:
I think It would be better if we show the complete text in tooltip when the label is truncated

snuziale (approved):

CalinaCristian (approved):

PR #253, opened 2026-02-23
chore(apollo-vertex): oxlint biome setup [AGVSOL-1265]

alincadariu:
add oxlint for linting and biome for formatting to be in sync with the template


apps/apollo-vertex/registry/sidebar/sidebar.tsx
```diff
@@ -1,9 +1,10 @@
+// oxlint-disable eslint(max-lines)
```
0xr3ngar: why do we need to disable maxlines? can't we just extract the functions into a internal/ folder or some other folder?

0xr3ngar (approved):

apps/apollo-vertex/registry/sidebar/sidebar.tsx
alincadariu: @0xr3ngar extracted the sidebar into separate files

PR #262, opened 2026-02-24
refactor(apollo-vertex): simplify shell props [AGVSOL-1333]

alincadariu:


frankkluijtmans (changes requested):
I think `window.location.origin` doesn't always match the base URL we need for auth endpoints 🤔  For instance local deployment, or my github pages deployment. It's fine as a default, but should be overridable.

alincadariu:
> I think `window.location.origin` doesn't always match the base URL we need for auth endpoints 🤔 For instance local deployment, or my github pages deployment. It's fine as a default, but should be overridable.

@frankkluijtmans could you show me an example which wouldn't work? 🤔 
also this auth file is overridable, so the consumer can def change it easily (if this happens, of course we will be seeing it in the adoption tracker as modified and we should probs adapt)

frankkluijtmans:
> > I think `window.location.origin` doesn't always match the base URL we need for auth endpoints 🤔 For instance local deployment, or my github pages deployment. It's fine as a default, but should be overridable.
> 
> @frankkluijtmans could you show me an example which wouldn't work? 🤔 also this auth file is overridable, so the consumer can def change it easily (if this happens, of course we will be seeing it in the adoption tracker as modified and we should probs adapt)

Anytime origin doesn't match, I can already assure you I can't adopt this change as is 😅 This will really only work when deployed in staging.

alincadariu:
> > > I think `window.location.origin` doesn't always match the base URL we need for auth endpoints 🤔 For instance local deployment, or my github pages deployment. It's fine as a default, but should be overridable.
> > 
> > 
> > @frankkluijtmans could you show me an example which wouldn't work? 🤔 also this auth file is overridable, so the consumer can def change it easily (if this happens, of course we will be seeing it in the adoption tracker as modified and we should probs adapt)
> 
> Anytime origin doesn't match, I can already assure you I can't adopt this change as is 😅 This will really only work when deployed in staging.

@frankkluijtmans checked invoice processing and I see what you mean, will leave the baseUrl up to the consumer

frankkluijtmans (approved):

PR #266, opened 2026-02-24
feat(apollo-vertex): badge component updates

petervachon:
Status badges combine a status type with a variant appearance (primary, secondary, or outline). Icons are optional and can be passed as children.


ruudandriessen (changes requested):
Like @petervachon mentioned to me on slack - colors don't seem to  be correct- we'll have to fix that before merging

frankkluijtmans (approved):

alincadariu (approved):

ruudandriessen (approved):

PR #272, opened 2026-02-26
feat(apollo-vertex): shell updates

petervachon:
Currently under review


petervachon:
These shell, card, and color general theme updates are ready for Eng review

frankkluijtmans (changes requested):
Other than the linting issues I noticed some small things in the shell UI:

Image

User avatar has some misaligned circle around it, maybe it's an incorrect drop shadow?


Image

Positioning of user menu is off. Would be better to align it at the bottom of the avatar in the simple shell layout.


Image

The KPI cards have some extra vertical space. What's the exact purpose of that? I get that we might want a minimum height, but then let's use a slightly bigger font.

If you need some help looking at any of this, let me know 😄

apps/apollo-vertex/registry/shell/shell.tsx
```diff
@@ -36,19 +46,33 @@ const ApolloShellComponent: FC<ApolloShellComponentProps> = ({
...
+
+const MockAuthProvider: FC<PropsWithChildren> = ({ children }) => (
+  <AuthContext.Provider value={MOCK_AUTH_CONTEXT}>{children}</AuthContext.Provider>
+);
+
+type ApolloShellProps = ApolloShellComponentProps &
+  (
+    | { bypassAuth: true; clientId?: string; scope?: string; baseUrl?: string }
+    | { bypassAuth?: false; clientId: string; scope: string; baseUrl: string }
+  );
```
alincadariu: do we need to bypass auth? we can add people in the org/app that the auth uses.
i think @ruudandriessen did this already with you @petervachon already 🤔

alincadariu (changes requested):

petervachon:
Shell
- Bypass Auth (for local environment only)
- Setup mock dashboard & menu — Can be removed
- Added “Open standalone preview” to allow users to 
- Light/Dark mode visual expressions updated (includes padding/spacing, colors etc.)
- Default app icon (added two new UiPath svgs to public folder)
- Moved localization and theme under user preferences

Components
Card
- Glass variant

Global colors
Primary 50-900 now available
- Colors now map better for light & dark mode (e.g. Dark Primary = primary-500, Light Primary = primary-600)

Reviewed by Design team 2.26.26

alincadariu (approved):

frankkluijtmans (approved):

PR #276, opened 2026-02-27
feat(apollo-react): loading state for toolbox

uipath-utkarsh:
## Description

Use loading state skeleton when + is clicked on stage node if the APIs are still in flight.

Once the APIs are successful, populate the toolbox. 

 1. Repurposed addTaskLoading prop on StageNode to drive loading state inside the Toolbox instead of showing a spinner on the + button. This is a breaking change
 2. The + button is now always clickable and shows the add icon regardless of addTaskLoading — loading is delegated to the Toolbox's skeleton/disabled-items UI.
 3. Updated the AddTaskLoading story to demonstrate both loading patterns:
 4. Top-level loading: addTaskLoading=true + empty taskOptions → Toolbox shows skeleton loaders until API returns items
 5. Async children loading: ListItem with async children function → Toolbox internally dims level 1 items while fetching level 2 subtasks (no extra props needed)
 6. Removed unused `ApCircularProgress` import from StageNode.tsx
 7. Updated tests to verify the new behavior (loading passed to Toolbox, button no longer disabled)

## Demo

https://github.com/user-attachments/assets/2c2e45f1-6d84-4c25-b6fe-79fed9282efd

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNode.stories.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNode.stories.tsx
@@ -10,7 +10,7 @@ import {
 } from '@uipath/apollo-react/canvas/xyflow/react';
 import { ApButton, ApMenu } from '@uipath/apollo-react/material/components';
 import type { IMenuItem } from '@uipath/apollo-react/material/components/ap-menu/ApMenu.types';
-import { useCallback, useMemo, useState } from 'react';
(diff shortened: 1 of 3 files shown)


ayushya:
Primarily, this PR fixes what we were trying to do.

As a follow-up on this PR, if you also achieve the following, it would save us from creating another ticket later.
In the demo, when you click the Email item, it should immediately jump to email and show a loader within its items section until the data is loaded

ayushya (approved):

PR #286, opened 2026-03-03
feat: Data table updates [AGVSOL-490]

frankkluijtmans:
# Notes

The following features we're added for the data table in order to get it ready for use in our vertical:

- Add faceted filter components
- Add search
- Add column selector with drag and drop support
- Small refactors

## Demo
 
data-table


apps/apollo-vertex/registry/data-table/data-table-column-header.tsx
```diff
@@ -0,0 +1,79 @@
...
+            className="data-[state=open]:bg-accent -ml-3 h-8"
+          >
+            <span>{title}</span>
+            {column.getIsSorted() === "desc" ? (
+              <ArrowDownIcon />
+            ) : column.getIsSorted() === "asc" ? (
+              <ArrowUpIcon />
+            ) : (
+              <ChevronsUpDownIcon />
+            )}
```
0xr3ngar: can't this be extracted into a component and we just do 

const order = column.getIsSorted();

if(order === "desc") return <ArrowDownIcon />
...

apps/apollo-vertex/registry/data-table/data-table-faceted-filter.tsx
```diff
@@ -0,0 +1,128 @@
...
+  noOptionsMessage,
+}: DataTableFacetedFilterProps<TData, TValue>) {
+  const [open, setOpen] = React.useState(false);
+
+  if (!column) return null;
+
+  const filterValue = column.getFilterValue();
+  const selectedValues = new Set<string>(
+    Array.isArray(filterValue) ? filterValue : [],
+  );
```
0xr3ngar: why do we need this set?

apps/apollo-vertex/registry/data-table/data-table-faceted-filter.tsx
```diff
@@ -0,0 +1,128 @@
...
+}
+
+function DataTableFacetedFilter<TData, TValue>({
+  column,
+  title,
+  options,
+  className,
+  noOptionsMessage,
+}: DataTableFacetedFilterProps<TData, TValue>) {
+  const [open, setOpen] = React.useState(false);
```
0xr3ngar: React. ?, just import useState directly

apps/apollo-vertex/registry/data-table/data-table-faceted-filter.tsx
```diff
@@ -0,0 +1,128 @@
...
+  );
+
+  const handleToggle = (value: string) => {
+    const next = new Set(selectedValues);
+    if (next.has(value)) {
+      next.delete(value);
+    } else {
+      next.add(value);
+    }
+    const arr = Array.from(next);
```
0xr3ngar: again, why do we opt for a set?

apps/apollo-vertex/registry/data-table/data-table-view-options.tsx
```diff
@@ -0,0 +1,211 @@
...
+      let reorderIdx = 0;
+      for (const id of allColumnIds) {
+        if (reorderableSet.has(id)) {
+          fullOrder.push(newReorderableOrder[reorderIdx]);
+          reorderIdx++;
+          continue;
+        }
+
+        fullOrder.push(id);
+      }
```
0xr3ngar: why do we need all of this?? can't we just do 
```ts
const newReorderedOrder = arrayMove(table.getAllColumns, oldIndex, newIndex);
```
and then directly set it?
table.setColumnOrder(fullOrder);

`arrayMove` already handles the reordering?

apps/apollo-vertex/registry/use-local-storage/use-local-storage.ts
0xr3ngar, line 1: why do we have our own implementation? can't we just use the react-hooks lib ? 

https://usehooks-ts.com/react-hook/use-local-storage

I don't like that we are making our own local storage hook

apps/apollo-vertex/registry/data-table/data-table.tsx
```diff
@@ -119,11 +272,23 @@ function DataTable<TData, TValue>({
...
                   key={row.id}
                   data-state={row.getIsSelected() && "selected"}
+                  className={
+                    onRowClick
+                      ? "cursor-pointer hover:bg-gradient-to-r hover:from-primary/5 hover:to-secondary/5 transition-all border-l-2 border-transparent hover:border-l-primary"
+                      : ""
+                  }
+                  {...(onRowClick && {
+                    onClick: () => onRowClick(row.original),
+                  })}
```
0xr3ngar, line 213: no need for this spreading, if onRowClick is undefined it will just return void, just do 
`onClick: () => onRowClick?.(row.original),`

apps/apollo-vertex/registry/data-table/data-table.tsx
```diff
@@ -63,43 +127,132 @@ interface DataTableProps<TData, TValue> {
...
     <div data-slot="data-table" className={cn("space-y-4", className)}>
+      {showToolbar && (
+        <DataTableToolbar>
+          <div className="flex items-center gap-2">
+            {enableSearch && <DataTableSearch table={table} />}
+            {toolbarContent?.(table)}
+          </div>
+          {enableViewOptions && <DataTableViewOptions table={table} />}
+        </DataTableToolbar>
+      )}
```
0xr3ngar: all of this and chaining is bad 

just make another component and pass the content like that
```tsx
const searchBar = enableSearch ? <DataTableSearch table={table} /> : null;
const viewOptions = enableViewOptions ? <DataTableViewOptions table={table} /> : null;
const customContent = toolbarContent?.(table);

return (
  <div data-slot="data-table" className={cn("space-y-4", className)}>
    <DataTableToolbar 
      leftContent={<>{searchBar} {customContent}</>}
      rightContent={viewOptions}
    />
    
    <div className="rounded-md border">
....
    </div>
  </div>
);
```

apps/apollo-vertex/registry/data-table/data-table.tsx
```diff
@@ -54,6 +36,88 @@ import {
...
+  // getSnapshot from being recreated every render, which would cause
+  // useSyncExternalStore to re-subscribe and trigger infinite loops
+  // when defaultValue is an object/array.
+  const defaultRef = React.useRef(defaultValue);
+
+  const [storedValue, setStoredValue] = useLocalStorage<T>(
+    `${storageKey ?? "__noop__"}-${subKey}`,
+    defaultRef.current,
+  );
+  const [localValue, setLocalValue] = React.useState<T>(defaultValue);
```
0xr3ngar: why do we have 2 states?, source of truth is local storage?

apps/apollo-vertex/registry/data-table/data-table.tsx
```diff
@@ -54,6 +36,88 @@ import {
...
+    return true;
+  const cellValue = String(row.getValue(columnId) ?? "");
+  return filterValue.includes(cellValue);
+};
+
+// ---------------------------------------------------------------------------
+// Persisted state helper
+// ---------------------------------------------------------------------------
+
+function usePersistedState<T>(
```
0xr3ngar: why is this hook needed?

apps/apollo-vertex/registry/data-table/data-table.tsx
```diff
@@ -63,43 +127,132 @@ interface DataTableProps<TData, TValue> {
...
+  globalFilterFn?: FilterFn<TData>;
+  enableSearch?: boolean;
+  enableViewOptions?: boolean;
+  toolbarContent?: (table: TanstackTable<TData>) => React.ReactNode;
+  noResultsMessage?: string;
+  storageKey?: string;
 }
 
 function DataTable<TData, TValue>({
   columns,
```
0xr3ngar, line 113: Why are we persisting state here when columns are already being passed in as props? We should let the consumer handle the state and just expose onRowClick, onClick, and onReorder handlers 

This would allow us to kill the local state entirely and make this a purely controlled component

apps/apollo-vertex/registry/data-table/data-table-view-options.tsx
```diff
@@ -0,0 +1,211 @@
...
+                    displayName={meta?.displayName ?? column.id}
+                    isVisible={column.getIsVisible()}
+                    onToggleVisibility={(value) =>
+                      column.toggleVisibility(value)
+                    }
+                  />
+                );
+              })}
+            </SortableContext>
+          </DndContext>
```
0xr3ngar: I suggest isolating the DnD functionality in a component

the DropdownMenu shouldn't be cluttered with drag-related state

moving that into a separate component makes the whole thing more modular and easier to maintain down the line

apps/apollo-vertex/registry/data-table/data-table-view-options.tsx
```diff
@@ -0,0 +1,211 @@
...
+// ---------------------------------------------------------------------------
+
+interface SortableColumnItemProps {
+  id: string;
+  displayName: string;
+  isVisible: boolean;
+  onToggleVisibility: (value: boolean) => void;
+}
+
+function SortableColumnItem({
```
0xr3ngar: extract in a component, no need to be in the same file.

apps/apollo-vertex/registry/data-table/data-table-skeleton.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <>
+      {Array.from({ length: rowCount }, (_row, rowIndex) => (
+        <TableRow key={generateSkeletonKey("row", rowIndex)}>
+          {Array.from({ length: columnCount }, (_col, colIndex) => (
+            <TableCell key={generateSkeletonKey("cell", rowIndex, colIndex)}>
+              <Skeleton className="h-4 w-36" />
+            </TableCell>
+          ))}
+        </TableRow>
+      ))}
```
0xr3ngar, line 30: I would personally extract the Array.from's in constants

const rows = Array.from({ length: rowCount });
  const columns = Array.from({ length: columnCount });

and just map over them

makes it easier to read

apps/apollo-vertex/registry/data-table/data-table-search.tsx
```diff
@@ -0,0 +1,29 @@
...
+  table,
+  placeholder,
+  className,
+}: DataTableSearchProps<TData>) {
+  const { t } = useTranslation();
+  return (
+    <div data-slot="data-table-search" className={className}>
+      <Input
+        placeholder={placeholder ?? t("search")}
+        value={String(table.getState().globalFilter ?? "")}
```
0xr3ngar, line 21: why is this called globalFilter?? and not searchTerm, shouldn't we change the state to searchTerm and filter based on that. All this time I thought that globalFilter was something with filters like we have in process mining

apps/apollo-vertex/registry/data-table/data-table-pagination.tsx
```diff
@@ -0,0 +1,115 @@
...
+            value={`${pageSize}`}
+            onValueChange={(value) => {
+              table.setPageSize(Number(value));
+            }}
+          >
+            <SelectTrigger size="sm" className="w-[70px]">
+              <SelectValue placeholder={pageSize} />
+            </SelectTrigger>
+            <SelectContent side="top">
+              {[10, 20, 30, 40, 50].map((size) => (
```
0xr3ngar: shouldn't we let the consumer define this? cuz paginated values can be different

0xr3ngar (changes requested):

apps/apollo-vertex/registry/data-table/data-table-faceted-filter.tsx
frankkluijtmans: I've seen Claude use Sets more often. I'll remove it here since regardless of whether or not we use it, duplicate values would still cause problems and it should simply not happen.

apps/apollo-vertex/registry/use-local-storage/use-local-storage.ts
frankkluijtmans: I think that would work fine, but I'd still like to isolate that change in a separate PR. I'm creating a follow-up ticket for it

apps/apollo-vertex/registry/data-table/data-table-skeleton.tsx
frankkluijtmans: Not entirely in line with your comment since it works a bit differently, but I did clean it up to make it a bit more clear what's actually being done here.

apps/apollo-vertex/registry/data-table/data-table-pagination.tsx
frankkluijtmans: I'd say let's use an overridable default here?

apps/apollo-vertex/registry/data-table/data-table-search.tsx
frankkluijtmans: That's simply an implementation detail of tanstack table 😅

0xr3ngar (approved):

PR #291, opened 2026-03-03
feat(apollo-react): position preview nodes by handle with vertical spreading

david-rios-uipath:
## Summary

Preview nodes now anchor their vertical position to the exact canvas coordinates of the source handle rather than the node center. When a source node has multiple handles on the right side, preview nodes are spread vertically — top-half handles shift upward, bottom-half handles shift down, and the middle handle (odd count) stays centered.

**NOTE: this change only affects node placement -- this does not affect our 'tidy up' function/`layoutAll` logic**
`layoutAll` logic will be updated in a follow-up PR.

#### Before

**Right-facing handles**

https://github.com/user-attachments/assets/f0adfb40-44b2-4bdd-8a9c-c7062364a5f2

**Top/bottom-facing handles**

https://github.com/user-attachments/assets/6a889f0c-4811-4605-a787-e81346d908d6

#### After

**Right-facing handles**

https://github.com/user-attachments/assets/db68c8be-37ea-4284-9511-182d519f3191

**Top/bottom-facing handles**

https://github.com/user-attachments/assets/2c54608a-50c0-45a3-ae9c-35d53b29b679

## Changes

- **`NodeUtils.ts`** — Added `getHandleIndex` utility: finds a handle's 0-based sort index among peers on the same side (top-to-bottom for Left/Right, left-to-right for Top/Bottom). Merged split imports from the same module into one. Refactored `getBoxesFromNodes` to use `.map()`.
- **`createPreviewNode.ts`** — Extended `calculateAutoPosition` with `handleAnchor`, `handleIndex`, and `handleCount` params. Resolves exact handle canvas coordinates via `InternalNode.internals.handleBounds`. Applies y-offset spreading for the `Position.Right` case when multiple handles are present.

## Testing
Will complete all tests before merge.

- [x] `pnpm run test` passes (62 suites, 1201 tests)
- [x] Manual testing performed
    - [x] Correct default vertical position from switch blocks w/ up to 5 connections
    - [x] Correct default horizontal position from top/bottom agent handles


packages/apollo-react/src/canvas/utils/createPreviewNode.ts
```diff
@@ -80,15 +84,19 @@ function calculateAutoPosition(
   previewNodeSize: { width: number; height: number },
   existingNodes: Node[],
   offset = GRID_SPACING * 5,
-  ignoredNodeTypes: string[] = []
+  ignoredNodeTypes: string[] = [],
+  handleAnchor?: { x: number; y: number },
```
david-rios-uipath: I left handle fields optional since I'm unsure if it's safe to assume that we will always find a handle.

Intuitively, I would assume that we are calling `calculateAutoPosition` if a position wasn't given, which _requires_ a reference handle for valid placement. If this is a fair assumption to make, I would prefer to have handle info (`handleAnchor`, `handleIndex`, and `handleCount`) as required params for this function.

CalinaCristian (approved):

snuziale (approved):

PR #292, opened 2026-03-03
fix(apollo-react): use combination search for add node panel [MST-6789]

BenGSchulz:
[CategoryTree.ts:316-388]() — Rewrote filterBySearch() to support combination search:
- Splits the search term into individual words.
- Each word must match at least one searchable attribute (label, nodeType, description, tags, or ancestor category names).
- Uses a custom recursive traversal that passes category names down the tree, so nodes inherit their parent categories as searchable context.

[CategoryTree.test.ts:454-508]() — Added 5 new tests:
- Multi-word matching across different node attributes.
- Multi-word matching across node + category attributes.
- Word order independence.
- All words must match (no partial matches).
- Matching across nested category names.

With this change, the following queries would all work for the "Archive Email" node in "Microsoft Outlook 365" > "Connector":
- "Outlook Email" — "outlook" matches category name, "email" matches node label
- "Connector Email Archive" — "connector" matches category, "email"/"archive" match label
- "Email outlook" — same as first, order-independent

### DIFF
--- a/packages/apollo-react/src/canvas/core/CategoryTree.test.ts
+++ b/packages/apollo-react/src/canvas/core/CategoryTree.test.ts
@@ -450,6 +450,66 @@ describe('CategoryTree', () => {
       const filtered = tree.filterBySearch('nonexistent-search-term');
       expect(filtered.isEmpty()).toBe(true);
     });
+
+    it('should match multi-word search across different node attributes', () => {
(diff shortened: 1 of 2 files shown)


snuziale (approved):
Approving for now, please create a ticket to revisit and update with a trie data structure.

PR #301, opened 2026-03-05
feat(apollo-react): update case stagenode for non-sequential tasks support

joelasher-ap:
This adds support for Non-sequential tasks to the StageNode. We're about to do large UI updates to the StageNode soon and this will be behind a FF until the new BPMN format and the new UI is ready so I'm not too concerned with how exactly it looks right now. It just needs to be functional right now until we get the new UI for this, at which point we can update it before we make it available to users.

These tasks can't be dragged or moved. You can only replace the task or delete it. A task becomes "non-sequential" based on it's properties from the properties panel (for right now at least). If it can only run based on a user trying to run it manually or if a Connector is hit and regardless of the tasks that came before it or what stage we're in, then it becomes non-sequential

Screenshot 2026-03-04 at 9 35 06 PM
Screenshot 2026-03-04 at 9 35 10 PM

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNode.stories.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNode.stories.tsx
@@ -200,6 +200,19 @@ export const Default: Story = {
             slaBreached: false,
             escalation: '1h',
             escalationsTriggered: false,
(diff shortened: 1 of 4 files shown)


joelasher-ap:
Closing this one, revisiting how we're going to do this one

PR #302, opened 2026-03-05
Feat/vertex chat ai [AGVSOL-1350]

pieman1313:
this component is implemented as a poc in 2 use cases:

- https://github.com/UiPath/vertical-solution-invoice-processing/pull/158
- https://github.com/UiPath/vertical-solution-invoice-processing/pull/159

Mar-05-2026 08-18-13


apps/apollo-vertex/registry.json
```diff
@@ -239,6 +239,69 @@
...
+          "path": "registry/ai-chat/ai-chat-input.tsx",
+          "type": "registry:ui"
+        },
+        {
+          "path": "registry/ai-chat/ai-chat-suggestions.tsx",
+          "type": "registry:ui"
+        },
+        {
+          "path": "registry/ai-chat/ai-chat-tool-group.tsx",
+          "type": "registry:ui"
```
VKravchuk: I think it would be good to put all these files at least into `lib` folder
maybe even `lib/ai(module name)/hooks|utils|components(type of file)/*.ts`, but not so sure about it

VKravchuk (changes requested):

apps/apollo-vertex/templates/AiChatTemplate.tsx
```diff
@@ -0,0 +1,350 @@
...
+        `You are a helpful demo assistant. The user is in the "${activeSection}" section.`,
+      fallbackResponse: "Done.",
+    },
+    accessToken: "demo-token",
+    storage: { type: "session", messagesKey: "ai-chat-demo-v2" },
+  });
+
+  const { clearChat, appendMessages } = chat;
+
+  const handleSendMessage = useCallback(
```
ruudandriessen, line 279: any reason why we didn't just make this the `sendMessage` function out of the chat api?

apps/apollo-vertex/templates/AiChatTemplate.tsx
```diff
@@ -0,0 +1,350 @@
...
+} from "@/lib/ai-chat-types";
+import { AiChat } from "@/registry/ai-chat/ai-chat";
+import { useAiChat } from "@/registry/ai-chat/use-ai-chat";
+
+const delay = (ms: number): Promise<void> =>
+  new Promise((r) => {
+    setTimeout(r, ms);
+  });
+
+function getMockResponses(
```
ruudandriessen: mmmm can't we hook in a real agent? this feels a bit shady tbh

apps/apollo-vertex/registry/ai-chat/use-ai-chat.ts
```diff
@@ -0,0 +1,330 @@
...
+      const systemMessage: ChatMessage = {
+        id: crypto.randomUUID(),
+        role: "system",
+        content,
+        timestamp: Date.now(),
+      };
+      appendMessages([systemMessage]);
+    },
+    [appendMessages],
+  );
```
ruudandriessen: why would you ever add a system message? those are the first one and it's a bit weird to add a second one?

apps/apollo-vertex/registry/ai-chat/hooks/use-ai-chat.ts
```diff
@@ -0,0 +1,330 @@
...
+  configRef.current = config;
+  const accessTokenRef = useRef(accessToken);
+  accessTokenRef.current = accessToken;
+  const onToolCallRef = useRef(onToolCall);
+  onToolCallRef.current = onToolCall;
+  const onErrorRef = useRef(onError);
+  onErrorRef.current = onError;
+  const storageRef = useRef<Storage | null>(null);
+  const storageKeyRef = useRef(storageKey);
+  storageKeyRef.current = storageKey;
```
ruudandriessen: this feels very shady

apps/apollo-vertex/registry/ai-chat/hooks/use-ai-chat.ts
```diff
@@ -0,0 +1,330 @@
...
+    onError,
+  } = options;
+
+  const storageType = storageConfig?.type ?? "session";
+  const storageKey = storageConfig?.messagesKey ?? "ai_chat_messages";
+
+  // Initialize empty to avoid SSR/client hydration mismatch; storage is browser-only.
+  const [messages, setMessages] = useState<ChatMessage[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<Error | null>(null);
```
ruudandriessen: ehh we don't really have to deal with ssr

apps/apollo-vertex/registry/ai-chat/ai-chat.tsx
```diff
@@ -0,0 +1,262 @@
...
+    Array.from(container.children).forEach((child) => {
+      resizeObserver.observe(child);
+    });
+
+    return () => {
+      clearTimeout(initialTimer);
+      clearTimeout(scrollTimeout);
+      resizeObserver.disconnect();
+    };
+  }, [messages]);
```
ruudandriessen: why are we doing this with timers? We can just scroll to bottom when a new message arrives instead?

apps/apollo-vertex/registry/ai-chat/ai-chat-utils.ts
```diff
@@ -0,0 +1,176 @@
...
+    }))
+    .find(
+      ({ navigation, index }) =>
+        navigation !== null && !hasUserResponseAfter(messages, index),
+    );
+
+  return result?.navigation ?? null;
+}
+
+export function groupMessages(
```
ruudandriessen: I like the idea of tool grouping - but if we do that, I think we should only group tools after a message is replying - since these tools don't show up all at once like they do now in our mock scenario - they take a bit. And the user is going to want to see those calls most likely.

apps/apollo-vertex/registry/ai-chat/ai-chat-utils.ts
```diff
@@ -0,0 +1,176 @@
...
+  return message.parts?.map((p) => p.content).join("") ?? message.content;
+}
+
+function parseToolResultForChoices(
+  message: ChatMessage,
+): ToolResultChoices | null {
+  if (message.role !== "tool") return null;
+
+  const content = extractMessageContent(message);
+  const parsed = safeJsonParse(content);
```
ruudandriessen: why are we just parsing json like this? I'd expect every tool has a schema that we validate too?

apps/apollo-vertex/registry/ai-chat/ai-chat-utils.ts
```diff
@@ -0,0 +1,176 @@
...
+
+function isValidChoices(parsed: unknown): parsed is ToolResultChoices {
+  return (
+    typeof parsed === "object" &&
+    parsed !== null &&
+    "type" in parsed &&
+    parsed.type === "choices" &&
+    "options" in parsed &&
+    Array.isArray(parsed.options)
+  );
```
ruudandriessen: why? what is this?

apps/apollo-vertex/registry/ai-chat/hooks/use-ai-chat.ts
```diff
@@ -0,0 +1,330 @@
...
+  getAccessToken,
+  resolveConfig,
+} from "@/lib/ai-chat-api";
+import {
+  getStorage,
+  loadFromStorage,
+  saveToStorage,
+} from "@/lib/ai-chat-storage";
+
+export function useAiChat(options: UseAiChatOptions): UseAiChatReturn {
```
ruudandriessen: looks like tools aren't type safe now, we should make sure that there's proper renderers for each of them, like for charts, etc.

apps/apollo-vertex/registry/ai-chat/ai-chat.tsx
```diff
@@ -0,0 +1,262 @@
...
+            )}
+
+            {latestNavigation && !isLoading && onNavigate && (
+              <nav aria-label="Navigation options" className="flex flex-wrap gap-2">
+                {latestNavigation.tabs.map((navTab) => (
+                  <button
+                    key={navTab.tab}
+                    type="button"
+                    className="h-auto py-2 px-3 text-sm rounded-lg border hover:bg-muted transition-colors"
+                    onClick={() => onNavigate(navTab.tab)}
```
ruudandriessen: This feels veeery specific, I don't know if this makes sense - what's the use case here and why did we build it not as a regular tool, but as this "special cookie"?

apps/apollo-vertex/registry/ai-chat/ai-chat.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  const prevLoadingRef = useRef(isLoading);
+  useEffect(() => {
+    if (prevLoadingRef.current && !isLoading) {
+      setTimeout(() => {
+        inputRef.current?.focus();
+      }, 50);
+    }
+    prevLoadingRef.current = isLoading;
+  }, [isLoading]);
```
ruudandriessen: why are we doing this odd autofocus thing?

apps/apollo-vertex/registry/ai-chat/ai-chat.tsx
```diff
@@ -0,0 +1,262 @@
...
+    };
+  }, [messages]);
+
+  const handleSubmit = (files?: File[]) => {
+    if ((!input.trim() && !files) || isLoading) return;
+    onSendMessage(input, files);
+    setInput("");
+  };
+
+  const prevLoadingRef = useRef(isLoading);
```
ruudandriessen: this previous loading state is pretty odd

apps/apollo-vertex/registry/ai-chat/ai-chat.tsx
ruudandriessen: another scroll to bottom? Why do we have two? same question here as well, just scroll to bottom upon sending a message?

apps/apollo-vertex/registry/ai-chat/ai-chat-types.ts
```diff
@@ -0,0 +1,132 @@
...
+  accessToken: string | (() => string | null);
+  storage?: StorageConfig;
+  onToolCall?: (
+    toolCall: ToolCall,
+    args: Record<string, unknown>,
+  ) => Promise<unknown> | unknown;
+  onError?: (error: Error) => void;
+}
+
+export interface UseAiChatReturn {
```
ruudandriessen: no generics so none of this seems type safe

apps/apollo-vertex/registry/ai-chat/ai-chat-types.ts
```diff
@@ -0,0 +1,132 @@
...
+  type: "function";
+  function: {
+    name: string;
+    description: string;
+    parameters: {
+      type: "object";
+      properties: Record<string, unknown>;
+      required?: string[];
+    };
+  };
```
ruudandriessen: please allow passing standard schema and not json schema's

ruudandriessen: https://standardschema.dev/

apps/apollo-vertex/registry/ai-chat/ai-chat-types.ts
```diff
@@ -0,0 +1,132 @@
...
+  parts?: MessagePart[];
+  timestamp: number;
+  toolCallId?: string;
+  toolCalls?: ToolCall[];
+  hidden?: boolean;
+  attachments?: Array<{
+    fileName: string;
+    fileType: string;
+    fileSize: number;
+  }>;
```
ruudandriessen: don't go all out so quickly- drop attachments until we need them

apps/apollo-vertex/registry/ai-chat/ai-chat-types.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export interface ChatMessage {
+  id: string;
+  role: ChatRole;
+  content: string;
+  parts?: MessagePart[];
+  timestamp: number;
+  toolCallId?: string;
+  toolCalls?: ToolCall[];
+  hidden?: boolean;
```
ruudandriessen: why do we have hidden messages? 😕

apps/apollo-vertex/registry/ai-chat/ai-chat-types.ts
```diff
@@ -0,0 +1,132 @@
...
+}
+
+export interface ToolResultNavigation {
+  type: "navigation";
+  tabs: NavigationTab[];
+}
+
+export type ToolResult = ToolResultChoices | ToolResultNavigation;
+
+export interface ChatMessage {
```
ruudandriessen: don't make things optional if you know they only exist on certain types, i.e. if chat is of type tool, make sure that `toolCalllId` is defined, etc.. so if we then check `role === Tool` then we get correct types

apps/apollo-vertex/registry/ai-chat/ai-chat-tool-group-message.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import { Sparkles } from "lucide-react";
+import type { ToolCall } from "@/lib/ai-chat-types";
+import { AiChatToolGroup } from "./ai-chat-tool-group";
+
+interface AiChatToolGroupMessageProps {
+  toolCalls: ToolCall[];
+  isLatest: boolean;
+  assistantName: string;
+  toolDisplayNames?: Record<string, string>;
```
ruudandriessen: If we have a nicer interface for the response (which we can map in our hook) this interface can become much nicer

apps/apollo-vertex/registry/ai-chat/utils/ai-chat-storage.ts
```diff
@@ -0,0 +1,36 @@
+export function getStorage(type: "session" | "local" | "none"): Storage | null {
```
ruudandriessen: shouldn't we just choose one of the two? why do we support both?

apps/apollo-vertex/registry/ai-chat/ai-chat-message.tsx
```diff
@@ -0,0 +1,184 @@
...
+import { useTranslation } from "react-i18next";
+import ReactMarkdown from "react-markdown";
+import remarkGfm from "remark-gfm";
+import { cn } from "@/lib/utils";
+import type { ChatMessage } from "@/lib/ai-chat-types";
+import { extractMessageContent } from "@/lib/ai-chat-utils";
+
+interface AiChatMessageProps {
+  message: ChatMessage;
+  renderToolCall?: (toolCall: ChatMessage["toolCalls"]) => ReactNode;
```
ruudandriessen: can we make this part of the tool definition?

apps/apollo-vertex/registry/ai-chat/ai-chat-input.tsx
```diff
@@ -0,0 +1,320 @@
...
+          }
+          return file.type === type;
+        });
+
+        if (!isAccepted) {
+          return t("file_type_not_accepted", { name: file.name });
+        }
+      }
+
+      return null;
```
ruudandriessen: don't deal with files unless we have to

ruudandriessen: did we have a user case where it was needed?

ruudandriessen (dismissed):

alincadariu (changes requested):
- we should wrap it in the locale provider for the example

Image

- shouldn't we add the useAiChat hook inside the component? exposing what we need as a prop? 🤔

apps/apollo-vertex/templates/AiChatTemplate.tsx
pieman1313: tried but we`d need a api key for a direct one, or if we're using agent hub, we need an org / tenant
ended up hardcoding this instead

apps/apollo-vertex/registry/ai-chat/hooks/use-ai-chat.ts
pieman1313: had to add this cause of the nextjs server hosting the examples
it constantly had a hydration error
will avoid it

apps/apollo-vertex/registry/ai-chat/ai-chat.tsx
pieman1313: there were multiple bugs from the 2 pic integrations where scroll to bottom wasn't enough when a chart was rendered and it finished loading later and the scroll didn't account for it, and the second bug was in the second pocket but I don't remember it off the top of my head
will take another look

apps/apollo-vertex/registry/ai-chat/ai-chat.tsx
pieman1313: yeah, came from the demo summit poc, they have a auto navigate on tool response feature

apps/apollo-vertex/registry/ai-chat/ai-chat-input.tsx
pieman1313: yeah, I saw it in Bulats vertical

apps/apollo-vertex/registry/ai-chat/ai-chat-types.ts
pieman1313: in Jaspers poc they use it to trigger an agent automatically when opening an invoice, or after processing an invoice, take "automatic" action by prompting the llm

pieman1313:
> * shouldn't we add the useAiChat hook inside the component? exposing what we need as a prop?

if we did that, people couldn't add their custom logic like in this example integration [pr](https://github.com/UiPath/vertical-solution-invoice-processing/pull/158/changes#diff-10af32625bfed8762c0eead4f4c1abd9f784b8267a1d483955d0cb4a728846a6) and the api we would need to introduce at the component level to support this turns into a mess fast

apps/apollo-vertex/registry/ai-chat/hooks/use-ai-chat.ts
```diff
@@ -88,8 +94,8 @@ export function useAiChat(options: UseAiChatOptions): UseAiChatReturn {
     try {
       await runAgentLoop(
         initialMessages,
-        connection,
-        options,
+        provider,
+        { tools: options.tools, maxToolIterations: options.maxToolIterations },
```
VKravchuk: I see that inside `runAgentLoop` we call `resolveConfig` one more time. looks like we can pass resolved tools here

apps/apollo-vertex/registry/ai-chat/metadata.json
```diff
@@ -0,0 +1,98 @@
+{
+  "name": "ai-chat",
+  "type": "component",
+  "category": "complex",
```
alincadariu: do we need this file?

apps/apollo-vertex/registry/ai-chat/README.md
```diff
@@ -0,0 +1,503 @@
+# AI Chat Component
+
+A generic, reusable, and type-safe AI chat component for the Apollo Design System. Built with React, TypeScript, Tailwind CSS, and Zod. Supports OpenAI-compatible LLM APIs (OpenAI, Azure OpenAI, Ollama) with a pluggable `LLMProvider` interface, with an agentic tool loop, Zod-typed tool calling, markdown rendering, collapsible tool groups, suggestion buttons, and error display.
```
alincadariu: we shouldn't need this file, we have the docs page.

alincadariu (changes requested):

apps/apollo-vertex/registry/ai-chat/metadata.json
pieman1313: image

apps/apollo-vertex/registry/ai-chat/utils/ai-chat-utils.ts
```diff
@@ -0,0 +1,120 @@
...
+function isValidChoices(parsed: unknown): parsed is ToolResultChoices {
+  return (
+    typeof parsed === "object" &&
+    parsed !== null &&
+    "type" in parsed &&
+    parsed.type === "choices" &&
+    "options" in parsed &&
+    Array.isArray(parsed.options)
+  );
+}
```
alincadariu: can't we use schemas for validation?

pieman1313: great catch, done

alincadariu (approved):

angeloaltamiranom (approved):

VKravchuk (approved):

PR #304, opened 2026-03-05
chore(components): update data-table from UiPath/fins-vertical-solution

alincadariu:
## Summary
Pulls changes to the `data-table` component from `UiPath/fins-vertical-solution`.

This PR was created from the Adoption Tracker dashboard.

### DIFF
--- a/apps/apollo-vertex/registry/data-table/data-table.tsx
+++ b/apps/apollo-vertex/registry/data-table/data-table.tsx
@@ -1,35 +1,48 @@
-"use client";
-
 import {
+  type Column,
   type ColumnDef,
   type ColumnFiltersState,
-  type FilterFn,
   flexRender,
   getCoreRowModel,
   getFilteredRowModel,
   getPaginationRowModel,
   getSortedRowModel,
-  type OnChangeFn,
-  type PaginationState,
-  type Row,
-  type RowData,
-  type RowSelectionState,
   type SortingState,
   type Table as TanstackTable,
   useReactTable,
   type VisibilityState,
 } from "@tanstack/react-table";
+import {
+  ArrowDownIcon,
+  ArrowUpIcon,
+  ChevronLeftIcon,
+  ChevronRightIcon,
+  ChevronsLeftIcon,
+  ChevronsRightIcon,
+  ChevronsUpDownIcon,
+  EyeOffIcon,
+  Settings2Icon,
+} from "lucide-react";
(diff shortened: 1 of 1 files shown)


tarasiuk-o:
Just for reference, here are the changes I made in one of the solutions to the [data-table](https://github.com/UiPath/lmcu-vertical-solution/blob/262a1769c88932b5c8578ddd3e2f959e73674c15/frontend/webapp/src/components/ui/data-table.tsx). I’ll port the same changes to another solution (to `UiPath/fins-vertical-solution`).

The update adds the following new props:

- loading
- scrollable
- onRowClick
- expandedRowContent

[table](https://github.com/UiPath/lmcu-vertical-solution/blob/262a1769c88932b5c8578ddd3e2f959e73674c15/frontend/webapp/src/components/ui/table.tsx#L9) 

- containerClassName

--- 

Also, it would be useful if the `data-table` header provided an expand/collapse all rows control when `expandedRowContent` is present. This could either be implemented as a separate option or enabled automatically whenever the `data-table` includes` expandedRowContent`.
Maybe we don’t need to remove any of the existing options—it’s probably better to keep them and simply extend the component with what’s missing. @alincadariu WDYT?

alincadariu:
> Just for reference, here are the changes I made in one of the solutions to the [data-table](https://github.com/UiPath/lmcu-vertical-solution/blob/262a1769c88932b5c8578ddd3e2f959e73674c15/frontend/webapp/src/components/ui/data-table.tsx). I’ll port the same changes to another solution (to `UiPath/fins-vertical-solution`).
> 
> The update adds the following new props:
> 
> * loading
> * scrollable
> * onRowClick
> * expandedRowContent
> 
> [table](https://github.com/UiPath/lmcu-vertical-solution/blob/262a1769c88932b5c8578ddd3e2f959e73674c15/frontend/webapp/src/components/ui/table.tsx#L9)
> 
> * containerClassName
> 
> Also, it would be useful if the `data-table` header provided an expand/collapse all rows control when `expandedRowContent` is present. This could either be implemented as a separate option or enabled automatically whenever the `data-table` includes` expandedRowContent`. Maybe we don’t need to remove any of the existing options—it’s probably better to keep them and simply extend the component with what’s missing. @alincadariu WDYT?

@tarasiuk-o makes sense, I'll close this PR and check the lmcu and try to adapt the core based on that 👌🏻

PR #311, opened 2026-03-06
feat: add data table hooks [AGVSOL-551, AGVSOL-1525]

frankkluijtmans:
# Notes

This adds 2 hooks:

- useDataTable

Supports the current Data Table implementation and enables the consumer to pass in their own row data.
- useEntityDataTable

Same as above but with built in data fetching based on a specific collection id.


apps/apollo-vertex/registry/use-data-table/useEntityColumns.tsx
```diff
@@ -0,0 +1,130 @@
+import type { ColumnDef } from "@tanstack/react-table";
+
+import { DataTableColumnHeader } from "@/components/ui/data-table";
+import { renderValueOrEmptyState } from "@/lib/renderValueOrEmptyState";
```
ruudandriessen, line 5: you can't use things outside of the registry in the registry - then we just get errors when they are installed.

ruudandriessen (dismissed):

apps/apollo-vertex/registry/use-data-table/useEntityColumns.tsx
frankkluijtmans: These dependencies should be clearly defined in the `registry.json`. Fixed ✅

apps/apollo-vertex/lib/constants.ts
```diff
@@ -0,0 +1,19 @@
...
+};
+
+export const DEFAULT_DATETIME_FORMAT_OPTIONS: DateFormatOptions = {
+  month: "short",
+  day: "numeric",
+  year: "numeric",
+  hour: "2-digit",
+  minute: "2-digit",
+  hourCycle: "h23",
+};
```
0xr3ngar, line 28: I'm questioning if we should just move to temporal ( the new datetime api for JS ) 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal

Only thing I'm worried about is that it is not yet in safari, maybe we should make a ticket for a followup 👀 

Image

apps/apollo-vertex/registry/use-data-table/useEntityDataTable.tsx
```diff
@@ -0,0 +1,128 @@
...
+        )
+      : allColumns.slice(0, initialVisibleColumnCount).map((col) => col.key);
+
+  const { columnVisibility, onColumnVisibilityChange } = useColumnVisibility({
+    storageKey,
+    allColumnKeys,
+    defaultVisibleColumns,
+  });
+
+  const { sorting, onSortingChange } = usePersistedSorting({ storageKey });
```
0xr3ngar, line 83: useEntityDataTable calls useSolution() on line 49, then conditionally throws before calling useLiveQuery, useColumnVisibility, usePersistedSorting, useMemo, usePersistedColumnOrder, and multiple useState calls.

If useSolution() returns null on a re-render after previously returning a value, React detects fewer hooks were called than expected, triggering a Rules of Hooks violation error that masks the intended error message 

all hooks need to be called unconditionally before any early throws

apps/apollo-vertex/registry/use-data-table/useEntityDataTable.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  const defaultColumnOrder = useMemo(
```
0xr3ngar: why do we need this memoized? we are on react 19

apps/apollo-vertex/registry/use-data-table/usePersistedColumnOrder.ts
```diff
@@ -0,0 +1,30 @@
...
+  const [columnOrder, setColumnOrder] = useLocalStorage<string[]>(
+    `${STORAGE_PREFIX}order-${storageKey}`,
+    defaultColumnOrder,
+  );
+
+  const onColumnOrderChange: OnChangeFn<string[]> = (updaterOrValue) => {
+    const newOrder =
+      typeof updaterOrValue === "function"
+        ? updaterOrValue(columnOrder)
+        : updaterOrValue;
```
0xr3ngar, line 26: isn't this always a function 🤔

apps/apollo-vertex/registry/use-data-table/usePersistedSorting.ts
```diff
@@ -0,0 +1,28 @@
...
+  const [sorting, setSorting] = useLocalStorage<SortingState>(
+    `${STORAGE_PREFIX}sorting-${storageKey}`,
+    [],
+  );
+
+  const onSortingChange: OnChangeFn<SortingState> = (updaterOrValue) => {
+    const newSorting =
+      typeof updaterOrValue === "function"
+        ? updaterOrValue(sorting)
+        : updaterOrValue;
```
0xr3ngar, line 24: isn't this always a function 🤔

apps/apollo-vertex/registry/use-data-table/usePersistedColumns.ts
```diff
@@ -0,0 +1,47 @@
...
+  const stored = localStorage.getItem(key);
+  if (stored) {
+    const parsed = JSON.parse(stored);
+    if (Array.isArray(parsed) && parsed.length > 0) {
+      return parsed;
+    }
+  }
+
+  return null;
+}
```
0xr3ngar: use zod here? for parsing?

apps/apollo-vertex/registry/use-data-table/usePersistedColumns.ts
```diff
@@ -0,0 +1,47 @@
...
+  );
+
+  const visibleColumns = availableColumns
+    ? rawColumns.filter((key) => availableColumns.includes(key))
+    : rawColumns;
+
+  const setVisibleColumns = (columns: string[]) => {
+    setRawColumns(columns);
+    localStorage.setItem(fullKey, JSON.stringify(columns));
+  };
```
0xr3ngar: why are we using rawColumns state? and then setting the values in localstorage, shouldn't we use localStorage as the source of truth and no state duplication?

apps/apollo-vertex/registry/use-data-table/useEntityDataTable.tsx
```diff
@@ -0,0 +1,128 @@
...
+    () => allColumns.map((col) => col.key),
+    [allColumns],
+  );
+
+  const { columnOrder, onColumnOrderChange } = usePersistedColumnOrder({
+    storageKey,
+    defaultColumnOrder,
+  });
+
+  const [columnFilters, setColumnFilters] = useState<ColumnFiltersState>([]);
```
0xr3ngar: shouldn't this be in localStorage 🤔

apps/apollo-vertex/registry/use-data-table/useEntityColumns.tsx
```diff
@@ -0,0 +1,126 @@
...
+      ? [
+          ...extraColumns
+            .filter((ec) => ec.position === "start")
+            .map((ec) => ec.column),
+          ...orderedColumns,
+          ...extraColumns
+            .filter((ec) => ec.position === "end")
+            .map((ec) => ec.column),
+        ]
+      : orderedColumns;
```
0xr3ngar: wow this is super messy

maybe just do 
```ts
const byPosition = (pos: "start" | "end") =>
  (extraColumns ?? [])
    .filter((ec) => ec.position === pos)
    .map((ec) => ec.column);

const startColumns = byPosition("start");
const endColumns = byPosition("end");

const allColumns =
  extraColumns?.length ?? 0 > 0
    ? [...startColumns(extraColumns), ...orderedColumns, ...endColumns(extraColumns)]
    : orderedColumns;
```

apps/apollo-vertex/registry/use-data-table/useEntityColumns.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const baseColumns = visibleFields.map((field) => ({
+    key: field.name,
+    label: field.displayName,
+  }));
+
+  const orderedColumns: Column[] =
+    columnOrderProp && columnOrderProp.length > 0
+      ? (() => {
+          const baseMap = new Map(baseColumns.map((col) => [col.key, col]));
```
0xr3ngar: why do we need this map?

apps/apollo-vertex/registry/use-data-table/useEntityColumns.tsx
```diff
@@ -0,0 +1,126 @@
...
+  EntityRecord,
+  ExtraColumn,
+  VssEntity,
+} from "./types";
+
+const formatTypeMap: Record<string, "date" | "datetime" | "number"> = {
+  DATE: "date",
+  DATETIME_WITH_TZ: "datetime",
+  DECIMAL: "number",
+};
```
0xr3ngar: shouldn't this be in constants?

apps/apollo-vertex/registry/use-data-table/usePersistedColumnOrder.ts
```diff
@@ -0,0 +1,30 @@
+import type { OnChangeFn } from "@tanstack/react-table";
+
+import { useLocalStorage } from "@/registry/use-local-storage/use-local-storage";
+
+const STORAGE_PREFIX = "entity-table-";
```
0xr3ngar: move to constants

apps/apollo-vertex/registry/use-data-table/usePersistedColumns.ts
```diff
@@ -0,0 +1,47 @@
+import { useState } from "react";
+
+const STORAGE_KEY_PREFIX = "entity-table-columns-";
```
0xr3ngar: move to constants

apps/apollo-vertex/registry/use-data-table/usePersistedColumns.ts
```diff
@@ -0,0 +1,47 @@
...
+
+function getStoredColumns(key: string): string[] | null {
+  if (typeof window === "undefined") {
+    return null;
+  }
+  const stored = localStorage.getItem(key);
```
0xr3ngar: why do we directly call localStorage here? shouldn't we use the state and see if it includes the key?

0xr3ngar (changes requested):

apps/apollo-vertex/lib/constants.ts
frankkluijtmans: I wouldn't include this in the scope of this story. Lack of full support in Safari tells me it's better to wait with doing this.

apps/apollo-vertex/registry/use-data-table/useEntityDataTable.tsx
frankkluijtmans: That's a good point! 👌 I'll just remove this since we should get a clear error when the collection is missing anyway.

apps/apollo-vertex/registry/use-data-table/usePersistedColumnOrder.ts
frankkluijtmans: Apparently not `type Updater<T> = T | ((old: T) => T)`

apps/apollo-vertex/registry/use-data-table/usePersistedSorting.ts
frankkluijtmans: Same as above.

apps/apollo-vertex/registry/use-data-table/useEntityDataTable.tsx
frankkluijtmans: Nah, this is related to the search bar which filters column values. We don't want to persist that. I realize the naming doesn't make that clear so I'll change this naming.

alincadariu (commented):
shouldn't we add some docs for these as well? 🤔

apps/apollo-vertex/registry.json
```diff
@@ -523,13 +523,13 @@
...
-        "select",
-        "input",
-        "skeleton"
+        "@uipath/table",
+        "@uipath/button",
+        "@uipath/checkbox",
+        "@uipath/dropdown-menu",
+        "@uipath/select",
+        "@uipath/input",
+        "@uipath/skeleton"
```
0xr3ngar: huh why did this change happen?
other regsitryDependencies do not use the `@uipath/` prefix 🤔

0xr3ngar (changes requested):

apps/apollo-vertex/registry.json
frankkluijtmans: I'll follow up on slack, dependency don't resolve causing tests to fail

0xr3ngar (approved):

PR #314, opened 2026-03-09
feat(apollo-react): BaseCanvas - add flag for toggling back navigation

david-rios-uipath:
## Why do we need this flag?
The `usePreventBackNavigation` hook that's used in `apollo-react/BaseCanvas` prevents horizontal '`wheel`' scroll events globally ([source](https://github.com/UiPath/apollo-ui/blob/728a2936566b07f6926e489d4d17b919542cbcb4/packages/apollo-react/src/canvas/components/BaseCanvas/usePreventBackNavigation.ts#L17)). It's intended to prevent accidental back navigation via gesture.

However, this hook adds a global '`wheel`' event listener that intercepts and stops any horizontal scroll events. This is preventing horizontal scrolling from being implemented naturally in flow, and affects other consumers the same way.

I would prefer to delete this hook altogether, but there may be other consumers that depend on this functionality. In the meantime, we can add a flag (`enableBackNavigationPrevention`) to disable this hook's functionality. This gives us the ability to disable this functionality in Flow while not affecting other consumers until we're able to delete this hook in its entirety.

### If we delete this hook, how do we prevent accidental gesture back navigation?
The same end result can be achieved by setting the [`overscroll-behavior-x`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overscroll-behavior-x) property to `contain` on the desired scroll element (e.g. `ScrollArea`, `BaseCanvas`, etc., or at page level, `html`)

**At the moment, I'm questioning the necessity of this hook, since even without it I'm unable to navigate backward from within `BaseCanvas`.**

### DIFF
--- a/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
+++ b/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
@@ -82,6 +82,7 @@ const BaseCanvasInnerComponent = <NodeType extends Node = Node, EdgeType extends
     zoomOnDoubleClick = true,
     snapToGrid = BASE_CANVAS_DEFAULTS.snapToGrid,
     snapGrid = BASE_CANVAS_DEFAULTS.snapGrid,
+    preventScrolling = false,
 
     // Layout
     initialAutoLayout,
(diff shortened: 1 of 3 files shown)


david-rios-uipath:
Will be removing hook -- no flag needed (see https://github.com/UiPath/apollo-ui/pull/318)

PR #318, opened 2026-03-10
fix(apollo-react): remove back navigation event handler

david-rios-uipath:
The `usePreventBackNavigation` hook in `apollo-react/BaseCanvas` prevents horizontal '`wheel`' scroll events globally ([source](https://github.com/UiPath/apollo-ui/blob/728a2936566b07f6926e489d4d17b919542cbcb4/packages/apollo-react/src/canvas/components/BaseCanvas/usePreventBackNavigation.ts#L17)). It's intended to prevent accidental back navigation via gesture.

On further testing, the canvas component handles scroll infinitely _(or at least for a very long finite distance)_, and thus adding a page-level scroll event lock is not needed for the purposes of preventing gestures while using the canvas. More context in [this Slack thread](https://uipath-product.slack.com/archives/C0AH25MT3L5/p1773081374050689).

**Test plan**
Test plan detailed in this Slack [thread](https://uipath-product.slack.com/archives/C0AH25MT3L5/p1773160102743319?thread_ts=1773081374.050689&cid=C0AH25MT3L5) completed successfully. No gesture behavior in relevant usages was changed after removing the hook (tested against staging)


packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
```diff
@@ -17,7 +17,6 @@ import type { BaseCanvasProps, BaseCanvasRef } from './BaseCanvas.types';
 import { CanvasBackground } from './CanvasBackground';
 import { CanvasProviders } from './CanvasProviders';
 import { PanShortcutTeachingUI } from './PanShortcutTeachingUI';
-import { usePreventBackNavigation } from './usePreventBackNavigation';
```
snuziale, line 20: i think the reason this was added, was on a mac in safari .. it was very easy to do a browser back gesture on a trackpad when on the canvas -- can you have a look at that

snuziale: even in chrome -- 2 finger swipe..

david-rios-uipath: yup, this is what I tested -- the hook made no visible difference on my machine in chrome. Was able to swipe backward; will take a look @ Safari

https://github.com/user-attachments/assets/d13ec3df-2be0-4f2f-9092-0ed4b3f9cf7c

david-rios-uipath: Same result on Safari (in alpha w/ hook removed, using staging in the next tab) 

https://github.com/user-attachments/assets/342ca15b-7eed-4c28-8efa-df879bb1aa9e

david-rios-uipath: This does happen really easily on scroll containers w/ horizontals scrolling in flow, which is why I added `overscroll-x-contain` to `ScrollArea` ([src](https://github.com/UiPath/apollo-ui/pull/315/changes#diff-d136c0c79684c2a99ae5be92ef0e7fbd4e505adffbe9c7e791a32af21077e9beR14)).

If we find ourselves having the same issue w/ canvas, we can use that css prop there instead.

snuziale (approved):

PR #322, opened 2026-03-11
fix(apollo-vertex): use-local-storage import path in shell components [AGVSOL-1672]

KokoMilev:


alincadariu (changes requested):
we don't need a separate hooks folder.
check how we do it for `use-mobile`

apps/apollo-vertex/registry.json
```diff
@@ -750,7 +745,7 @@
         "button",
         "tooltip",
         "avatar",
-        "@uipath/use-local-storage",
+        "use-local-storage",
```
alincadariu: we still need @uipath 👍🏻

KokoMilev: Why? Use-mobile doesn't have it

KokoMilev: Nvm I got it. Thanks for mentioning it.

alincadariu: @KokoMilev nice. yea, when components are not part of shadcn basic we need to add @uipath as its our custom one.

alincadariu (approved):

PR #334, opened 2026-03-12
feat(apollo-vertex): add expand row support to DataTable [AGVSOL-1577]

alincadariu:
Screenshot 2026-03-12 at 11 04 05

## Summary

DataTable now supports row expansion via optional `expanded`, `onExpandedChange`, and `renderExpandedRow` props. Consumers define their own expand toggle column (similar to the select column pattern) and provide a render callback for expanded content. The feature is fully opt-in and follows the same controlled-state pattern as selection and sorting.

## Changes

- Added `ExpandedState` and `getExpandedRowModel` from TanStack Table
- New optional props: `expanded`, `onExpandedChange`, `renderExpandedRow`
- Renders expanded content as a full-width row after expanded rows
- Updated DataTableTemplate with example expand column

🤖 Generated with Claude Code

### DIFF
--- a/apps/apollo-vertex/registry/data-table/data-table.tsx
+++ b/apps/apollo-vertex/registry/data-table/data-table.tsx
@@ -3,9 +3,11 @@
 import {
   type ColumnDef,
   type ColumnFiltersState,
+  type ExpandedState,
   type FilterFn,
   flexRender,
   getCoreRowModel,
+  getExpandedRowModel,
   getFilteredRowModel,
   getPaginationRowModel,
   getSortedRowModel,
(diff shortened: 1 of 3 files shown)


apps/apollo-vertex/registry/data-table/data-table.tsx
```diff
@@ -209,27 +222,35 @@ function DataTable<TData, TValue>({
...
-                  {row.getVisibleCells().map((cell) => (
-                    <TableCell key={cell.id}>
-                      {flexRender(
-                        cell.column.columnDef.cell,
-                        cell.getContext(),
-                      )}
-                    </TableCell>
-                  ))}
-                </TableRow>
+                <React.Fragment key={row.id}>
```
frankkluijtmans: Let's just use `<></>` for fragments

frankkluijtmans (changes requested):
Just one small thing

frankkluijtmans (approved):

PR #338, opened 2026-03-12
feat(apollo-vertex): shell configurable nav items [AGVSOL-1718]

KokoMilev:
The shell registry components (`ApolloShell`, `ShellLayout`, `Sidebar`) had hardcoded nav items, making them unusable for the 7+ repos that consume them because each repo has different navigation routes. Additionally, clicking any nav item in the preview navigated away from the shell entirely, showing a 404 without the sidebar.


apps/apollo-vertex/templates/ShellTemplate.tsx
```diff
@@ -1,15 +1,36 @@
...
+  </>
+);
+
+const minimalNavItems = (
+  <>
+    <MinimalNavItem to="/preview/shell-minimal" label="dashboard" />
+    <MinimalNavItem to="/preview/shell-minimal/projects" label="projects" />
+    <MinimalNavItem to="/preview/shell-minimal/analytics" label="analytics" />
+  </>
+);
```
0xr3ngar: why do we pass a react node, when we should pass navItems: Array<{ path: string; label: TranslationKey}>

and then we just do .map() render inside of the shell

0xr3ngar (changes requested):

apps/apollo-vertex/templates/ShellTemplate.tsx
KokoMilev: True, my bad.

0xr3ngar (approved):

apps/apollo-vertex/app/(preview)/preview/shell/page.tsx
```diff
@@ -1,12 +1,20 @@
...
+        <h1 className="text-2xl font-semibold">Page not found</h1>
+        <p className="text-muted-foreground">
+          The page you're looking for doesn't exist.
+        </p>
+        <Link
+          href="/preview/shell/dashboard"
+          className="inline-block text-sm text-primary hover:underline"
+        >
+          Go back to dashboard
+        </Link>
```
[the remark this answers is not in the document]
0xr3ngar: @KokoMilev clanka is right

PR #340, opened 2026-03-12
feat(apollo-vertex): add pointer cursor for all clickable elements

sdignum:
Adding back in the pointer cursor to clickable elements (didn't do it for dropdown and select menus because seemed like a bigger override) because users have come to expect this visual cue when hovering clickable elements.

### DIFF
--- a/apps/apollo-vertex/app/globals.css
+++ b/apps/apollo-vertex/app/globals.css
@@ -40,4 +40,16 @@
   a {
     @apply text-primary-700 dark:text-primary-400;
   }
+
+  /* Ensure all clickable elements show pointer cursor */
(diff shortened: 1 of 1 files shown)


apps/apollo-vertex/app/globals.css
0xr3ngar, line 1: Shouldn't this fix also live in the verticals template (the quick-start base for new verticals)? https://github.com/UiPath/vertical-solutions-template/blob/main/client/src/index.css

This globals.css is scoped to apollo-vertex / the Nextra docs themes - verticals bootstrapped from the template would have the same Tailwind preflight issue and wouldn't get this fix

PR #354, opened 2026-03-16
feat(apollo-react): add keyboard navigation and focus lock to Toolbox [AG-801]

david-rios-uipath:
## Summary

Adds full keyboard navigation and ARIA-compliant focus locking to the Toolbox (AddNodePanel). Users can:
- navigate items vertically with up and down arrow keys (or tab/shift-tab)
- select with `Enter`
- drill into/out of categories with left and right arrow keys
- jump to first/last with `Home`/`End`

The search input acts as a combobox with `aria-activedescendant` pointing into the listbox, satisfying the WAI-ARIA spec ([condition 3](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_focus_activedescendant:~:text=The%20focused%20referencing%20element%20has%20role,a%20descendant%20of%20the%20controlled%20element.)). Also refactors `renderedItems` ownership into ListView via `useImperativeHandle`, so Toolbox reads it from a ref instead of recomputing it.

https://github.com/user-attachments/assets/3737b90c-36c2-434d-aa15-a6d862a4156c

## Changes

- Add `onNavigationKeyDown` and `activeDescendantId` to `SearchBox` input
- Add `role="combobox"`, `aria-controls`, and `aria-expanded` to search input for ARIA compliance
- Add `id="toolbox-listbox"` to the virtualized list
- Add keyboard handler in `Toolbox` for ArrowDown/Up/Left/Right, Enter, Home, End
- Move `renderedItems` computation into `ListView`, expose via `useImperativeHandle`
- Add `ListViewHandle` interface and `enableSections` prop to `ListView`
- Add comprehensive keyboard navigation tests to `Toolbox.test.tsx`

## Flow

```mermaid
flowchart TD
    A[SearchBox input] -->|onKeyDown| B[Toolbox handleNavigationKeyDown]
    B -->|ArrowDown/Up| C[navigateToIndex]
    B -->|Enter/ArrowRight| D[handleItemSelect]
    B -->|ArrowLeft| E[handleBackTransition]
    C -->|scrollToRow| F[ListView ref]
    F -->|useImperativeHandle| G[renderedItems]
    A -->|aria-activedescendant| H[toolbox-item-id]
    A -->|aria-controls| I[toolbox-listbox]
```

## Testing

- [x] `pnpm run test` passes (64 suites, 1254 tests)
- [x] Manual testing
    - [x] Left/right arrows should navigate out of parents/into children respectively
    - [x] Navigate vertically and horizontally, then attempt typing (focus should still be on search bar). Repeat in reverse order
    - [x] Navigate with shift + tab/tab and then up/down arrow keys, sequentially
    - [x] Ensure client scrolls when selecting an element outside of current content area 
- [x] Try to tab out of `AddNodePanel`. Focus should remain within component
- [x] Other `Toolbox` consumers
   - [x] Ensure `StageNode`'s `FloatingCanvasPanel` functions as expected

https://github.com/user-attachments/assets/239412ff-4321-4677-992d-7586c751c6c6

- [x] Unit tests added/updated where relevant

### DIFF
--- a/packages/apollo-react/src/canvas/components/AddNodePanel/AddNodePanel.tsx
+++ b/packages/apollo-react/src/canvas/components/AddNodePanel/AddNodePanel.tsx
@@ -78,7 +78,7 @@ export const AddNodePanel = memo(function AddNodePanel({
 
   return (
     <Toolbox
-      title={title ?? 'Add Node'}
+      title={title ?? 'Add node'}
       initialItems={nodeListOptions}
       loading={loading}
       onItemSelect={handleNodeListItemSelect}
--- a/packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
+++ b/packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
@@ -3,8 +3,8 @@ import { Column } from '@uipath/apollo-react/canvas/layouts';
 import { NodeIcon, partition } from '@uipath/apollo-react/canvas/utils';
 import { ApSkeleton, ApTypography } from '@uipath/apollo-react/material';
 import { ApIcon, ApTooltip } from '@uipath/apollo-react/material/components';
-import { memo, useCallback, useMemo } from 'react';
-import type { RowComponentProps } from 'react-window';
+import { forwardRef, memo, useCallback, useImperativeHandle, useMemo } from 'react';
+import type { ListImperativeAPI, RowComponentProps } from 'react-window';
 import { useCanvasTheme } from '../BaseCanvas/CanvasThemeContext';
 import { IconContainer, ListItemButton, SectionHeader, StyledList } from './ListView.styles';
 
@@ -35,12 +35,55 @@ export type ListItem<T = any> = {
   children?: ListItem<T>[] | ((id: string, name: string) => Promise<ListItem<T>[]>);
 };
 
-type RenderItem<T extends ListItem> =
+export type RenderItem<T extends ListItem> =
   | { type: 'section'; sectionName: string }
   | { type: 'item'; item: T };
 
+export function buildRenderedItems<T extends ListItem>(
+  items: T[],
+  enableSections: boolean
+): RenderItem<T>[] {
+  const result: RenderItem<T>[] = [];
+
+  if (items.length === 0) return result;
+
+  if (!enableSections) {
+    // if sections are disabled, return all items as is
+    for (const item of items) {
+      result.push({ type: 'item', item });
+    }
+    return result;
+  }
+
+  const [itemsWithSection, itemsWithoutSection] = partition(items, (item) => !!item.section);
+
+  // process items without sections first
+  for (const item of itemsWithoutSection) {
+    result.push({ type: 'item', item });
+  }
+
+  // return early if no items with sections
+  if (itemsWithSection.length === 0) {
+    return result;
+  }
+
+  // process items with sections next
+  const sections = Array.from(new Set(itemsWithSection.map((item) => item.section)));
+
+  for (const section of sections) {
+    result.push({ type: 'section', sectionName: section as string });
+
+    for (const item of itemsWithSection.filter((item) => item.section === section)) {
+      result.push({ type: 'item', item });
+    }
+  }
+
+  return result;
+}
+
 export interface ListViewRowProps<T extends ListItem> {
   renderedItems: RenderItem<T>[];
+  activeIndex: number;
   isLoading?: boolean;
   isDarkMode?: boolean;
   onItemClick: (item: T) => void;
@@ -55,6 +98,7 @@ const ListViewRow = memo(
     style,
     ariaAttributes,
     renderedItems,
+    activeIndex,
     isLoading,
     isDarkMode,
     onItemClick,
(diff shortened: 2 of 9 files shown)


packages/apollo-react/src/canvas/components/AddNodePanel/AddNodePanel.tsx
```diff
@@ -77,14 +78,16 @@ export const AddNodePanel = memo(function AddNodePanel({
...
-      title={title ?? 'Add Node'}
-      initialItems={nodeListOptions}
-      loading={loading}
-      onItemSelect={handleNodeListItemSelect}
-      onSearch={handleSearch}
-      onClose={onClose}
-      onItemHover={onNodeHover}
-    />
+    <FocusLock>
+      <Toolbox
```
david-rios-uipath: no change other than wrapping `Toolbox` component in `FocusLock`

packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
```diff
@@ -35,12 +35,55 @@ export type ListItem<T = any> = {
   children?: ListItem<T>[] | ((id: string, name: string) => Promise<ListItem<T>[]>);
 };
 
-type RenderItem<T extends ListItem> =
+export type RenderItem<T extends ListItem> =
   | { type: 'section'; sectionName: string }
   | { type: 'item'; item: T };
 
+export function buildRenderedItems<T extends ListItem>(
```
david-rios-uipath, line 42: Same functionality as before, just refactored to a helper outside the scope of `ListView`

packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
```diff
@@ -97,13 +141,19 @@ const ListViewRow = memo(
     const item = renderItem.item;
     const bgColor = isDarkMode ? (item.colorDark ?? item.color) : item.color;
 
+    const isActive = index === activeIndex;
+
     return (
       <ListItemButton
         {...ariaAttributes}
+        tabIndex={-1}
```
david-rios-uipath, line 149: The content in the popup (list of items in this case) should not be focusable/tabbable as per WCAG guidelines (we should maintain focus on the input): https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#:~:text=Note%3A%20The%20popup,sequence.

This has the dual purpose of keeping our focus on the input while allowing us to have tab and arrow keys navigate the same way in our list.

packages/apollo-react/src/canvas/components/AddNodePanel/AddNodePanel.tsx
```diff
@@ -1,4 +1,5 @@
 import { memo, useCallback, useMemo } from 'react';
+import FocusLock from 'react-focus-lock';
```
snuziale: was this package already installed?

packages/apollo-react/src/canvas/components/AddNodePanel/AddNodePanel.tsx
```diff
@@ -77,14 +78,16 @@ export const AddNodePanel = memo(function AddNodePanel({
...
-      initialItems={nodeListOptions}
-      loading={loading}
-      onItemSelect={handleNodeListItemSelect}
-      onSearch={handleSearch}
-      onClose={onClose}
-      onItemHover={onNodeHover}
-    />
+    <FocusLock>
+      <Toolbox
+        title={title ?? 'Add Node'}
```
snuziale: Add node - sentence casing please

packages/apollo-react/src/canvas/components/AddNodePanel/AddNodePanel.tsx
david-rios-uipath: yes -- it's used in `chat-history.tsx` and `chat-settings.tsx`

BenGSchulz (approved):
Really great improvement, will make toolbox a lot better to use thank you!

One interaction I might still be missing would be remembering the active index per navigation level. So when I nav back in the stack, the item that was stepped into is now active again.

Bonus if we can set the scroll back to that element too and it works via mouse clicking too.

Also, I might expect for the left arrow to nav back if I am still focused in the search but nothing entered, but that is maybe up for debate / WCAG best practices that I am not sure of.

We could spin off a separate ticket / PR for that too since this looks ready to go.

david-rios-uipath:
> One interaction I might still be missing would be remembering the active index per navigation level. So when I nav back in the stack, the item that was stepped into is now active again.
>
> Bonus if we can set the scroll back to that element too and it works via mouse clicking too.

This is a great idea.

> Also, I might expect for the left arrow to nav back if I am still focused in the search but nothing entered, but that is maybe up for debate / WCAG best practices that I am not sure of.

Agreed with you here -- I don't think this violates any WCAG best practices in this case input since our focus not moving from the input, and there's no in-input content to navigate. Will check before implementing it.
**_edit: [it does not](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#keyboardinteraction:~:text=node%2C%20does%20nothing.-,Left%20arrow,%3A,-When%20focus%20is). it was a quick change, so I added it to this PR._**

Will implement these features in a fast-follow PR. Thanks for the feedback!

packages/apollo-react/src/canvas/components/AddNodePanel/AddNodePanel.tsx
david-rios-uipath: I believe it's unnecessary now that we're handling tab events within `Toolbox` -- I'll test & remove it if so.
_**edit: removed `FocusLock` usage from `AddNodePanel` -- tab behavior stayed the same**_

PR #356, opened 2026-03-17
feat(apollo-react): add line break support to sticky notes [MST-7431]

david-rios-uipath:
This PR adds support for preserving newlines for display in our markdown renderer within `StickyNoteNode`.

In markdown, newlines are treated as content delimiters (they mark separation between content blocks). As a result, the markdown parser collapses consecutive blank lines into a single paragraph break.

This method is a bit of a hack -- so much so that it's documented [here](https://www.markdownguide.org/hacks/). To preserve extra blank lines, we replace runs of 3+ `\n` with `\n\n`-separated `&nbsp;` characters so that the markdown renderer treats the `&nbsp;` as a standalone HTML element that is separated from other blocks by newlines.

**Additional changes**
- Sticky note scrollbar styled
    - scrollbar track made transparent

### Before
https://github.com/user-attachments/assets/d17dc503-82c1-42ab-b63b-65eed5fb39da

### After
https://github.com/user-attachments/assets/8a733b98-9dad-441d-91d2-663af12fb1b1

### Tests
- [x] Headings, lists, code blocks, and tables render as they did before w/o extra newlines
- [x] Newlines are preserved correctly within tables
- [x] Newlines are preserved correctly within code blocks

## Note
If we highly value features like newline preservation and text alignment in our sticky notes, it might be worth considering using a `WYSIWYG` editor instead.

# Available options/rationale
### Why preprocessing?
I went with string preprocessing over attempting writing a `remark` plugin because I don't want to formalize a workaround for an established spec. This solution works well for our needs at the moment.

**Option A (preprocessing)**
The method that's used in this PR.

- Take markdown typed in by user
- Replace runs of 3+ `\n` with `\n\n`-separated `&nbsp;` characters

| Pros | Cons |
| ------------- | ------------- |
| Easy to understand and debug | Uses `&nbsp;` HTML entity nodes instead of more understandable semantic break nodes  |              
| Does not depend on `micromark`'s API | |          

**Option B (`remark` plugin)**
This solution involves 3 steps at different layers of the parse pipeline:
1. `micromark` syntax extension — tokenize consecutive `\n` as a real token                                                             
2. `mdast` extension — map token to `{ type: 'extraBreaks', count: N }`                                                  
3. `hast` handler — convert node to `<br>` elements             
                                                                   
| Pros | Cons |
| ------------- | ------------- |
| Doesn't involve mutating input directly |  Harder to debug when something goes wrong  |              
| Safe for round-tripping (markdown → AST → markdown) | `micromark`'s API is sparsely documented and can change between versions |
| Publishable as a standalone package | |

### DIFF
--- a/packages/apollo-react/src/canvas/components/StickyNoteNode/StickyNoteNode.styles.ts
+++ b/packages/apollo-react/src/canvas/components/StickyNoteNode/StickyNoteNode.styles.ts
@@ -5,6 +5,31 @@ import { motion } from 'motion/react';
 // Sticky notes must be below edges (z-index: 0). We use -10 instead of -1 to allow for future layers between sticky notes and edges if needed.
 export const STICKY_NOTE_BELOW_EDGE_Z_INDEX = -10;
 
+const stickyNoteContentStyles = `
+  width: 100%;
+  height: 100%;
+
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, sans-serif;
+  font-size: 14px;
+  line-height: 1.5;
+  color: var(--uix-canvas-foreground);
+
+  overflow-y: auto;
+
+  &::-webkit-scrollbar {
+    width: 6px;
+  }
+
+  &::-webkit-scrollbar-track {
+    background: transparent;
+  }
+
+  &::-webkit-scrollbar-thumb {
+    background: var(--uix-canvas-border);
+    border-radius: 4px;
+  }
+`;
+
 // Resize controls need to be above the sticky note content to be interactable
 export const RESIZE_CONTROL_Z_INDEX = 100;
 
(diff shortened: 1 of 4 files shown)


packages/apollo-react/src/canvas/components/StickyNoteNode/StickyNoteNode.styles.ts
```diff
@@ -5,6 +5,21 @@ import { motion } from 'motion/react';
 // Sticky notes must be below edges (z-index: 0). We use -10 instead of -1 to allow for future layers between sticky notes and edges if needed.
 export const STICKY_NOTE_BELOW_EDGE_Z_INDEX = -10;
 
+const scrollbarStyles = `
+  &::-webkit-scrollbar {
```
snuziale, line 19: why are these needed?

david-rios-uipath: w/o styling the scrollbar the track renders (can see it in the 'before' video in the PR desc) -- I'd prefer if the scrollbar aligned in appearance with our other components. Unrelated to how we handle markdown, purely a matter of preference

Image

david-rios-uipath: tangentially, it'd be better to have the container (`StickyNoteContainer`) handle the overflow behavior so we don't have repeated styling between `StickyNoteMarkdown` and `StickyNoteTextArea`. I'll try making that improvement if we want to go ahead w/ this change

snuziale (approved):

PR #357, opened 2026-03-17
refactor(apollo-vertex): remove Nextjs coupling from shell components [AGVSOL-1434]

KokoMilev:
Shell components are synced to vertical solution repos that use TanStack Router, not Next.js. This removes the framework coupling so they work correctly in consuming repos.


ruudandriessen (approved):
Shouldn't we now also include tanstack router as a dep for the shell?

alincadariu:
> Apollo Vertex Registry Check

@ruudandriessen the registry check was failing because of this, we need to make it required :trollface:

PR #366, opened 2026-03-18
feat(apollo-vertex): add TanStack Router to shell template [AGVSOL-1773]

KokoMilev:
Added `RouterProvider` from TanStack Router to the shell template so `Link` and `useLocation` in nav items stay synced.
`next/dynamic` is used with `ssr: false` to avoid hydration mismatches and `window is not defined` errors during SSR.

Added fullscreen preview toggle instead of using Nextjs routes in the standalone preview.

https://github.com/user-attachments/assets/431af069-000a-4d6b-96f2-b2e4da10a050


apps/apollo-vertex/templates/shell/ShellContextTemplate.tsx
```diff
@@ -38,25 +46,40 @@ export function ShellTemplate({
...
+  const history = createMemoryHistory({
+    initialEntries: [
+      typeof window === "undefined" ? "/" : window.location.pathname,
+    ],
+  });
+  const tanstackRouter = createRouter({ routeTree: rootRoute, history });
+
+  tanstackRouter.subscribe("onBeforeNavigate", ({ toLocation }) => {
+    nextRouter.push(toLocation.pathname);
+  });
```
alincadariu: why do we need this?

apps/apollo-vertex/templates/shell/ShellTemplateDynamic.tsx
```diff
@@ -0,0 +1,16 @@
...
+  typeof import("./ShellTemplate").ShellTemplate
+>;
+
+export const ShellTemplate = dynamic<ShellTemplateProps>(
+  () =>
+    import("./ShellTemplate").then((mod) => ({
+      default: mod.ShellTemplate,
+    })),
+  { ssr: false },
+);
```
alincadariu, line 16: i don't think we need this

alincadariu (changes requested):
can we also nuke the preview routes? i think we should only have the shell in the vertex components. without the preview. as a follow up we can add a "fullscreen toggle" such that we can see in as in preview.

Image

apps/apollo-vertex/templates/shell/ShellContextTemplate.tsx
KokoMilev: Navigation is intercepted via [onBeforeNavigate](https://tanstack.com/router/latest/docs/api/router/RouterEventsType#type-property) and delegated to Next.js's router, so the URL updates and pages render correctly.

apps/apollo-vertex/templates/shell/ShellTemplateDynamic.tsx
KokoMilev: When this is not used Tanstack begins to fight with Next js and some hydration mismatches start to appear.

KokoMilev:
> can we also nuke the preview routes? i think we should only have the shell in the vertex components. without the preview. as a follow up we can add a "fullscreen toggle" such that we can see in as in preview.
> 
> Image

So you mean to remove the `Open standalone preview` and just add a full screen option? In that case we can remove the ShellContextTemplate, since no routes have to updated in the browser's url and createMemoryHistory will do the work.

alincadariu:
> > can we also nuke the preview routes? i think we should only have the shell in the vertex components. without the preview. as a follow up we can add a "fullscreen toggle" such that we can see in as in preview.
> > Image
> 
> So you mean to remove the `Open standalone preview` and just add a full screen option? In that case we can remove the ShellContextTemplate, since no routes have to updated in the browser's url and createMemoryHistory will do the work.

@KokoMilev yea, basically to have no preview as an entry in the docs. but only the vertex components shell. let me know if its not clear. we can get in a call

alincadariu (commented):
lets remove all the code from the preview entry too as there is no need for that one now.
Image

alincadariu (approved):

PR #370, opened 2026-03-18
fix(apollo-react): improve Toolbox keyboard nav with Space, Tab, and index restore

david-rios-uipath:
Fast-follow improvements for https://github.com/UiPath/apollo-ui/pull/354

## Summary

Improves keyboard navigation in the Toolbox component by:
- adding Space key support for item selection
- fixing Tab key cycling behavior (including search clear button handling)
- restoring the active item index when navigating back from a category (for both mouse and keyboard interactions)

https://github.com/user-attachments/assets/54f1b93b-7b5f-460d-84bc-2eda1ce3a3f6

## Changes

- **Space key selection**: Space now selects the active item, matching Enter/ArrowRight behavior
- **Tab cycling**: Tab/Shift+Tab cycles through the list; when search has text, Tab from search bar defers to SearchBox (clear button focus)
- **SearchBox clear button**: Tab on the clear button returns focus to the input or navigates to the first list item
- **Back-navigation index restore**: Saves `activeIndex` in the navigation stack on category entry; restores it on back navigation. If search was active when entering, resets to search bar instead (filtered index would be invalid)
- **Refactor**: Extracted `navigateToFirstItem` as a reusable callback, replacing inline duplicates

## Flow

```mermaid
flowchart TD
    A[User enters category] -->|push| B[Save items + parentItem + activeIndex to stack]
    C[User navigates back] -->|pop| D[Restore items + parentItem]
    D --> E{Was searching on entry?}
    E -->|Yes| F[Reset to search bar]
    E -->|No| G[Restore saved activeIndex]
```

## Testing

- [x] `pnpm run test` passes (64 suites, 1257 tests)
- [x] Manual testing performed
    - [x] Tabbing to/from search input with text in input -- focused on clear button in between list items and search
    - [x] Tabbing to/from search input without text in input -- did not focus on clear button
    - [x] List item index was preserved when navigating back to parent 
    - [x] Ensure same parent/child navigation results when using arrow keys/mouse

### DIFF
--- a/packages/apollo-react/src/canvas/components/Toolbox/ListView.test.tsx
+++ b/packages/apollo-react/src/canvas/components/Toolbox/ListView.test.tsx
@@ -236,7 +236,8 @@ describe('ListView', () => {
         expect.objectContaining({
           id: 'item-1',
           name: 'Item 1',
-        })
+        }),
+        0
       );
     });
 
--- a/packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
+++ b/packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
@@ -86,7 +86,7 @@ export interface ListViewRowProps<T extends ListItem> {
   activeIndex: number;
   isLoading?: boolean;
   isDarkMode?: boolean;
-  onItemClick: (item: T) => void;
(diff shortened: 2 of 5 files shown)


BenGSchulz (approved):
🙌

PR #380, opened 2026-03-20
fix(apollo-wind): skip rendering null children on metadata form

soomin-uipath:
Addresses an issue where even if a field of the metadata form returned null to be hidden, gaps would still be formed around it
image
image


packages/apollo-wind/src/components/forms/field-renderer.tsx
```diff
@@ -261,7 +261,7 @@ export function FormFieldRenderer({
   if (isCustomField(field) && customComponents[field.component]) {
     const CustomComponent = customComponents[field.component];
     return (
-      <div style={gridStyle}>
+      <div className="empty:hidden" style={gridStyle}>
```
rahul-sharma-uipath, line 264: shouldn't there be a condition for hidden

soomin-uipath:
already supported by custom rules

PR #387, opened 2026-03-23
feat(apollo-vertex): add column resizing in data table [AGVSOL-1789]

alincadariu:


apps/apollo-vertex/registry/data-table/data-table.tsx
```diff
@@ -182,6 +188,24 @@ function DataTable<TData, TValue>({
     },
   });
 
+  /**
+   * Performant column resizing: compute all column sizes once at the table
+   * level as CSS variables, instead of calling `column.getSize()` per cell.
+   * @see https://tanstack.com/table/latest/docs/guide/column-sizing#nested-table-header-groups-with-column-resizing
+   */
+  const { columnSizingInfo, columnSizing } = table.getState();
+  const columnSizeVars = React.useMemo(() => {
```
frankkluijtmans: Is the comment needed? Also; useMemo shouldn't be necessary

apps/apollo-vertex/next-env.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/types/routes.d.ts";
+import "./.next/dev/types/routes.d.ts";
```
frankkluijtmans: We should revert this 😁

frankkluijtmans (changes requested):
Image
We should probably enforce a min width and prevent overflow

apps/apollo-vertex/registry/data-table/data-table.tsx
alincadariu: @frankkluijtmans added the react compiler to be able to remove the memo

frankkluijtmans (approved):

PR #393, opened 2026-03-24
fix(apollo-vertex): fix stale data table controls with React Compiler [AGVSOL-1906]

KokoMilev:
Data table toolbar controls (column visibility, search, pagination) were not updating in real time, only after a page refresh:

https://github.com/user-attachments/assets/8001b4fa-7473-4dc9-9b38-a649703e8509

Root cause is that React Compiler + `useReactTable` returning a mutable stable reference. The compiler treats unchanged references as unchanged data and skips re-renders: https://github.com/TanStack/virtual/issues/736

By adding`useReactTableCompat` wrapper that creates a fresh object reference per render so the compiler detects state changes: https://github.com/TanStack/virtual/issues/736#issuecomment-3065658277

https://github.com/user-attachments/assets/86f790be-2875-43f9-aa8c-effeea6458ba

### DIFF
--- a/apps/apollo-vertex/registry/data-table/data-table.tsx
+++ b/apps/apollo-vertex/registry/data-table/data-table.tsx
@@ -18,7 +18,6 @@ import {
   type RowSelectionState,
   type SortingState,
   type Table as TanstackTable,
-  useReactTable,
   type VisibilityState,
 } from "@tanstack/react-table";
 import * as React from "react";
@@ -40,8 +39,8 @@ import {
   TableHeader,
   TableRow,
 } from "@/components/ui/table";
+import { useReactTableCompat } from "@/hooks/useReactTableCompat";
 import { cn } from "@/lib/utils";
-
 import { DataTablePagination } from "./data-table-pagination";
 import { DataTableSkeleton } from "./data-table-skeleton";
 import { DataTableToolbar } from "./data-table-toolbar";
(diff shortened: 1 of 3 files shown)


apps/apollo-vertex/registry/data-table/data-table.tsx
```diff
@@ -41,7 +40,7 @@ import {
   TableRow,
 } from "@/components/ui/table";
 import { cn } from "@/lib/utils";
-
+import { useReactTableCompat } from "@/registry/use-data-table/useReactTableCompat";
```
alincadariu: this needs to be imported from hooks. check how we do it with use-local-storage

alincadariu (changes requested):

apps/apollo-vertex/registry/data-table/data-table.tsx
KokoMilev: Fixed

apps/apollo-vertex/tsconfig.json
```diff
@@ -86,8 +86,8 @@
       "@/components/ui/toggle-group": ["./registry/toggle-group/toggle-group"],
       "@/components/ui/tooltip": ["./registry/tooltip/tooltip"],
       "@/hooks/use-mobile": ["./registry/use-mobile/use-mobile"],
-      "@/hooks/use-local-storage": [
-        "./registry/use-local-storage/use-local-storage"
```
alincadariu, line 90: don't we still need the use local storage one?

alincadariu (changes requested):

apps/apollo-vertex/tsconfig.json
KokoMilev: we switched to `import { useLocalStorage } from "@uidotdev/usehooks";`

alincadariu (approved):

PR #413, opened 2026-03-27
fix(apollo-vertex): add useReactTableCompat to data-table registry

angeloaltamiranom:
The hook added in https://github.com/UiPath/apollo-ui/pull/393 was missing from the data-table files array in registry.json, so external consumers installing via shadcn CLI would not receive it and have a missing dependency.

CC: @KokoMilev


KokoMilev:
Thanks!

KokoMilev (approved):

PR #418, opened 2026-03-30
feat(apollo-react): change icon for case management

uipath-utkarsh:
## Description 

Update Case management icon 

## Demo

image

image


packages/apollo-react/src/canvas/icons/CaseManagementProject.tsx
kittyyueli, line 1: can you check this in the different themes? I think some of the fill/stroke values should be currentColor

kittyyueli:
The PR title doesn't make sense lol

packages/apollo-react/src/canvas/icons/CaseManagementProject.tsx
uipath-utkarsh: updated PR description with demo from light and dark mode after installing dev version in CM 
Please review

kittyyueli (approved):

PR #428, opened 2026-03-31
fix(apollo-vertex): registry ci check [AGVSOL-2062]

KokoMilev:
## Summary

Prevents issues like #413 from happening again by automatically catching missing registry entries at build time.

- **Replace `next build` with `tsc --noEmit`** in the registry CI check — `next build` only type-checks files in the build graph, silently skipping components not imported by any page
- **Add `--overwrite` to `shadcn add`** — without it, shadcn silently skips installation on file conflicts, so `tsc --noEmit` runs against an empty app and always passes
- **Add explicit `target` paths for data-table registry files** — preserves the subdirectory structure during installation instead of flattening files into `components/ui/`
- **Re-export `DataTableColumnHeader` from `data-table.tsx`** — works around a known shadcn bug where barrel imports are rewritten to the main file

## shadcn barrel import bug

shadcn has a known bug ([shadcn-ui/ui#7309](https://github.com/shadcn-ui/ui/issues/7309)) where imports targeting a barrel `index.ts` in a multi-file component get rewritten to the main file instead. For example:

```ts
// Source code
import { DataTableColumnHeader } from "@/components/ui/data-table";

// After shadcn install — rewrites to the main file, bypassing index.ts
import { DataTableColumnHeader } from "@/components/ui/data-table/data-table";
```
Since data-table.tsx only exported DataTable and DataTableProps, the import broke. The fix is to re-export DataTableColumnHeader from data-table.tsx so both resolution paths work. This is documented in AGENTS.md for future contributors.

### DIFF
--- a/apps/apollo-vertex/registry/data-table/data-table.tsx
+++ b/apps/apollo-vertex/registry/data-table/data-table.tsx
@@ -358,3 +358,4 @@ function DataTable<TData, TValue>({
 }
 export { DataTable };
 export type { DataTableProps };
+export { DataTableColumnHeader } from "./data-table-column-header";
--- a/.github/scripts/test-registry/test-registry.ts
+++ b/.github/scripts/test-registry/test-registry.ts
@@ -27,20 +27,21 @@ function testComponent(component: string, baseAppPath: string): TestResult {
     });
 
     // Install the component from registry
-    const addOutput = execFileSync('pnpm', ['exec', 'shadcn', 'add', `@uipath/${component}`], {
+    const addOutput = execFileSync('pnpm', ['exec', 'shadcn', 'add', `@uipath/${component}`, '--overwrite'], {
       encoding: 'utf-8',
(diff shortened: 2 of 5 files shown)


apps/apollo-vertex/registry/data-table/data-table.tsx
```diff
@@ -358,3 +358,7 @@ function DataTable<TData, TValue>({
 }
 export { DataTable };
 export type { DataTableProps };
+// Re-exported here because shadcn rewrites barrel imports (`@/components/ui/data-table`)
+// to the main file (`data-table/data-table.tsx`), bypassing index.ts entirely.
+// See: https://github.com/shadcn-ui/ui/issues/7309
```
frankkluijtmans: Good that you amended AGENTS.md to deal with these cases, but I don't necessarily think we need to leave this comment here.

frankkluijtmans (approved):
One small comment that's not blocking

PR #433, opened 2026-04-01
feat(apollo-vertex): add standalone FilterDropdown component

creilly11235:
## Summary
- Add a new standalone `FilterDropdown` registry component supporting multi-select, single-select, built-in search, and optional TanStack Table column integration
- Replace the limited `DataTableFacetedFilter` with `FilterDropdown` in the data-table registry
- Add documentation page under Vertex Components with live demos and full props reference


ruudandriessen (changes requested):
The dropdown shifts when values are selected - this is because it's centered at the "middle" of the dropdown at the moment, which means that when the content changes the dropdown also shifts.

Instead, I'd propose we anchor the dropdown differently (likely the "left" side should stick to the dropdown's left as well)

frankkluijtmans (approved):

creilly11235:
@ruudandriessen updated to left align the dropdown to prevent the shifting, should be good now!

ruudandriessen (approved):
all good, just needs a rebase

PR #453, opened 2026-04-04
fix(apollo-react): refactor adhoc tasks

kittyyueli:
## Description
Refactor adhoc tasks to add menu on right click and 3 dot click
Update style for title padding
image

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/AdhocTask.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/AdhocTask.tsx
@@ -1,0 +1,131 @@
+import { Spacing } from '@uipath/apollo-core';
+import { ApCircularProgress, ApIconButton, ApTooltip } from '@uipath/apollo-react/material';
+import debounce from 'debounce';
+import { memo, useCallback, useMemo, useRef, useState } from 'react';
+import { PlayIcon } from '../../icons';
+import type { NodeMenuItem } from '../NodeContextMenu';
+import { TaskContent } from './DraggableTask';
+import { StageTask } from './StageNode.styles';
+import type { StageTaskExecution, StageTaskItem } from './StageNode.types';
+import { TaskMenu, type TaskMenuHandle } from './TaskMenu';
+
+const AdhocTaskPlayButton = memo(
+  ({ taskId, onTaskPlay }: { taskId: string; onTaskPlay: (taskId: string) => Promise<void> }) => {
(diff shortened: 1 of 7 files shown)


packages/apollo-react/src/canvas/components/StageNode/StageNode.stories.tsx
```diff
@@ -1707,6 +1707,14 @@ export const AdhocTasks: Story = {
           onTaskClick: (taskId: string) => {
             window.alert(`Task clicked: ${taskId}`);
           },
+          onTaskGroupModification: (type: string, groupIndex: number, taskIndex: number) => {
+            console.log(
+              `Task group modification: ${type}, group: ${groupIndex}, task: ${taskIndex}`
```
joelasher-ap, line 1712: Do we wanna keep these console logs? Are you gonna use them for debugging when adding to PO Frontend? 😅

packages/apollo-react/src/canvas/components/StageNode/AdhocTask.tsx
```diff
@@ -0,0 +1,131 @@
...
+        ),
+      [onTaskPlay]
+    );
+
+    const handlePlayClick = useCallback(
+      (e: React.MouseEvent) => {
+        e.stopPropagation();
+        e.preventDefault();
+        debouncedTaskPlay(taskId);
+      },
```
joelasher-ap, line 40: Do we wanna do nothing if `playLoading` is true? Or let them spam the button to run multiple times?

joelasher-ap: Not sure if this was already discussed maybe we know users will be running adhoc tasks multiple times

packages/apollo-react/src/canvas/components/StageNode/StageNode.stories.tsx
kittyyueli: This is a story so only for the story book~

packages/apollo-react/src/canvas/components/StageNode/AdhocTask.tsx
kittyyueli: I thought about this, and I think is the best way to do it~ the loading is only showing for when the onTaskPlay callback is running. There is also a 500ms debounce in between clicks. Mostly so click play will be more intentional instead of spamming

joelasher-ap (approved):

PR #465, opened 2026-04-07
feat(apollo-vertex): add page-header component to vertex registry

creilly11235:
## Summary
- Ports the `page-header` compound component from vertical-medical-mrs to the apollo-vertex shadcn registry
- Adds docs page with live demos using a `ScaledPreview` wrapper to simulate full-width layout in narrow docs containers
- Adds shell preview pages (InvoiceDashboard header + InvoiceDetail page) using the component
- Includes improvements over MRS source: two-pass collapsible actions algorithm, named prop types for all sub-components, ref forwarding, lint compliance

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/apps/apollo-vertex/registry/page-header/page-header.tsx
+++ b/apps/apollo-vertex/registry/page-header/page-header.tsx
@@ -1,0 +1,490 @@
+"use client";
+
+/* eslint-disable max-lines -- compound component with collapsible actions */
+import { cva, type VariantProps } from "class-variance-authority";
+import { ArrowLeft, MoreHorizontal } from "lucide-react";
+import * as React from "react";
+
+import { Button } from "@/components/ui/button";
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuSeparator,
+  DropdownMenuTrigger,
+} from "@/components/ui/dropdown-menu";
+import { cn } from "@/lib/utils";
+
+const pageHeaderVariants = cva("", {
+  variants: {
+    size: {
+      default: [
+        "flex flex-wrap items-center gap-4 py-4 px-6 min-h-[92px]",
+        "@3xl:grid @3xl:grid-cols-[1fr_auto] @3xl:py-0",
+        "@3xl:has-[[data-slot=page-header-content]]:grid-cols-[3fr_6fr_3fr]",
+      ].join(" "),
+      content: "flex flex-col gap-3 pt-[10px] pb-4",
+    },
+  },
+  defaultVariants: {
+    size: "default",
+  },
+});
+
+interface PageHeaderProps
+  extends React.ComponentProps<"div">,
+    VariantProps<typeof pageHeaderVariants> {
+  bordered?: boolean;
+}
+
+function PageHeader({
+  className,
+  size,
+  bordered = false,
+  ...props
+}: PageHeaderProps) {
+  return (
+    <div className="@container shrink-0">
+      <div
+        data-slot="page-header"
+        data-size={size}
+        className={cn(
+          pageHeaderVariants({ size }),
+          bordered && "border-b border-border",
+          className,
+        )}
+        {...props}
+      />
+    </div>
+  );
+}
+
+type PageHeaderNavProps = React.ComponentProps<"div">;
+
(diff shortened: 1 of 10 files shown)


apps/apollo-vertex/registry/page-header/page-header.tsx
```diff
@@ -0,0 +1,540 @@
...
+      content: "flex flex-col gap-3 pt-[10px] pb-4",
+    },
+  },
+  defaultVariants: {
+    size: "default",
+  },
+});
+
+interface PageHeaderProps
+  extends React.ComponentProps<"div">,
```
pieman1313, line 34: import from react? same for other occurrences

apps/apollo-vertex/registry/page-header/page-header.tsx
```diff
@@ -0,0 +1,540 @@
...
+}
+
+function PageHeaderCollapsibleActions({
+  items,
+  overflowContent,
+  navMinWidth = 200,
+}: PageHeaderCollapsibleActionsProps) {
+  const containerRef = React.useRef<HTMLDivElement>(null);
+  const rulerRef = React.useRef<HTMLDivElement>(null);
+  const [visibleCount, setVisibleCount] = React.useState(-1);
```
pieman1313: why -1? wouldn't switching to null make more sense?

apps/apollo-vertex/registry/page-header/page-header.tsx
```diff
@@ -0,0 +1,540 @@
+"use client";
+
+/* eslint-disable max-lines -- compound component with collapsible actions + skeletons */
+import { cva, type VariantProps } from "class-variance-authority";
```
pieman1313, line 4: doesn't this need to be in the dependencies?

apps/apollo-vertex/registry/page-header/page-header.tsx
```diff
@@ -0,0 +1,540 @@
+"use client";
+
+/* eslint-disable max-lines -- compound component with collapsible actions + skeletons */
```
pieman1313: im a bit offput by the size as well, not sure what others think, but I would prefer each component in each file

pieman1313 (changes requested):
is this intended behaviour, or am I not using it correctly?
I made the viewport unreasonably small, and the header shrinks to the point its unreadable / unusable

Image

apps/apollo-vertex/registry/page-header/page-header.tsx
creilly11235: fixed!

apps/apollo-vertex/registry/page-header/page-header.tsx
creilly11235: fixed!

creilly11235:
> is this intended behaviour, or am I not using it correctly? I made the viewport unreasonably small, and the header shrinks to the point its unreadable / unusable
> 
> Image

On this doc page, I ended up using this work around because the preview container is artificially constraining the component and was causing some weird behavior (component was calculating page width not container width) so I had zoom to fit. If you have ideas for a better work around I'd be happy to try it!

apps/apollo-vertex/registry/page-header/page-header.tsx
pieman1313: I don't think you pushed the changes yet

pieman1313 (changes requested):
the old changes still show up, did you push the latest ones?

pieman1313:
> On this doc page, I ended up using this work around because the preview container is artificially constraining the component and was causing some weird behavior (component was calculating page width not container width) so I had zoom to fit. If you have ideas for a better work around I'd be happy to try it!

shouldn't we then switch to container queries instead of page widths? it would make this more reusable, since this will break as well if used in conjunction with the shell, right?

apps/apollo-vertex/registry/page-header/page-header.tsx
creilly11235: my bad, should be pushed now

creilly11235:
@pieman1313 should be updated and ready for review

pieman1313 (approved):

PR #472, opened 2026-04-08
feat(apollo-vertex): add feature flag provider to registry

frankkluijtmans:
Note: Check out the demo page for more details and a description of how this works. I did test in FINS coded app to verify this would work with Proteus. I will add a new `fins-solution` space in a PR to the gitops repo.

Screenshot 2026-04-08 at 14 53 59

## Summary

- Adds 4 new registry items: `feature-flag-provider` (lib), `use-feature-flag` (hook), `feature-flag-guard` (component), and `protected-feature-route` (component)
- Provider is generic with a Proteus SDK adapter, supporting reactive flag changes and Sentry integration
- Includes an interactive demo page under Vertex Components with a simulated in-memory provider to showcase all pieces

## Test plan

- [ ] Run `pnpm registry:build` — verify all 4 new items build successfully (67 total)
- [ ] Run `pnpm lint` and `pnpm format:check` — no errors
- [ ] Visit `/vertex-components/feature-flags` — toggle flags and verify reactive updates across hook, guard, and route guard demos


apps/apollo-vertex/app/vertex-components/feature-flags/page.mdx
```diff
@@ -0,0 +1,119 @@
...
+import { FeatureFlagKind } from "@uipath/proteus-client";
+import type { FlagMapping } from "@/lib/feature-flag-provider/types";
+
+export enum Flag {
+  DarkMode = "DarkMode",
+  NewOnboarding = "NewOnboarding",
+}
+
+export const flagConfig = {
+  DarkMode: {
```
ruudandriessen: I don't trust people reading docs, lets not put dark mode here, since feature flags are a really bad thing to use feature flags for and I'm afraid people are gonna do that now 😛

apps/apollo-vertex/registry/use-feature-flag/use-feature-flag.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from "react";
+import { getFeatureFlagProvider } from "@/lib/feature-flag-provider";
+
+export function useFeatureFlag(key: string): unknown {
```
ruudandriessen: would be cool if this is all type safe

ruudandriessen: we currently kinda assume these are booleans right? So maybe lets reflect that in the API?

apps/apollo-vertex/registry/feature-flag-provider/feature-flag-provider.ts
```diff
@@ -0,0 +1,28 @@
+import type { FeatureFlagProviderConfig } from "@/lib/feature-flag-provider/types";
+
+let _provider: FeatureFlagProviderConfig<
```
ruudandriessen: i don't really like this pattern where it's a global, can't we make this a real "provider" like the name kind of implies?

ruudandriessen (changes requested):

apps/apollo-vertex/app/vertex-components/feature-flags/page.mdx
frankkluijtmans: I get your point, I'll make sure we have a demo that makes sense for the context of feature flags.

frankkluijtmans: Replaced DarkMode with NewDashboard in the demo and docs.

apps/apollo-vertex/registry/use-feature-flag/use-feature-flag.ts
frankkluijtmans: Done — `useFeatureFlag` now returns `boolean`, and the full provider config (`FeatureFlagProviderConfig`) is typed around boolean values. The `on`/`off` callbacks are `(value: boolean) => void`.

apps/apollo-vertex/registry/feature-flag-provider/feature-flag-provider.ts
frankkluijtmans: Refactored to a proper React context provider. The singleton is gone — consumers now wrap their app with `<FeatureFlagProvider provider={...}>` and the hook reads from context internally.

ruudandriessen (approved):

PR #493, opened 2026-04-13
feat(apollo-vertex): components grid overview page

ruudandriessen:
## Summary
- Adds a new components overview page (`/components`) that displays all 59 components as a responsive grid of clickable cards with names and descriptions
- Hides individual component entries from the sidebar navigation — components are now accessed by clicking through the grid overview
- Keeps all existing component detail pages intact at their current URLs

image


angeloaltamiranom (approved):

hfrancis31:
@ruudandriessen - The addition of the grid view of all the components is cool. But can we bring back the list of components? We use it to view the visuals of the components all the time. Also, the fact that the navigation disappears and there is no real indication of how to bring it back is an odd behavior. Can we make sure that navigation doesn't disappear? 

If you want me to make the updates and push a PR I can. Just wanted to ping ya.

PR #512, opened 2026-04-14
fix(apollo-react): update tool call for different display mode

shannonsuhendra:
- Show 3 different display modes
Screenshot 2026-04-14 at 4 26 26 PM
Screenshot 2026-04-14 at 4 26 36 PM


packages/apollo-react/src/material/components/ap-tool-call/assets/spanIcons.ts
```diff
@@ -0,0 +1,187 @@
+/* eslint-disable max-len */
+export const spanIconsMap = new Map<string, string>([
+  [
+    'agent-run',
```
CalinaCristian, line 4: https://apollo-ui-react.vercel.app/core/icons
are those icons here or do they need to be defined ?

CalinaCristian (approved):

PR #526, opened 2026-04-16
feat(apollo-react): expose fallbackPlacement prop on FloatingCanvasPanel

venkateshcherukuri:
## Summary
- Expose floating-ui's `fallbackPlacement` option as an optional prop on `FloatingCanvasPanel` and `useFloatingPosition`.
- When the primary axis (e.g. top/bottom) overflows in both directions, this allows the popover to fall back to cross-axis placements (left/right).
- Defaults to `'none'` (no behavior change for existing consumers)

## Changes
- `useFloatingPosition.ts`: Added `fallbackPlacement` to options interface and passed it to `flip()`.
- `FloatingCanvasPanel.tsx`: Added `fallbackPlacement` prop and forwarded it to `useFloatingPosition`.

### DIFF
--- a/packages/apollo-react/src/canvas/components/FloatingCanvasPanel/FloatingCanvasPanel.tsx
+++ b/packages/apollo-react/src/canvas/components/FloatingCanvasPanel/FloatingCanvasPanel.tsx
@@ -32,6 +32,15 @@ export type FloatingCanvasPanelProps = {
   anchorRect?: AnchorRect;
   placement?: Placement;
   offset?: number;
+  /**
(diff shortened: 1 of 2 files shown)


jevinkosasih (commented):
nit: `fallbackAxisSideDirection` is a floating-ui API name, consumers might not be familiar with it.
 
It might be worth using a more friendly name + a comment explaining what it does and link to the doc https://floating-ui.com/docs/flip#fallbackaxissidedirection

Maybe `overflowPlacement` or something like that

jevinkosasih (approved):

PR #542, opened 2026-04-20
feat(apollo-vertex): ai-chat registry component — core [1/2]

petervachon:
## Summary

- Adds the full `registry/ai-chat/` component suite: message rendering, input, code blocks, markdown, thinking/loading states, flow UI, selection menu, message actions, autopilot icons, hooks (`use-sticky-scroll`, `use-thinking-label`, `use-typewriter`), shared types, and utilities
- Updates `registry.json` with all ai-chat registry entries
- Adds `lib/auth.ts` auth helper and `next-env.d.ts` type reference
- Adds `@ai-sdk/react` dependency to `package.json`

## Stacking

This is **PR 1 of 2**. PR 2 ([`pr/2-aichat-templates`]) targets this branch and adds the templates and documentation pages.

> Supersedes #511 — split into stacked PRs per reviewer request.

## Test plan

- [ ] `git checkout pr/1-aichat-core && pnpm dev` — dev server starts without errors
- [ ] No broken imports: `pnpm build` passes
- [ ] Registry components exist at `registry/ai-chat/` with all subfiles present
- [ ] `registry.json` contains ai-chat entries

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #511, opened 2026-04-14
AI Chat UX

petervachon:
## Summary

This PR delivers the next wave of AI Chat UX improvements for Apollo Vertex: a new multi-step guided flow interaction, richer input with file/image attachments, context-menu "Ask Autopilot" selection, message actions, and a full-coverage component preview page.

---

### Multi-step flow (`AiChatFlow`)
- New `AiChatFlow` component renders a guided, step-by-step interaction card docked above the input
- Card visually wraps the input — input floats in front with 8px inset gap on left/right/bottom, creating a connected "Claude co-work" style layout
- Numbered badges (secondary styling), dividers between options, animated question text
- Chevron back/forward navigation always visible, dimmed at boundaries; step counter (e.g. 2/4) inline with controls
- Free-text answers supported at every step — input placeholder changes to "Or type your own answer..." when flow is active, and submitting routes the text directly into the flow without needing a "Something else" option
- LLM prompts updated to prohibit catch-all options ("Something else", "Other", "None of the above") — the free-text input handles that case
- `flowTool` wired into `AiChatAgentHubMode` alongside `choicesTools`
- Stale resolver ref bug fixed — "Ask Autopilot" context menu no longer hangs after a flow completes

### File & image attachments
- `+` menu in the input opens a file picker (multiple files)
- Paste an image directly into the input to attach it
- File chips show name + remove button for documents; image files render a 40×40 thumbnail with a 1px border
- Click a thumbnail to open a full lightbox overlay (backdrop blur, close button top-right, Escape to dismiss)
- Object URLs are properly revoked on remove and submit to prevent memory leaks
- Note: files are UI-only for now — LLM integration deferred until the target model/endpoint is confirmed

### Context menu — "Ask Autopilot"
- Select text anywhere in the chat and a context menu appears with "Ask Autopilot"
- Quoted text is shown as a chip inside the input; clear button removes it
- Selecting text in a new location replaces the previous quote

### Message actions
- Copy, quote reply, and regenerate actions on assistant messages

### Preview page
- `/vertex-components/ai-chat/preview` now covers all isolated sub-components with live demos: loading states, thinking animation, suggestion chips, multi-step flow, message actions, error banner, empty state
- Link added to the `/patterns/ai-chat` documentation page

### Bug fixes (PR review)
- Restored `useTranslation` and `aria-label` in `PreviewFullScreen`
- Reverted `import type * as React` → `import * as React` in `accordion.tsx`
[...]

PR #544, opened 2026-04-20
feat(apollo-vertex): ai-chat foundation — types, hooks, utils, icons [1/6]

petervachon:
## Summary

- Adds shared TypeScript types (`types.ts`) and utility functions (`ai-chat-utils.ts`)
- Adds hooks: `use-sticky-scroll`, `use-typewriter`
- Adds autopilot icons (`autopilot.tsx`, `autopilot-gradient.tsx`)
- Updates `registry.json` with ai-chat registry entries
- Adds `@ai-sdk/react` to `package.json`, `lib/auth.ts`, and `next-env.d.ts`

## Stack

```
pr/1a-aichat-foundation  →  main          ← this PR, ready for review
pr/1b-aichat-display     →  1a            draft
pr/1c-aichat-input       →  1b            draft
pr/1d-aichat-messages    →  1c            draft
pr/1e-aichat-chat        →  1d            draft
pr/2-aichat-templates    →  1e            draft
```

> Supersedes #542 / #511 — re-split to keep each PR under 800–1000 LoC.

## Test plan

- [ ] `git checkout pr/1a-aichat-foundation && pnpm dev` — dev server starts without errors
- [ ] `registry/ai-chat/types.ts`, `utils/`, `hooks/`, and `icons/` all present
- [ ] `registry.json` contains ai-chat entries


apps/apollo-vertex/registry/ai-chat/components/icons/autopilot.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+interface AutopilotIconProps
+  extends Omit<React.SVGProps<SVGSVGElement>, "width" | "height"> {
+  size?: string | number;
+}
+
+export const AutopilotIcon = React.forwardRef<
+  SVGSVGElement,
+  AutopilotIconProps
+>(({ size = 24, ...props }, ref) => (
```
0xr3ngar: why use forwardRef? With React 19, you can finally treat ref like any other prop. No forwardRef

https://react.dev/reference/react/forwardRef

apps/apollo-vertex/registry/ai-chat/components/icons/autopilot-gradient.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * Uses the same gradient stops as `--ai-gradient-strong` so it matches
+ * the gradient used on the chat send button, suggestion buttons, and title.
+ *
+ * Each instance generates a unique gradient ID via `useId` to prevent
+ * collisions when multiple instances are rendered on the same page.
+ */
+export const AutopilotGradientIcon = React.forwardRef<
+  SVGSVGElement,
+  AutopilotGradientIconProps
+>(({ size = 24, ...props }, ref) => {
```
0xr3ngar: why use forwardRef? With React 19, you can finally treat ref like any other prop. No forwardRef

https://react.dev/reference/react/forwardRef

apps/apollo-vertex/registry/ai-chat/hooks/use-thinking-label.ts
```diff
@@ -0,0 +1,79 @@
...
+const THINKING_LABELS = [
+  "Thinking…",
+  "Se gândește…",
+  "Thinking…",
+  "सोच रहा है…",
+  "Thinking…",
+  "Denkt nach…",
+  "Thinking…",
+  "Réflexion…",
+];
```
0xr3ngar: I don't think this should be here, I think the user should provide these as props since these should be translated in the app ( via the useTranslate() hook )

apps/apollo-vertex/registry/ai-chat/hooks/use-thinking-label.ts
0xr3ngar, line 1: do we really need this thinking cycling ? I don't expect that the LLM will take more than 2 seconds to give a response. I feel like we are just copying claude code. 

I think this is purely cosmetic (no actual information), it can feel twee or try-hard ("Réflexion…" 🙄), it adds state and timers to manage, and on slow devices the cycling can feel jank

We have to keep in mind that not all users have macbook pros and some are running on shitty thinkpads with 8 gb ram

0xr3ngar: Could we replace this with a CSS pulse or shimmer on a single "Thinking…"? Same reassurance that the UI isn't frozen, zero state to manage, and it actually looks better on slow devices

apps/apollo-vertex/registry/ai-chat/hooks/use-typewriter.ts
0xr3ngar, line 1: Same question as the thinking-labels hook: do we actually need this?

The comment says ChatGPT/Claude throttle to 30-50 cps, but the reason they do it is to smooth out bursty token streams - not because throttling is inherently good. If our backend streams tokens at a reasonable speed, the browser handles pacing for free

My main concerns are
- Fast readers end up waiting on the animation instead of the model (a real ChatGPT complaint)
- We're re-rendering up to 60fps via rAF instead of per-token (maybe 10-30/sec). If the message contains markdown, that's re-parsing on every frame - not free on the 8GB ThinkPads
- The "drain 2× when streaming ends" logic is a smell - it exists because the throttle falls behind reality, which means it was adding latency, not matching reading speed
- `useRef` + `useReducer` to force re-renders is fighting React. If we keep any version of this, it should be `useState` for `displayedLength` - in short we need to refactor this

0xr3ngar: Also why do we opt in for an in-house hook implementation when we can just use a package that does this? plenty of people have already solved this problem

apps/apollo-vertex/registry/ai-chat/utils/ai-chat-utils.ts
```diff
@@ -1,6 +1,65 @@
...
+        const result = toolResultFlowSchema.safeParse(
+          (part as { output?: unknown }).output,
+        );
+        if (result.success) return result.data;
+      }
+    }
+  }
+  return null;
+}
+
```
0xr3ngar: findLatestFlow scans backwards with hasUserAfter check 

This is O(n²) because messages.slice(i + 1).some(...) re-scans for every iteration 

For a long conversation this really matters. 

Simpler approach could be to find the last user message index once, then only look at assistant messages after it - same structure as findActiveChoicesMessageIds already uses. These two functions are doing the same "find trailing assistant messages" work with different implementations.

apps/apollo-vertex/registry/ai-chat/utils/ai-chat-utils.ts
```diff
@@ -41,3 +104,78 @@ export function findLatestChoices(
...
+        if (
+          parsed !== null &&
+          typeof parsed === "object" &&
+          "type" in parsed &&
+          (parsed as { type: unknown }).type === "choices"
+        )
+          return true;
+      } catch {
+        // invalid JSON, skip
+      }
```
0xr3ngar: this should already be typed

0xr3ngar: The casts (part as { content: string }) and (part as { output?: unknown }) suggest the UIMessagePart type isn't discriminated properly. If type === "tool-result" doesn't narrow part to have content, that's a type definition problem upstream worth fixing

apps/apollo-vertex/registry/ai-chat/utils/ai-chat-utils.ts
```diff
@@ -41,3 +104,78 @@ export function findLatestChoices(
...
+      if (
+        output != null &&
+        typeof output === "object" &&
+        "type" in output &&
+        (output as { type: unknown }).type === "choices"
+      ) {
+        return true;
+      }
+    }
+    return false;
```
0xr3ngar: same thing as above, this is a type issue

apps/apollo-vertex/registry/ai-chat/utils/ai-chat-utils.ts
```diff
@@ -41,3 +104,78 @@ export function findLatestChoices(
...
+  const trailingAssistants = messages
+    .slice(lastUserIdx + 1)
+    .filter((m) => m.role === "assistant");
+
+  // Only suppress actions if at least one trailing assistant has choices
+  const hasActiveChoices = trailingAssistants.some((m) => messageHasChoices(m));
+  if (!hasActiveChoices) return new Set();
+
+  return new Set(trailingAssistants.map((m) => m.id));
+}
```
0xr3ngar: this suppresses actions on *all* trailing assistant messages when any of them has choices. Is that what we want if an unrelated assistant message (status update, other tool result) happens to land after a choices message? It'd lose its copy/regenerate too.

If our turn structure guarantees that can't happen, worth a comment stating the invariant. Otherwise we probably want to scope the suppression more tightly

Also: this can be one backwards pass instead of three (scan -> slice/filter -> some -> map), and the name implies it finds the choices messages when it actually finds their whole turn - `getActiveChoiceTurnMessageIds` would be clearer.

0xr3ngar (changes requested):

0xr3ngar (changes requested):
also another thing, ci checks are all failing

apps/apollo-vertex/registry/ai-chat/utils/ai-chat-utils.ts
```diff
@@ -1,6 +1,72 @@
 import type { UIMessage } from "@tanstack/ai-client";
 import { z } from "zod";
 
+// ─── Flow (client-side multi-step) ───────────────────────────────────────────
+
+const flowOptionSchema = z.object({
```
pieman1313: this entire file is gone on my pr where we refactor the choices tool to be more generic https://github.com/UiPath/apollo-ui/pull/584

I suggest we wait until mine is merged, and restart this approach because having just types being checked in with implementation on subsequent prs is hard to follow

we should strive to have end to end concerns / implementations in one pr

pieman1313 (changes requested):

petervachon:
Hi @0xr3ngar and @pieman1313 — thanks for the detailed feedback. Here's a summary of what's been addressed since your reviews:

**@0xr3ngar**

- **`use-typewriter.ts`** — removed entirely. Stream pacing from the backend handles the reveal naturally; no rAF loop, no re-parsing markdown on every frame.
- **`use-thinking-label.ts`** — removed. `AiChatLoading` now uses a static "Thinking…" with a CSS shimmer (no state, no timers). We've left a TODO comment in the component outlining a latency-aware progression (500ms→2s→4s+) for when we want to build that out properly.
- **`forwardRef` on icons** — the icons use the React 19 ref-as-prop pattern already; no `forwardRef` wrapper.
- **`findLatestFlow` O(n²)** — fixed. One backwards pass to find `lastUserIdx`, then a single scan from the tail down to that index.
- **Type casts** — removed. `tool-result` content is accessed directly; `tool-call` output is routed through `unknown` before being passed to `safeParse`.
- **`getActiveChoiceTurnMessageIds` naming** — already using the clearer name you suggested.

**@pieman1313**

PR #584 has merged, so the conflict around `ai-chat-utils.ts` is no longer an issue. Our branches were rebased onto main after the merge.

Would appreciate a re-review when you get a chance!

apps/apollo-vertex/registry/ai-chat/utils/ai-chat-utils.ts
```diff
@@ -0,0 +1,185 @@
+import type { UIMessage } from "@tanstack/ai-client";
```
pieman1313: since we moved away from this approach, I am doubting if its a good ideea to re-build on top of it. I would prefer we have a complete end to end working example to judge, rather than just some types.

pieman1313 (changes requested):
I had a look at the [initial pr](https://github.com/UiPath/apollo-ui/pull/511/changes), and considered it as the end state / north start towards which your current prs are headed.
I do not agree with the splitting of prs, and would prefer we try to have small, end to end features building one upon the other, instead of what we have now, where we took the end state pr and just broke it down to smaller prs that on their own don't give the full picture.

I poked Claude on the end state pr to break it down into complete end to end example prs, although I do doubt some of them due to the features themselves. 

Here is an example breakdown of the [big pr](https://github.com/UiPath/apollo-ui/pull/511/changes), although, I do not agree with some features, and would like to postpone them a bit until we get the bulk of the work in, and then we can focus on doing those features better.
- We should not add any mock responses or the like. we have hooked up actual llms, and should showcase everything using them
- we should postpone the flow step + choices feature until after we get the bulk of the features in

what do you think?

PR 1 — feat(apollo-vertex): ai-chat welcome screen + Autopilot rebrand

What the user gets: Open /patterns/ai-chat, see a centered "What are we
tackling today?" empty state with Autopilot branding and 3 starter
suggestion chips. Click a chip, it sends as a message.

Includes:
- New ai-chat-empty-state component
- emptyState and suggestions props on AiChat
- --ai-chat-muted-foreground token (only what the empty state paints) added
  to registry.json
- AgentHub wiring: title set to "Autopilot", emptyState passed in,
  suggestions array passed in
- Preview page: scaffolds the ai-chat preview page with an "Empty state"
  section and a nav entry


PR 2 — feat(apollo-vertex): syntax-highlighted code blocks with copy

What the user gets: Ask Autopilot for a code snippet — the response renders
in a styled block with language label, syntax highlighting (light and dark
via github-dark-dimmed), and a copy button.

Includes:
- New ai-chat-code-block component, with ref-based innerHTML safety folded
  in from the CI-fix commit
- ai-chat-markdown rewires pre/code blocks to render AiChatCodeBlock
- Preview page: adds a "Code blocks" section showing three languages and the
  dark variant


PR 3 — feat(apollo-vertex): multi-step interactive flows

What the user gets: Ask Autopilot something open-ended ("help me draft a
brief") and it presents a 2–8 step decision flow inline. User clicks
through, can go back or skip, then a single message with all answers is
sent. Suggestion chips also gain the same step counter / back / skip /
dismiss UI for the existing presentChoices tool.

Includes:
- New ai-chat-flow component and flow-tool example
- ai-chat-suggestions extended with step, totalSteps, canGoBack, canSkip and
  framer-motion entrance animations
- choices-tool prompt updated to describe multi-step usage
- findLatestFlow, tryParseFlow, and findActiveChoicesMessageIds added to
  ai-chat-utils, plus the nesting-depth refactor from the CI-fix commit
- AgentHub wiring: register flowTool, append FLOW_TOOL_PROMPT to the system
  prompt
- Preview page: adds an interactive "Flow" demo using the existing MOCK_FLOW


PR 4 — feat(apollo-vertex): animated thinking + assistant typewriter reveal

What the user gets: Send a message — see an animated sparkle morph into a
pulsing circle, with the "Thinking…" label cycling through translations.
When the response streams, text is throttled to ~75 characters/second for
readable reveal. Suggestion chips wait until the typewriter finishes before
appearing.

Includes:
- New ai-chat-thinking component, use-thinking-label hook, use-typewriter
  hook
- ai-chat-loading rewritten (shimmer plus label cycler)
- Introduces ai-chat-provider and the AiChatConfig context, with only the
  fields this PR needs: typewriterCps, latestAssistantMessageId,
  isLatestResponseAnimating, and its setter. Future PRs extend it
- AgentHub wiring: wrap the chat with AiChatProvider configured at 75 cps
- Preview page: adds the ThinkingDemo as an isolated playground


PR 5 — feat(apollo-vertex): message actions + inline user edit

What the user gets: Hover any assistant message — get copy, thumbs up,
thumbs down, and regenerate buttons. Hover a user message — get an edit
pencil. Click edit, inline textarea with save and cancel; on save the
conversation rewinds and re-sends. Actions are suppressed during an active
choice turn, using the helper introduced in PR 3.

Includes:
- New ai-chat-message-actions component
- Edit mode added to ai-chat-message
- AiChatConfig extended with showMessageActions, showCopyButton,
  activeChoicesMessageIds, onFeedback, onRegenerate, onEditMessage, plus the
  MessageFeedbackType type
- AgentHub wiring: onRegenerate calls reload, onEditMessage calls sendMessage
- Preview page: adds "Message actions" and "Edit mode" sections


PR 6 — feat(apollo-vertex): file attachments, paste image, source citations

What the user gets: Paste an image into the input — thumbnail appears as a
chip, click to expand in a lightbox. Click the paperclip — file picker for
any file type. Send — message bubble shows the attachment chips. Assistant
responses can render source citation chips below.

Includes:
- ai-chat-input rewrite: PendingFile model, file picker, paste handler,
  thumbnail and dialog-based lightbox. Folds in the uid, instanceof
  narrowing, and void onClick fixes from the CI-fix commit
- MessageSource and MessageAttachment types, with rendering in
  ai-chat-message
- AgentHub wiring: onSubmit accepts (text, files?) and forwards files
- Preview page: adds "Attachments" and "Sources" sections using
  MOCK_ATTACHMENTS_BASIC and MOCK_SOURCES_BASIC


PR 7 — feat(apollo-vertex): "Ask Autopilot" text selection menu

What the user gets: Highlight any text in an assistant message — a small
"Ask Autopilot" pill appears above the selection. Click it — the input gets
a quote chip with the highlighted text. Type a follow-up and send. Quote
chip is removable.

Includes:
- New ai-chat-selection-menu component, autopilot icon, and autopilot
  gradient icon
- Quote chip slot added to ai-chat-input
- Selection capture wired into ai-chat-message
- AiChatConfig extended with onQuoteSelect, plus a new enableTextSelection
  prop on AiChat
- AgentHub wiring: enableTextSelection turned on, onQuoteSelect populates
  the input quote chip
- Preview page: adds a "Selection menu" section with mock assistant text the
  user can highlight

apps/apollo-vertex/registry/ai-chat/utils/ai-chat-utils.ts
petervachon: Okay. I originally had that in there to use with the recommended prompts (as part of assistant response) as well as a multi-step interaction which presents in more of a card carousel of questions and answers. 

I've removed those to clean things up, we can revisit later on.

petervachon:
This PR is part of the original ai-chat stack which has been superseded by a restructured version. Closing in favour of the new stack:

- #628 — Shell, empty state, thinking indicator [1/5]
- #629 — Syntax-highlighted code blocks [2/5]
- #630 — Message actions: copy, feedback, edit, regenerate [3/5]
- #631 — File attachments: upload, paste images, preview [4/5]
- #632 — Text selection: Ask AI floating pill + quote chip [5/5]

The new stack is reorganised by user-facing feature rather than component type, includes all the same functionality plus attachments and text selection, and has a clean single-commit-per-PR history.

PR #556, opened 2026-04-21
feat(apollo-react): add loop node [MST-8596]

SreedharAvvari:
This PR adds a container-style loop node to `apollo-react` and wires the canvas preview/add-node flows needed to place real child nodes inside the loop body.

## Core changes
- adds `LoopNode` / `LoopCanvasNode` for manifests with `display.shape: 'container'`, including the loop shell, header, body frame, empty-state add CTA, resize controls, adornments, status styling, and toolbar support
- extends node manifests with the `container` shape and handle groups with `boundary: 'outer' | 'inner'`, allowing loop handles to render on the outer shell or inner body frame while keeping React Flow edge anchors aligned
- updates `ButtonHandle`, `HandleButton`, and `NodeViewportOverlay` so handle affordances can use separate visual/connection positions, inward-facing container handles, stable hit areas, and portal-rendered buttons/labels above React Flow node layers
- introduces `createPreviewGraph` on top of `createPreviewNode`, supporting one or two preview edges, replacing an existing edge while inserting a node, and reparenting preview nodes into containers with `parentId` / `extent: 'parent'`
- updates add-node flows (`createAddNodePreview`, connect-end insertion, edge-toolbar insertion, and `AddNodeManager`) to use preview graphs, preserve container parentage when materializing nodes, restore removed edges on cancellation, and treat any edge touching the preview node as a preview edge
- maps container-shaped manifests to `LoopCanvasNode` in `HierarchicalCanvas` and Storybook helpers while keeping other manifest-driven nodes on `BaseNode`
- updates container sizing constants, collapsed/expanded container sizing, React Flow z-index reset behavior, and toolbar/handle overlay layering needed for loop nodes
- updates the Storybook `uipath.control-flow.foreach` manifest to `shape: 'container'` with outer input/success handles and inner `Start`, `Continue`, and `Break` handles

## Reviewer focus
- `LoopNode.tsx` / `LoopCanvasNode.tsx`: container rendering, resize behavior, empty-state first-child preview, inner/outer handle visibility, toolbar/adornment/status behavior
- `ButtonHandle.tsx` / `HandleButton.tsx` / `NodeViewportOverlay.tsx`: separate visual vs connection anchors and portal layering for handle affordances and toolbars
- `createPreviewGraph.ts` and callers: multi-edge preview creation, edge replacement/restoration, and reparenting previews into containers
- `HierarchicalCanvas.tsx` and Storybook manifest helpers: container manifest detection and preview-edge cleanup during connect flows

## Coverage
- updates existing `useAddNodeOnConnectEnd` and `useEdgeToolbarState` unit tests for the move to `showPreviewGraph`
- adds a Storybook `Canvas/LoopNode` scenario with connected ingress, loop body, and egress nodes

Ref: https://uipath.atlassian.net/browse/MST-8596


BenGSchulz (commented):
I noticed some unexpected behavior when adding nodes within a loop.

The "add between" seems to preserve parenting (i.e. I can't move the node out of the container).

But the "add from handle" does not seem to preserve parenting. Is this expected, like we could technically break out of the loop that way? Or do we want to lock down parenting so any node created off of a node with a parent will also be parented?

Dragging a node within the container causes the full canvas to pan, which was a bit disorienting.

https://github.com/user-attachments/assets/adda161b-6b10-48e2-8e7c-eba0815e1d97

SreedharAvvari:
> I noticed some unexpected behavior when adding nodes within a loop.
> 
> The "add between" seems to preserve parenting (i.e. I can't move the node out of the container).
> 
> But the "add from handle" does not seem to preserve parenting. Is this expected, like we could technically break out of the loop that way? Or do we want to lock down parenting so any node created off of a node with a parent will also be parented?
> 
> Dragging a node within the container causes the full canvas to pan, which was a bit disorienting.
> 
>  Screen.Recording.2026-04-22.at.1.32.58.PM.mov
 
The behaviour I was aiming for is -
  - any node added inside a loop container should be parented to that container by default
  - if the user drags that node outside the container, it should detach and move freely

So the reported issues are valid -
  - add from handle should preserve parenting when the new node belongs inside the container - will fix that
  - dragging a child out should detach it rather than causing the canvas to pan

My original thought was to keep the detachment logic in the consuming layer (flow-workbench), since that’s where the full reparenting/edge-repair behaviour belongs. But I agree the current Storybook demo feels broken/confusing without it, so for now I can add a small story-only helper to make the interaction behave as intended in the demo, while keeping the real production logic in flow-workbench. wdyt?

packages/apollo-react/src/canvas/storybook-utils/manifests/node-definitions.ts
```diff
@@ -255,10 +259,72 @@ export const allNodeManifests: NodeManifest[] = [
...
+      {
+        position: 'right',
+        boundary: 'inner',
+        handles: [
+          {
+            id: 'continue',
+            label: 'Continue',
+            type: 'target',
+            handleType: 'input',
+            isDefaultForType: true,
```
BenGSchulz: As a heads up for creating the final OOTB node manifest, using `isDefaultForType` here causes the source node for the loop node to auto connect to the continue handle. I am thinking that you probably just need to set this on the outer target handle, but maybe we need to explore taking the boundary into consideration when determining default handle.

Image

BenGSchulz:
> The behaviour I was aiming for is -
> 
> * any node added inside a loop container should be parented to that container by default
> * if the user drags that node outside the container, it should detach and move freely
> 
> So the reported issues are valid -
> 
> * add from handle should preserve parenting when the new node belongs inside the container - will fix that
> * dragging a child out should detach it rather than causing the canvas to pan
> 
> My original thought was to keep the detachment logic in the consuming layer (flow-workbench), since that’s where the full reparenting/edge-repair behaviour belongs. But I agree the current Storybook demo feels broken/confusing without it, so for now I can add a small story-only helper to make the interaction behave as intended in the demo, while keeping the real production logic in flow-workbench. wdyt?

👍 Yes that makes sense, thank you.

packages/apollo-react/src/canvas/components/AddNodePanel/AddNodeManager.tsx
```diff
@@ -207,7 +228,9 @@ export const AddNodeManager: React.FC<AddNodeManagerProps> = ({
       const [firstConnection] = previewNodeConnectionInfo;
       if (firstConnection) {
         const firstEdgeSourceHandle = firstConnection.addNewNodeAsSource
-          ? newNodeManifest && registry?.getDefaultHandle(newNodeManifest.nodeType, 'source')?.id
+          ? newNodeManifest &&
+            registry?.getDefaultHandle(newNodeManifest.nodeType, 'source', nodeItem.data.version)
+              ?.id
```
CalinaCristian: nit: this looks harder to read

packages/apollo-react/src/canvas/components/AddNodePanel/createAddNodePreview.ts
```diff
@@ -1,38 +0,0 @@
-import { Position, type ReactFlowInstance } from '@uipath/apollo-react/canvas/xyflow/react';
```
CalinaCristian, line 1: Image

we are using this in flow-workbench, this seems like a breaking change.

packages/apollo-react/src/canvas/components/ButtonHandle/SmartHandle.tsx
```diff
@@ -721,7 +721,7 @@ export function SmartHandle({
       {/* Only render visual elements when visible */}
       {visible && (
         <>
-          {onAction && type === 'source' ? (
+          {onAction && type === 'source' && showButton ? (
```
CalinaCristian: intended?

packages/apollo-react/src/canvas/components/HierarchicalCanvas/HierarchicalCanvas.tsx
```diff
@@ -105,17 +108,19 @@ export const HierarchicalCanvas: React.FC<HierarchicalCanvasProps> = ({
...
+        .map(({ nodeType }) => nodeType)
+        .filter((nodeType) => !Object.hasOwn(DEFAULT_NODE_TYPES, nodeType))
+        .map((nodeType) => [nodeType, BaseNode])
+    ) as NodeTypes;
+
+    return {
+      ...DEFAULT_NODE_TYPES,
+      ...manifestNodeTypes,
+      ...additionalNodeTypes,
+    } as NodeTypes;
```
CalinaCristian: do we need the castings?

CalinaCristian (commented):
https://github.com/user-attachments/assets/6f7154af-4b7a-4ab4-957e-3f044c697461

some issues:
1. container doesn't grow when adding more nodes to it.
2. If I go on a different path, it doesn't auto-connect to continue/break/etc, so what happens ?
image

In the [prototype](https://flow-prototype-is.vercel.app/) it looks like it automatically connects it.

packages/apollo-react/src/canvas/components/AddNodePanel/createAddNodePreview.ts
SreedharAvvari: reverted

packages/apollo-react/src/canvas/components/ButtonHandle/SmartHandle.tsx
SreedharAvvari: removed

packages/apollo-react/src/canvas/components/HierarchicalCanvas/HierarchicalCanvas.tsx
SreedharAvvari: removed

packages/apollo-react/src/canvas/storybook-utils/manifests/node-definitions.ts
SreedharAvvari: fixed this

packages/apollo-react/src/canvas/utils/createPreviewGraph.ts
```diff
@@ -0,0 +1,199 @@
...
+  );
+
+  return {
+    ...previewNode,
+    position: {
+      x: previewNode.position.x - containerAbsolutePosition.x,
+      y: previewNode.position.y - containerAbsolutePosition.y,
+    },
+    parentId: containerId,
+    extent: 'parent',
```
SreedharAvvari: check if this is needed

packages/apollo-react/src/canvas/components/AddNodePanel/AddNodeManager.tsx
```diff
@@ -141,6 +141,8 @@ export const AddNodeManager: React.FC<AddNodeManagerProps> = ({
         position: currentPreviewNode.position,
         selected: true,
         data: nodeData,
+        parentId: currentPreviewNode.parentId,
+        extent: currentPreviewNode.extent,
```
CalinaCristian: keeping?

packages/apollo-react/src/canvas/components/ButtonHandle/ButtonHandle.tsx
```diff
@@ -183,11 +281,187 @@ const ButtonHandleBase = ({
 
 export const ButtonHandle = memo(ButtonHandleBase);
 
+function getHandleActionPortal({
```
CalinaCristian: can we extract this one and the `getInwardHandleLayout` into it's own utils?

packages/apollo-react/src/canvas/components/LoopNode/LoopNode.helpers.ts
```diff
@@ -0,0 +1,149 @@
...
+  manifest: Pick<NodeManifest, 'display'> | undefined
+): boolean {
+  return manifest?.display.shape === 'container';
+}
+
+export function resolveContainerHandleGroups(
+  groups: ResolvedHandleGroup[]
+): ContainerHandleGroup[] {
+  return groups.map((group) => {
+    const boundary = group.boundary === 'inner' ? 'inner' : 'outer';
```
CalinaCristian: group.boundary is already `inner` or `outer`

packages/apollo-react/src/canvas/components/AddNodePanel/AddNodeManager.tsx
SreedharAvvari: Removed it since the loop will allow dragging nodes out of the container.

packages/apollo-react/src/canvas/components/LoopNode/LoopNode.helpers.ts
SreedharAvvari: the 'outer' was a fallback in this case, since group.boundary can be undefined
Made it more explicit by changing to 
```
    const boundary = group.boundary ?? 'outer';
```

packages/apollo-react/src/canvas/utils/createPreviewGraph.ts
```diff
@@ -0,0 +1,199 @@
...
+        .filter((node) => node.id !== PREVIEW_NODE_ID)
+        .map((node) => ({ ...node, selected: false })),
+      preview.node,
+    ]);
+
+    reactFlowInstance.setEdges((edges) => [
+      ...edges.filter((edge) => !isPreviewEdge(edge) && !removedEdgeIds.has(edge.id)),
+      ...preview.edges,
+    ]);
+  }, 0);
```
CalinaCristian, line 198: why setTimeout?

SreedharAvvari: I think its needed to defer preview insertion until React Flow finishes its connection-end cleanup, otherwise a preview created from the handle drag might be reset back.

Followed the same pattern as https://github.com/UiPath/apollo-ui/blob/main/packages/apollo-react/src/canvas/utils/createPreviewNode.ts#L314 for this one.

CalinaCristian (approved):
Follow up (should log tickets if it makes sense to take them):

- On empty loop, the + is not aligned with the start:
Image

 

- Initial placement of the loop node should be centered by the edge (edge should go straight).
Image

- Adding a new node from the initally added node adds it outside (this is blocker for this PR in case it behaves the same way in flow workbench and it's not just a storybook isssue):
https://github.com/user-attachments/assets/74a883dd-4526-489a-b7b0-76c5702440d3

- Is the break handle optional ? (it should be toggle from the property panel in flow-workbench).

SreedharAvvari:
> Follow up (should log tickets if it makes sense to take them):
> 
> * On empty loop, the + is not aligned with the start:
> 
> Image
> * Initial placement of the loop node should be centered by the edge (edge should go straight).
> 
> Image
> * Adding a new node from the initally added node adds it outside (this is blocker for this PR in case it behaves the same way in flow workbench and it's not just a storybook isssue):
>   [github.com/user-attachments/assets/74a883dd-4526-489a-b7b0-76c5702440d3](https://github.com/user-attachments/assets/74a883dd-4526-489a-b7b0-76c5702440d3)
> * Is the break handle optional ? (it should be toggle from the property panel in flow-workbench).

Thanks Cristian
For 1 - the prototype also has this issue, i agree with this & will follow up with design on the treatment for this

image

2 - I'm tracking this & fixed in https://github.com/UiPath/apollo-ui/pull/605
3 - looks like a bug, will fix this in the PR before merging
4 - yeah break will be optional & opt-in similar to error handle, will configure this in flow-workbench 

Ticket for the issues 1, 2 
https://uipath.atlassian.net/browse/MST-9193

PR #565, opened 2026-04-22
feat(apollo-vertex): add GroupMembershipGuard to shell registry [AGVSOL-2260]

frankkluijtmans:
# Notes

The purpose of this PR is to add a `GroupMembershipGuard` component to the shell along with a hook called 'useIsGroupMember`. These can be used to 
Screenshot 2026-04-22 at 16 32 11
Screenshot 2026-04-22 at 16 32 17
Screenshot 2026-04-22 at 16 32 06
block access based on group memberships.


apps/apollo-vertex/registry/shell/group-membership-guard.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+const MembershipCheckFailed = () => {
+  const { t } = useTranslation();
+  return (
+    <div className="flex h-screen items-center justify-center bg-background">
+      <div className="w-full max-w-md text-center">
+        <p className="text-muted-foreground mb-4">{t("access_check_failed")}</p>
+        <Button variant="outline" onClick={() => window.location.reload()}>
+          {t("retry")}
+        </Button>
```
alincadariu: is this an actual retry? we should probs say `reload page` or something

apps/apollo-vertex/registry/shell/group-membership-guard.tsx
```diff
@@ -0,0 +1,134 @@
...
+    return { hasError: true };
+  }
+
+  render() {
+    if (this.state.hasError) {
+      return <MembershipCheckFailed />;
+    }
+    return this.props.children;
+  }
+}
```
alincadariu: react-error-boundary? 🤔

alincadariu (changes requested):

apps/apollo-vertex/registry/shell/group-membership-guard.tsx
frankkluijtmans: That's fair enough 👌

apps/apollo-vertex/registry/shell/group-membership-guard.tsx
frankkluijtmans: Hmm, I'm thinking that maybe we shouldn't even have this error boundary in here. I'll just leave the Suspense.

alincadariu (approved):

PR #566, opened 2026-04-22
feat(apollo-vertex): expand Components menu in sidebar

frankkluijtmans:
# Notes

This changes the components section back into the list. The overview is still available by clicking `Overview` in that list. This looks like the most straightforward compromise we could do using `Nextra`.

Screenshot 2026-04-22 at 17 29 41

https://github.com/user-attachments/assets/6ed5b5ed-01a3-4642-a413-2e8111a0c06c


KokoMilev (changes requested):
In the overview page of the components we should also keep the nav bar.

KokoMilev (approved):

PR #577, opened 2026-04-24
Creilly11235/vertex shell sidebar provider migration

creilly11235:
Switch to using shadcn-based sidebar component. This PR swaps the hand built sidebar we currently have in the shell and adds the sub-nav capabilities. Demo docs and preview updates are in a separate pr to keep this one small. This PR is not dependent on #576 but without #576 it breaks the preview in the docs.


apps/apollo-vertex/registry/shell/shell-sidebar.tsx
```diff
@@ -1,38 +1,57 @@
...
   companyName,
   productName,
   variant,
   companyLogo,
   navItems,
-}: SidebarProps) => {
-  const [isCollapsed, setIsCollapsed] = useLocalStorage<boolean>({
-    key: SIDEBAR_COLLAPSED_KEY,
-    defaultValue: false,
-  });
```
alincadariu, line 31: up until now the state of the sidebar (open/close) was stored in the local storage, such that on a refresh the sidebar would be in the state before the refresh. now its not anymore. we should bring this behaviour back

alincadariu (changes requested):

apps/apollo-vertex/registry/shell/shell-sidebar.tsx
creilly11235: @alincadariu Updated!

alincadariu:
@petervachon let me know if we should start reviewing this PR, or if you are gonna start again as there are some drafts as follow ups

petervachon:
Hey @alincadariu — starting fresh. All the feedback from this PR has been addressed in a new clean stack:

- #681 — Sidebar migration to shadcn `SidebarProvider` with sub-nav (open, ready for review)
- #682 — User profile rework with fade-in text and minimal variant (draft, stacked on #681)

The localStorage persistence concern you flagged is resolved — collapsed state now uses the shadcn `SidebarProvider` cookie (`sidebar_state`), so it survives refreshes without touching localStorage at all. Closing this one out in favour of the new stack.

PR #588, opened 2026-04-24
fix(apollo-wind): DataTable headers stay put, resize feels grabbable, body scrolls independently

OlenaKosinska:
## Summary:

- Column headers now stay pinned as you scroll through long tables, you no longer lose track of which column is which
- Resizing a column narrow no longer causes its label to bleed into the next header
- The resize handle is easier to grab, more visible on hover, thicker while dragging
- New maxBodyHeight prop: the scrollbar sits inside the data rows only, keeping the header row visually separate.

## Also included:

- Keyboard users can scroll the table body without a mouse
- Two new Storybook stories: sticky header with 40 rows, and a resizable wide table to showcase the narrow-column fixes

### DIFF
--- a/packages/apollo-wind/src/components/ui/data-table.stories.tsx
+++ b/packages/apollo-wind/src/components/ui/data-table.stories.tsx
@@ -1061,6 +1061,180 @@ export const Density = {
 };
 
 // ============================================================================
+// Scrolling & Resizing
+// ============================================================================
+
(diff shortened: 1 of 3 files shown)


CalinaCristian (commented):
Image

we should have tooltips for elipsis otherwise users won't know what the text is

OlenaKosinska:
@CalinaCristian  - I fixed - tooltips are there.

CalinaCristian (approved):

PR #591, opened 2026-04-24
feat(apollo-vertex): add settings template

creilly11235:
Adds a settings page template composed of apollo vertex components.


0xr3ngar (changes requested):
same thing as https://github.com/UiPath/apollo-ui/pull/596#issuecomment-4335727066
@petervachon

needs platform integration

PR #596, opened 2026-04-24
feat(apollo-vertex): add customize appearance pattern (store + form + utilities)

creilly11235:
Add a pattern for allowing users to customize the appearance of their apps by updating the company name, logo, and primary and accent colors (custom colors lock the user into light mode only).


alincadariu:
@petervachon this one needs integration with the platform, as this PR currently stores it in local storage. so if user A changes company logo, it won't be visible to others (only to him). switched it to draft for now

cc: @0xr3ngar

petervachon:
Totally makes sense. When it comes to storage and relying on the platform, what is the best way to flag requests like this? I imagine our team may use local or session storage to work through the UX, but will need to work in the hooks later on. Thank you!

PR #612, opened 2026-04-28
feat(apollo-vertex): modular dashboard — foundation [1/5]

petervachon:
This is the first of a 5-PR series that introduces a modular, AI-augmented dashboard template. Each PR builds on the last — this one establishes everything needed to render the shell and load data.

## What's here

- `dashboard-data.ts` — the typed data model that drives the entire dashboard. Brand info, headline text, prompt suggestions, and the four insight card configs all live here. This is the primary file a consumer edits to adapt the template.
- `DashboardDataProvider` — context provider that makes the active dataset available to all dashboard components
- `DashboardContent` — top-level component wiring the header, layout, and the AI Assistant panel that slides in from the right
- Registry entry, route, and shell integration — the template is accessible at `/experiment/modular-dashboard`

The Vercel preview shows the full experience. The remaining components (charts, grid, interactions, glow) land in PRs #613–616.


apps/apollo-vertex/public/Autopilot_dark.svg
pieman1313, line 1: is this used?

apps/apollo-vertex/templates/dashboard/dashboard-data.ts
```diff
@@ -0,0 +1,228 @@
...
+  promptSuggestions: string[];
+  insightCards: [
+    InsightCardData,
+    InsightCardData,
+    InsightCardData,
+    InsightCardData,
+  ];
+}
+
+export const defaultDataset: DashboardDataset = {
```
pieman1313, line 51: are these used?

apps/apollo-vertex/templates/dashboard/DashboardLoading.tsx
```diff
@@ -0,0 +1,135 @@
...
+      setExiting(false);
+      setPhase("logo");
+    }
+  }, [triggerReplay]);
+
+  useEffect(() => {
+    if (phase === "done") return;
+
+    if (phase === "logo") {
+      const timer = setTimeout(() => {
```
pieman1313, line 102: why are we simulating loading? how will consumers handle these? will they use this file, or is it just for show?

pieman1313 (changes requested):
its really hard to see where this pr is going towards, without probably reviewing the rest of the dependent prs.
im not clear on how this will be used / or is just for show since I see a bunch of hardcodings, and consumers will need to adapt the code to fit their needs?

apps/apollo-vertex/public/Autopilot_dark.svg
petervachon: Yes — these SVGs are used across several dashboard components (the overview card, insight card actions, prompt bar, expanded drilldown, and the AI Assistant panel). They've been updated to use "AI Assistant" branding in the latest revision.

apps/apollo-vertex/templates/dashboard/dashboard-data.ts
petervachon: Good catch on visibility — `defaultDataset` is referenced as the `'default'` key in the `datasetPresets` map a few lines below (line 226). It's the seed data for the Loan Setup view that renders when no dataset is explicitly selected.

apps/apollo-vertex/templates/dashboard/DashboardLoading.tsx
petervachon: Fair question. The intent was to show what a dataset-switch transition could feel like — a logo → skeleton → content sequence that plays when AI generates or switches a dashboard view. That said, your concern about how consumers would use this is valid, and the trigger was also accidentally wired to a constant so it never actually fired. We've removed it from this revision — it will come back later as a more intentional, documented pattern.

petervachon:
Totally fair — this PR is the first layer of a stacked series, which makes it hard to evaluate in isolation. The hardcoded values (brand names, metric data, insight copy) are intentional: this is a reference template, not a drop-in component. The design pattern is for teams to fork it and replace the data layer with their own, similar to how shadcn/ui blocks work. We're also restructuring these PRs into feature-based slices (data layer → card components → layout → interactions → full integration) to make each piece independently reviewable. The next revision will make that intent clearer in the PR descriptions.

PR #623, opened 2026-04-29
feat(apollo-react): add ListItem.childrenLoading skeleton API [MST-9224]

Holychung:
Latest:
image

https://github.com/user-attachments/assets/2110a2c4-6ab9-493f-bdfc-6138baf501ef

## Summary

Apollo-side support for the loading-skeleton work tracked in MST-9164. The current `loading` prop on `<AddNodePanel>` / `<Toolbox>` is panel-wide — it dims and disables every row via the `.loading` CSS, which makes it impossible to show skeletons inside a single drilled-in level without breaking interactions everywhere else. This PR adds a per-item `childrenLoading` opt-in plus a `ListItemSkeleton` component that mirrors a real row's geometry, so consumers can preview a category's loading state without affecting the surrounding panel.

## Demo

> _insert demo here_

(Storybook → `Canvas/AddNodePanel` → 3 new stories: default skeletons, preemptive section headers, streaming → real content.)

## Changes

- **`ListItem.childrenLoading?: boolean | ChildrenLoadingSpec`** — when the user drills into an item with this flag, Apollo renders skeleton placeholder rows. The optional `sections` config preempts section headers so real items slot under their final header without layout shift.
- **`ListView.loadingSkeleton`** prop forwards the spec from `Toolbox` (read off `currentParentItem.childrenLoading`).
- **`ListItemSkeleton`** — circle + 2 stacked bars, replaces the legacy `<Skeleton h-8 w-full />` × 3 used by the `isLoading && items.length === 0` empty-state path.
- Skeletons are injected as `ListItem` entries into the virtualized list and grouped through the normal section pipeline. They append after real content while `childrenLoading` is set so multi-source categories keep showing "more on the way" until the last source resolves.
- `Toolbox` suppresses skeletons during search.
- Storybook: 3 new stories under `Canvas/AddNodePanel`.

## Flow

```mermaid
flowchart TD
    A[Consumer sets ListItem.childrenLoading] --> B[Toolbox.handleItemSelect drills in]
    B --> C[Toolbox passes currentParentItem.childrenLoading<br/>to ListView as loadingSkeleton]
    C --> D[ListView.buildSkeletonItems creates sentinels<br/>with optional section markers]
    D --> E[Skeletons merged into items array]
    E --> F[buildRenderedItems groups by section]
    F --> G[ListViewRow renders ListItemSkeleton<br/>for items with __listview_skeleton__ id]
    G --> H[Real items arrive → consumer drops childrenLoading<br/>→ skeletons removed, layout stable]
```

## Backward compatibility

- `loading: boolean` prop semantics unchanged
- Existing `ListItem` shape unchanged (`childrenLoading` is opt-in)
- Only behaviour change is the visual update to the skeleton row design (bar → row-shaped placeholder)

## Testing

- [x] `pnpm --filter @uipath/apollo-react test` — 97/97 pass (added 9 new tests)
- [x] Typecheck clean
- [x] Lint clean (no new warnings)
- [x] Storybook stories cover defaults, sections, and streaming swap
- [ ] Manual verification in consumer (flow-workbench MST-9164 / PR #1280)

## Consumer

flow-workbench MST-9164 / PR [UiPath/flow-workbench#1280](https://github.com/UiPath/flow-workbench/pull/1280) — drops the sentinel-row hack in `InternalAddNodePanel.tsx` and switches to `item.childrenLoading` once Apollo publishes.

### DIFF
--- a/packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
+++ b/packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
@@ -25,15 +25,22 @@ export interface ListItemIcon {
 /**
  * Configuration for the skeleton placeholder shown when an item is drilled
  * into and its children are still loading.
+ *
+ * For best perf, callers passing this as an object should memoize it —
+ * passing a fresh object each render rebuilds the skeleton items and
+ * cascades through the row virtualizer.
  */
 export type ChildrenLoadingSpec = {
-  /** Number of skeleton rows to render when no `sections` are provided. Defaults to 3. */
+  /**
+   * Number of skeleton rows. When `sections` is set, this is the per-section
+   * default for any section that omits its own `count`. Defaults to 3.
+   */
   count?: number;
   /**
    * Pre-render section headers + skeleton rows under each. Once real items
    * arrive with a matching `section`, they slot under the same header so the
-   * layout doesn't shift. Use this when you already know the final section
-   * structure (e.g. "Published", "In this solution").
+   * layout doesn't shift. Each section's `count` overrides the top-level
+   * `count` (or the built-in default of 3).
    */
   sections?: { name: string; count?: number }[];
 };
@@ -159,8 +166,8 @@ const ListViewRow = memo(
     const item = renderItem.item;
 
     // Skeleton sentinels render as the placeholder bar, never as a clickable
-    // row. They can't appear in `activeIndex` because Toolbox's keyboard
-    // navigation skips items injected by ListView itself.
+    // row. Toolbox's keyboard nav also skips them via the exported
+    // `isSkeletonItem` helper, so they never become the active descendant.
     if (isSkeletonItem(item)) {
       return (
         <div style={style}>
@@ -261,10 +268,20 @@ const ListItemSkeleton = () => (
   </div>
 );
 
-// Internal id prefix used to mark skeleton sentinel items injected by
-// `ListView`. Consumers must not produce ids with this prefix.
-const SKELETON_ID_PREFIX = '__listview_skeleton__';
-const isSkeletonItem = (item: ListItem): boolean => item.id.startsWith(SKELETON_ID_PREFIX);
+// Skeleton sentinel items are tracked by reference in this `WeakSet`, not by
+// an id prefix. This keeps the marker entirely internal — consumers can't
+// collide with it from their own `id` strings — and entries auto-expire when
+// the items become unreferenced.
+const skeletonItemRegistry = new WeakSet<ListItem>();
+
+/**
+ * True for skeleton sentinel rows produced by {@link ListView} (via the
+ * `loadingSkeleton` prop or `ListItem.childrenLoading`). Exposed so callers
+ * such as `Toolbox`'s keyboard navigation can skip these presentational rows.
+ */
+export function isSkeletonItem(item: ListItem): boolean {
+  return skeletonItemRegistry.has(item);
+}
 
 /**
  * Build the synthetic skeleton ListItem entries that get appended to the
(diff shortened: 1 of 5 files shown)


packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
```diff
@@ -196,6 +231,64 @@ interface ListViewProps<T extends ListItem> {
...
+
+/**
+ * Skeleton row used for the {@link ListItem.childrenLoading} drill-in path
+ * and the legacy `isLoading && items.length === 0` empty-state. Mirrors a
+ * real row's geometry — 32×32 icon + name + description — so the layout
+ * stays stable when real items take over.
+ */
+const ListItemSkeleton = () => (
+  <div
+    className="flex items-center gap-2.5 h-8 px-2"
```
BenGSchulz: This looks cool!
Wondering if we can just tweak to make it a little closer to the actual rows.
Removing the `px-2` on the wrapper and changing the icon skeleton to `rounded-lg` produces the top row in screenshot.

Image

packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
```diff
@@ -210,14 +326,33 @@ const ListViewInner = forwardRef(function ListView<T extends ListItem>(
...
+  );
+  const allItems = useMemo<T[]>(
+    () => (skeletonItems.length > 0 ? [...items, ...skeletonItems] : items),
+    [items, skeletonItems]
+  );
+
   const renderedItems = useMemo(
-    () => buildRenderedItems(items, enableSections),
-    [items, enableSections]
+    () => buildRenderedItems(allItems, enableSections),
```
BenGSchulz: Maybe instead of treating skeleton items like regular items, we can create a new `RenderItem` with type: 'skeleton'. `buildRenderedItems` can take the ChildrenLoadingSpec. That way we won't necessarily need the downstream checks for id or the cast above either maybe.

packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
david-rios-uipath: +1

packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
Holychung: Refactored: `RenderItem` now has a third variant `{ type: 'skeleton' }` and `buildRenderedItems` takes the `ChildrenLoadingSpec` directly. Dropped the `as T[]` cast, the WeakSet marker, and the `isSkeletonItem` helper — keyboard nav and click/hover paths just filter on `type === 'item'` like before. edf5e1d1.

packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
Holychung: Done — `px-2` dropped on the wrapper and the icon is now `rounded-lg`. Pushed in edf5e1d1.

Holychung: image

packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
```diff
@@ -32,49 +55,88 @@ export type ListItem<T = any> = {
...
-  for (const item of itemsWithoutSection) {
-    result.push({ type: 'item', item });
+  // Items without a section render at the top.
+  for (const item of itemsWithoutSection) result.push({ type: 'item', item });
+
+  // Build the union of section names from real items + preemptive skeleton
+  // sections, preserving insertion order. Real-item sections come first to
+  // match the existing behaviour.
+  const sectionNames: string[] = [];
+  const seenSections = new Set<string>();
```
BenGSchulz: Thinking that this section looping (and nested looping below) could be simplified a little. But not a blocker.

BenGSchulz (approved):
Looks awesome, big improvement, thanks!

packages/apollo-react/src/canvas/components/Toolbox/ListView.tsx
Holychung: Simplified — group items by section in one pass with a `Map`, look skeleton sections up by name, then walk a single `sectionNames` array. Drops the manual `seenSections` Set, the `O(N×S)` filter loop, and the per-section `.find()`. eef434a0.

PR #628, opened 2026-04-30
feat(apollo-vertex): ai-chat shell — branding, empty state, thinking indicator [1/5]

petervachon:
## What this does

Sets up the structural foundation for the AI Chat component — the outer shell users see before and during a conversation.

- **Empty state** with a centered heading and subtitle so the chat feels intentional rather than blank on first load
- **Autopilot branding** (gradient icon + title) in the header, consistent with the rest of the product
- **Thinking indicator** (`AiChatLoading`) shown while the AI is working, so users know something is happening
- **Sticky scroll** behavior — the view follows new messages as they stream in, with a "scroll to bottom" button when the user has scrolled up
- **Error banner** at the bottom with a retry action when a request fails

This PR contains no input, message rendering, or interaction logic — just the chrome around the conversation.

## Test plan
- [ ] Empty state renders correctly on first load
- [ ] Header shows gradient icon and title
- [ ] Thinking indicator appears while `isLoading` is true and no assistant text is streaming yet
- [ ] Scroll-to-bottom button appears when scrolled up; disappears when at bottom
- [ ] Error banner appears with retry button when `error` prop is set

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #630, opened 2026-04-30
feat(apollo-vertex): ai-chat message actions — copy, feedback, edit, regenerate [3/5]

petervachon:
## What this does

Adds the per-message action layer — the controls that let users interact with individual messages rather than just reading them.

- **Copy message** — copies the full text of any message to clipboard with a brief "Copied!" confirmation
- **Thumbs up / thumbs down** on assistant messages — surfaces a `onFeedback` callback so the host app can record quality signals
- **Edit user message** — tap the pencil on any user bubble to revise your message; the conversation rewinds to that point and re-sends, so the thread stays coherent
- **Regenerate** — reruns the last AI response without resending the user message, useful when the answer is off
- Actions appear on hover (desktop) or are always visible on touch, so they don't clutter the UI by default
- Shared state lives in `AiChatProvider` context so all message components can read it without prop-drilling

## Test plan
- [ ] Hover a message — action bar appears
- [ ] Copy copies the message text and shows "Copied!" briefly
- [ ] Thumbs up/down fires `onFeedback` with the message ID and type
- [ ] Edit pencil opens inline edit mode; submitting rewinds history and re-sends
- [ ] Regenerate button (on latest assistant message) fires `onRegenerate`


0xr3ngar (changes requested):
@petervachon since other PRs mentioned here were closed, can we close this PR as well?

petervachon:
Hey @0xr3ngar  this is part of an active stacked PR series ([1/5] #628 → [2/5] #629 → **[3/5] #630** → [4/5] #631 → [5/5] #632). Not ready to close — happy to keep it open for review!

0xr3ngar:
> Hey @0xr3ngar this is part of an active stacked PR series ([1/5] #628 → [2/5] #629 → **[3/5] #630** → [4/5] #631 → [5/5] #632). Not ready to close — happy to keep it open for review!

Ah okay since it is still active, please put this PR in draft and fix the merge conflicts of #629

petervachon:
Sounds good, went ahead and fixed the merge conflicts and updated the status to Draft for this and the following stacked PRs

VKravchuk (approved):

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message-actions.tsx
```diff
@@ -0,0 +1,163 @@
...
+                <Copy
+                  className="size-3.5 text-ai-chat-muted-foreground"
+                  aria-hidden="true"
+                />
+              )}
+            </button>
+          </TooltipTrigger>
+          <TooltipContent>{copyLabel}</TooltipContent>
+        </Tooltip>
+      )}
```
0xr3ngar: I would extract this copy state handling into a seperate component. This way the state  const [copied, setCopied] = useState(false); and the handler `handleCopy` don't live in this component.

Extract into a generic component

```tsx
import { useEffect, useState } from 'react';
import { useTranslation } from 'react-i18next';

type CopyToClipboardButtonProps = {
    copyValue: string;
};

export const CopyToClipboardButton = ({
    copyValue,
}: CopyToClipboardButtonProps) => {
    const [hasCopied, setHasCopied] = useState(false);

    const onClickHandler = async () => {
        await navigator.clipboard
            .writeText(copyValue)
            .then(() => setHasCopied(true))
            .catch(() => setHasCopied(false));
    };

    useEffect(() => {
        let timeout: NodeJS.Timeout;
        if (hasCopied) {
            timeout = setTimeout(() => {
                setHasCopied(false);
            }, 1500);
        }
        return () => clearTimeout(timeout);
    }, [hasCopied]);

    return (
        <Button
            testId={TestId.copyToClipboard}
            size="small"
            variant="text"
            color={hasCopied ? 'success' : 'primary'}
            onClick={async () => {
                await onClickHandler();
            }}
            startIcon={hasCopied ? <CheckIcon /> : <FileCopyIcon />}
        >
            {hasCopied ? t('copied') : t('copy')}
        </Button>
    );
};
```

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message-actions.tsx
```diff
@@ -0,0 +1,163 @@
...
+    : "opacity-0 group-hover/message:opacity-100 group-focus-within/message:opacity-100";
+
+  return (
+    <div
+      className={`flex items-center gap-0.5 -ml-[7px] transition-opacity ${visibilityClass}`}
+    >
+      {showCopy && (
+        <Tooltip>
+          <TooltipTrigger asChild>
+            <button
```
0xr3ngar: also why are we using <button> and not the <Button> component from the lib?

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message-actions.tsx
```diff
@@ -0,0 +1,163 @@
...
+            >
+              <Pencil
+                className="size-3.5 text-ai-chat-muted-foreground"
+                aria-hidden="true"
+              />
+            </button>
+          </TooltipTrigger>
+          <TooltipContent>{t("edit")}</TooltipContent>
+        </Tooltip>
+      )}
```
0xr3ngar: all these Buttons are literally the same with the only difference being the handler + icon + text. Extract into a shared Component and just pass the handler + text + icon

apps/apollo-vertex/registry/ai-chat/types.ts
```diff
@@ -0,0 +1,28 @@
+import type React from "react";
+
+export type MessageFeedbackType = "up" | "down";
```
0xr3ngar: I would call it positive | negative

apps/apollo-vertex/registry/ai-chat/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  /** Custom avatar for the user */
+  userAvatar?: React.ReactNode;
+  /** Show relative timestamps on messages */
+  showTimestamps: boolean;
+  /** Show hover action toolbar on messages */
+  showMessageActions: boolean;
+  /** Show copy button in message actions */
+  showCopyButton: boolean;
+  /** Whether the chat is currently loading */
+  isLoading: boolean;
```
0xr3ngar: why are these in the AiChatConfig type? these booleans make no sense to me. Why would a config have the isLoading boolean? Also the showCopyButton is sus? shouldn't these always be shown, same thing goes for showMessageActions

apps/apollo-vertex/registry/ai-chat/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  /** ID of the latest assistant message — its actions stay always-visible */
+  latestAssistantMessageId: string | null;
```
0xr3ngar: how is this used exactly? why do we need this lastMessageID? Shouldn't each chat have like a root id and all messages belong to it and we just fetch these

apps/apollo-vertex/registry/ai-chat/types.ts
```diff
@@ -0,0 +1,28 @@
...
+  /** Callback when the user gives thumbs up/down on an assistant message */
+  onFeedback?: (messageId: string, type: MessageFeedbackType) => void;
+  /** Callback to regenerate the last assistant response */
+  onRegenerate?: () => void;
+  /** Callback when the user saves an edited user message */
+  onEditMessage?: (messageId: string, content: string) => void;
```
0xr3ngar: Im confused about these as well, why do they live here? These don't seem relevant to the ChatConfig. When I read AiChatConfig I think about like the state, name, model name, thinking level, not side effects like feedback, regenerate and edit message

apps/apollo-vertex/registry/ai-chat/components/ai-chat.tsx
```diff
@@ -166,82 +184,187 @@ export function AiChat({
...
-        <div className="flex-1 flex flex-col items-center justify-center min-h-0">
-          <div className="w-full">
-            <div className="px-4 text-center mb-7">
-              {emptyState ?? defaultEmptyState}
+                {showLoadingIndicator && <AiChatLoading />}
+              </div>
             </div>
+
+            {!isStuck && (
+              <button
```
0xr3ngar: use <Button> component not raw <button>

apps/apollo-vertex/registry/ai-chat/components/ai-chat.tsx
```diff
@@ -166,82 +184,187 @@ export function AiChat({
...
+        {error && (
+          <div
+            role="alert"
+            aria-live="assertive"
+            className="relative z-0 mx-4 -mb-6 flex items-center gap-2 rounded-t-lg bg-destructive/10 px-[23px] pt-3 pb-6 text-sm text-destructive"
+          >
+            <AlertCircle className="h-4 w-4 flex-shrink-0" aria-hidden="true" />
+            <span className="flex-1">{error.message}</span>
+            {onRetry && (
+              <button
```
0xr3ngar: use <Button> component not raw <button>

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message.tsx
```diff
@@ -1,70 +1,217 @@
...
+                  e.preventDefault();
+                  handleSave();
+                }
+                if (e.key === "Escape") setIsEditing(false);
+              }}
+              className="w-full px-4 py-2 text-sm leading-6 rounded-2xl rounded-br-md bg-ai-chat-bubble-user text-ai-chat-bubble-user-foreground resize-none focus:outline-none border-2 border-primary/40 min-h-[60px]"
+              rows={3}
+            />
+            <div className="flex items-center gap-2">
+              <button
```
0xr3ngar: use <Button> component not raw <button>

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message.tsx
```diff
@@ -1,70 +1,217 @@
...
+                type="button"
+                onClick={() => setIsEditing(false)}
+                className="px-3 py-1 text-xs rounded-md hover:bg-muted text-muted-foreground transition-colors"
+              >
+                {t("cancel")}
+              </button>
+              <button
```
0xr3ngar: use <Button> component not raw <button>

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message.tsx
```diff
@@ -1,70 +1,217 @@
...
+    if (isEditing) {
+      return (
+        <motion.div
+          className="flex w-full justify-end"
+          initial={ENTRANCE_INITIAL}
+          animate={ENTRANCE_ANIMATE}
+          transition={ENTRANCE_TRANSITION}
+        >
+          <div className="flex flex-col items-end gap-2 w-[80%]">
+            <textarea
```
0xr3ngar: dont we also jhave a textarea component in the registry?

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message.tsx
```diff
@@ -1,70 +1,217 @@
...
+    setIsEditing(false);
+  };
+
+  // Streaming state — explicit prop wins, otherwise derive from chat-level isLoading
+  // and whether this is the latest assistant message currently being generated.
+  const isStreaming =
+    isStreamingProp ??
+    (config.isLoading &&
+      !isUser &&
+      config.latestAssistantMessageId === message.id);
```
0xr3ngar: this is super sus, this feels like a code smell. Why is the isStreaming based on the loading and if there is no user and if the last msg id is the current message id

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message.tsx
```diff
@@ -1,70 +1,217 @@
...
+  useEffect(() => {
+    if (isEditing && editTextareaRef.current) {
+      editTextareaRef.current.focus();
+      editTextareaRef.current.select();
+      const el = editTextareaRef.current;
+      requestAnimationFrame(() => {
+        el.scrollIntoView({ behavior: "smooth", block: "center" });
+      });
+    }
+  }, [isEditing]);
```
0xr3ngar: what is this? cant we create an authofocus ref and attach it to the button or textarea?

const autoFocus = (container: HTMLDivElement | null) => {
    if (!container) return;
    const focusable = container.querySelector<HTMLElement>(FOCUSABLE_SELECTOR);
    focusable?.focus();
};

export const useAutoFocusFirst = () => {
    return autoFocus;
};

const ref = useAutofucusFirst();

<div ref={ref}>...

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message.tsx
```diff
@@ -1,70 +1,217 @@
...
+  const displayContent = getDisplayText(message);
+
+  const [isEditing, setIsEditing] = useState(false);
+  const [editValue, setEditValue] = useState(displayContent);
+  const editTextareaRef = useRef<HTMLTextAreaElement>(null);
+
+  // Keep editValue in sync if message content changes externally (e.g. regenerate)
+  useEffect(() => {
+    if (!isEditing) setEditValue(displayContent);
+  }, [displayContent, isEditing]);
```
0xr3ngar: this is also super weird side effect that I think should never happen

0xr3ngar (changes requested):
@pieman1313 hey Razvan I feel like this is going in the wrong direction, the component and data structure seem super off. Can you take a deeper look into this and let me know if I'm going crazy or something, because the component structure feels super off, and we have to patch things with effects or contexts that I do not think should be there.

pieman1313:
@0xr3ngar valid callouts, @VKravchuk can you have a look? not sure about the provider altogether tbh

apps/apollo-vertex/registry/ai-chat/components/ai-chat.tsx
```diff
@@ -166,82 +184,187 @@ export function AiChat({
...
+              <div className="flex items-center gap-1.5 min-w-0">
+                <AutopilotGradientIcon
+                  size={21}
+                  className="flex-shrink-0"
+                  aria-hidden="true"
+                />
+                <span className="text-sm font-bold tracking-tight bg-clip-text text-transparent truncate pt-[2px] [background-image:linear-gradient(97.73deg,#5D4ED0_8.79%,#1076A0_91.48%)] dark:[background-image:linear-gradient(97.73deg,#9485F5_8.79%,#69C7DD_91.48%)]">
+                  {title}
+                </span>
+              </div>
```
pieman1313: this file is getting a bit large; can we extract?

apps/apollo-vertex/registry/ai-chat/components/ai-chat.tsx
```diff
@@ -166,82 +184,187 @@ export function AiChat({
...
+              {messages.length > 0 && (
+                <AlertDialog>
```
pieman1313: extract as well

apps/apollo-vertex/registry/ai-chat/components/ai-chat.tsx
```diff
@@ -166,82 +184,187 @@ export function AiChat({
...
-                    <AlertDialogAction onClick={() => onClearChat?.()}>
-                      {t("new_conversation")}
-                    </AlertDialogAction>
-                  </AlertDialogFooter>
-                </AlertDialogContent>
-              </AlertDialog>
-            )}
+                      {suggestion}
+                    </button>
+                  ))}
```
pieman1313: extract

pieman1313 (changes requested):

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message.tsx
```diff
@@ -1,70 +1,217 @@
...
+      className="flex w-full justify-start"
+      initial={ENTRANCE_INITIAL}
+      animate={ENTRANCE_ANIMATE}
+      transition={ENTRANCE_TRANSITION}
+    >
+      <div className="group/message flex flex-col gap-3 max-w-[85%]">
+        {displayContent &&
+          (isResponseFullyRevealed ? (
+            <AiChatMarkdown>{displayContent}</AiChatMarkdown>
+          ) : (
```
pieman1313: don't we want to render the markdown as its coming in? only at the end?

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message.tsx
```diff
@@ -1,70 +1,217 @@
...
+            initial={{ opacity: 0 }}
+            animate={{ opacity: 1 }}
+            transition={{ duration: 0.2, ease: [0.22, 1, 0.36, 1] }}
+          >
+            <AiChatMessageActions
+              content={displayContent}
+              messageRole="assistant"
+              isLatest={isLatestAssistant}
+              showCopy={config.showCopyButton}
+              {...((onFeedback ?? config.onFeedback)
```
pieman1313: why do we need to support both?

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message.tsx
```diff
@@ -1,70 +1,184 @@
...
   const { t } = useTranslation();
   const isUser = message.role === "user";
-  const displayName = assistantName ?? t("ai_assistant");
+  const displayContent = getDisplayText(message);
+
+  const [editValue, setEditValue] = useState<string | null>(null);
+  const isEditing = editValue !== null;
+
+  // Stable ref so React calls it once on textarea mount, not on every keystroke.
+  const editTextareaRef = useCallback((el: HTMLTextAreaElement | null) => {
```
pieman1313: don't need this

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message.tsx
```diff
@@ -1,70 +1,185 @@
...
 import { AiChatMarkdown } from "./ai-chat-markdown";
+import { AiChatMessageActions } from "./ai-chat-message-actions";
+
+// Quick, subtle entrance — fade + 8px slide up. Quartic ease-out for a soft settle.
+const ENTRANCE_INITIAL = { opacity: 0, y: 8 };
+const ENTRANCE_ANIMATE = { opacity: 1, y: 0 };
+const ENTRANCE_TRANSITION = {
+  duration: 0.22,
+  ease: [0.22, 1, 0.36, 1] as const,
+};
```
0xr3ngar: extract this into constants, since all values are duplicated in other files

apps/apollo-vertex/registry/ai-chat/components/ai-chat-message.tsx
```diff
@@ -1,70 +1,185 @@
...
+    el.scrollIntoView({ behavior: "smooth", block: "center" });
+  });
+}
 
 interface AiChatMessageProps {
   message: UIMessage;
-  assistantName?: string;
   children?: ReactNode;
+  /** Whether this message is currently being streamed */
+  isStreaming?: boolean;
```
0xr3ngar: naming is misleading on this since it only hides assistant actions.

0xr3ngar (changes requested):
OK I started reviewing the code again but I see smells everywhere :panic:

rn the data flow feels pretty shady. We pass `messages` into `<AiChat />`, but then the consumer also maps over `messages` and renders `<AiChatMessage />` manually. On top of that, `getAiChatMessageProps` re-derives streaming/latest/action state, `AiChatMessage` re-derives display text/content visibility, and the template owns tool rendering as children

So the same source of truth is being interpreted in a bunch of different places
1. `AiChat` handles empty state, loading, header, copy conversation, scroll, etc..
2. `getAiChatMessageProps` handles latest message/streaming/action visibility
3. `AiChatMessage` handles display text and whether the message is renderable
4. the template handles message mapping and tool rendering

that feels like a pretty big code smell, we’ve kind of split ownership in the worst middle-ground way- `AiChat` both owns and doesn’t own messages at the same time :/

It also creates a bunch of perf smells. During streaming, we’re constantly re-rendering/recomputing across multiple layers: the template maps every message again, tool parts are checked/rendered again, message props are re-derived, display text is rebuilt, and `AiChat` also walks messages for shell-level state like copy conversation/loading

I don’t think the current pattern/architecture is healthy at all. I think we need to go back to the drawing board and create a better architecture where message state is derived in one place instead of being re-interpreted and re-computed constantly across multiple components

@pieman1313

0xr3ngar (approved):

pieman1313 (approved):

PR #632, opened 2026-04-30
feat(apollo-vertex): ai-chat text selection — Ask AI floating pill + quote chip [5/5]

petervachon:
## What this does

Lets users ask the AI about specific text they've selected — a natural way to go from reading a response to asking a follow-up about a particular part of it.

- **"Ask AI" floating pill** — appears above any selected text within the chat container, with the Autopilot icon and a gradient background to make it feel intentional
- **Quote chip** in the input — when the user clicks "Ask AI", the selected text appears as a small truncated chip above the textarea so they can see the context they're referencing before they type their question
- **Dismiss** — clicking × on the chip removes the quoted context; submitting the message clears it automatically
- Scoped to the chat container — selecting text elsewhere on the page (browser UI, other panels) doesn't trigger the pill
- Enabled via `enableTextSelection` prop (off by default) so teams can opt in only where it makes sense

## Test plan
- [ ] Select text in an assistant message — "Ask AI" pill appears above the selection
- [ ] Click pill — quote chip appears in input with the selected text, input is focused
- [ ] Long quoted text truncates with ellipsis rather than wrapping to two lines
- [ ] × on chip clears the quote
- [ ] Submitting clears the chip and sends the message
- [ ] Selecting text outside the chat container does not show the pill
- [ ] `enableTextSelection={false}` (default) — pill never appears


frankkluijtmans:
I think this is part of a PR stack? Seems like `[feat/ai-chat-4-attachments](https://github.com/UiPath/apollo-ui/tree/feat/ai-chat-4-attachments)` needs to go in first. Let's keep this in draft until then.

apps/apollo-vertex/registry/ai-chat/components/ai-chat-selection-menu.tsx
```diff
@@ -0,0 +1,103 @@
...
+}
+
+function deriveMenuState(
+  selection: Selection | null,
+  container: HTMLElement | null,
+): MenuState | null {
+  if (!container) return null;
+  if (!selection || selection.isCollapsed || selection.rangeCount === 0) {
+    return null;
+  }
```
frankkluijtmans: NIT: can be one if statement

frankkluijtmans (changes requested):

frankkluijtmans (approved):

PR #637, opened 2026-05-01
feat(apollo-react): add Toolbox quickActions slot

venkateshcherukuri:
## Summary

Adds an optional `quickActions` prop to `Toolbox` that renders a row of icon shortcuts inside the popover chrome — a leading group plus an optional trailing group separated by a vertical divider. Canvas consumers can now absorb fast-access icon strips into the existing Toolbox padding rather than stacking them above the popover where they get clipped against canvas edges.

Motivation: PO.Frontend's BPMN canvas-element picker (`HandleQuickAddPopover`) currently renders an icon strip *outside* `<Toolbox>` via a separate `<IconRowBorder>` wrapper. When the picker opens near the top of the canvas, that extra row gets clipped by the canvas's `overflow: hidden`. Pulling the row into Toolbox's chrome reduces total popover height by ~50 px and lets floating-ui's flip placement fit more often.

## Demo
image

## What's in the diff

- **`Toolbox.tsx`** — new optional prop `quickActions?: ToolboxQuickAction[]`. When set, renders `<QuickActionsRow>` inside the existing `<Column>` above the title. No-op when omitted.
- **`QuickActionsRow.tsx`** — new component. Renders an apollo-wind ghost-`Button` per action wrapped in `CanvasTooltip`. Trailing actions render after a vertical separator; leading-only and trailing-only configurations are both supported.
- **`Toolbox.styles.ts`** — `min-height: 300px → 200px` on `AnimatedContainer` + `AnimatedContent`. The 300 floor was never enforced for current consumers (Toolbox at 440 px height has 324 px available for the inner list — flex:1 already exceeds 300). 200 only matters when `quickActions` are added; existing consumers see no functional change.
- **`ConnectorHandleIcon.tsx`** — new BPMN-style sequence-flow arrow (dot + line + arrowhead, 36×24 viewBox). Used by the trailing "Connect" action in the BPMN picker.
- **`Toolbox.stories.tsx`** — new `Default` and `WithQuickActions` stories so the slot is discoverable in the design system.
- **`QuickActionsRow.test.tsx`** — 10 tests covering leading-only, trailing-only, separator presence, click/hover wiring, disabled state.

## Backwards compatibility

- `quickActions` is optional — existing consumers (e.g. `AddNodePanel` in apollo-ui, multiple Toolbox usages in PO.Frontend) are unaffected.
- The `min-height` reduction is functionally a no-op for current consumers since `flex: 1` always provides ≥ 300 px at the standard `TOOLBOX_HEIGHT` of 440 px. 200 only kicks in when quickActions add ~56 px above the inner list.
- 1452/1452 apollo-react tests pass.

## Test plan

- [x] `pnpm test` in `packages/apollo-react` — 1452/1452 passing
- [x] Manually verified in PO.Frontend with a local apollo-react build — BPMN canvas-element picker renders the icon row inside Toolbox chrome, popover height fits within canvas bounds, click/hover wiring intact
- [x] Storybook story renders both `Default` and `WithQuickActions` variants
- [ ] Reviewer to sanity-check the storybook visuals on Chromatic / preview deploy

### DIFF
--- a/packages/apollo-react/src/canvas/components/Toolbox/QuickActionsRow.tsx
+++ b/packages/apollo-react/src/canvas/components/Toolbox/QuickActionsRow.tsx
@@ -1,4 +1,3 @@
-import styled from '@emotion/styled';
 import { Button } from '@uipath/apollo-wind';
 import { memo, type MouseEvent, type ReactNode } from 'react';
 import { TOOLBOX_PADDING_X } from '../../constants';
@@ -18,26 +17,6 @@ export type ToolboxQuickAction = {
 interface QuickActionsRowProps {
   actions: ToolboxQuickAction[];
 }
-
-const Container = styled.div`
-  display: flex;
-  align-items: center;
-  justify-content: center;
-  gap: 12px;
-  min-height: 44px;
-  padding: 0 ${TOOLBOX_PADDING_X}px 8px;
-  margin: 0 -${TOOLBOX_PADDING_X}px;
-  border-bottom: 1px solid var(--canvas-border-de-emp);
-  flex-shrink: 0;
-`;
-
-const Separator = styled.div`
-  width: 1px;
-  align-self: stretch;
-  margin: 0 4px;
-  background: var(--canvas-border-de-emp);
-  flex-shrink: 0;
-`;
 
 function QuickActionButton({ action, wide }: { action: ToolboxQuickAction; wide?: boolean }) {
   return (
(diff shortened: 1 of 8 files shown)


packages/apollo-react/src/canvas/components/Toolbox/QuickActionsRow.tsx
```diff
@@ -0,0 +1,82 @@
...
+  onMouseLeave?: () => void;
+  trailing?: boolean;
+  disabled?: boolean;
+};
+
+interface QuickActionsRowProps {
+  actions: ToolboxQuickAction[];
+}
+
+const Container = styled.div`
```
CalinaCristian: can you use this [skill](https://github.com/UiPath/apollo-ui/blob/main/.claude/skills/migrate-canvas-styled-to-tailwind/SKILL.md) to use wind classnames instead?

CalinaCristian (approved):

PR #642, opened 2026-05-04
feat(apollo-react): add readOnly state to stickyNote [MST-9331]

ayush578:
added readOnly state for stickyNote

https://github.com/user-attachments/assets/a3762887-03e8-4995-b22c-b6a98917dd6b

### DIFF
--- a/packages/apollo-react/src/canvas/components/StickyNoteNode/StickyNoteNode.tsx
+++ b/packages/apollo-react/src/canvas/components/StickyNoteNode/StickyNoteNode.tsx
@@ -42,6 +42,7 @@ export interface StickyNoteNodeProps extends NodeProps {
   data: StickyNoteData;
   placeholder?: string;
   renderPlaceholderOnSelect?: boolean;
+  readOnly?: boolean;
   onContentChange?: (content: string) => void;
   onColorChange?: (color: StickyNoteColor) => void;
   onResize?: (width: number, height: number) => void;
@@ -57,12 +58,13 @@ const StickyNoteNodeComponent = ({
   dragging,
   placeholder = 'Add text',
   renderPlaceholderOnSelect = false,
+  readOnly = false,
   onContentChange,
   onColorChange,
   onResize,
 }: StickyNoteNodeProps) => {
   const { updateNodeData, deleteElements } = useReactFlow();
-  const [isEditing, setIsEditing] = useState(data.autoFocus ?? false);
+  const [isEditing, setIsEditing] = useState(!readOnly && (data.autoFocus ?? false));
   const [isResizing, setIsResizing] = useState(false);
   const [isColorPickerOpen, setIsColorPickerOpen] = useState(false);
   const [localContent, setLocalContent] = useState(data.content || '');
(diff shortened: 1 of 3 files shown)


packages/apollo-react/src/canvas/components/StickyNoteNode/StickyNoteNode.tsx
```diff
@@ -57,12 +58,13 @@ const StickyNoteNodeComponent = ({
   dragging,
   placeholder = 'Add text',
   renderPlaceholderOnSelect = false,
+  readOnly = false,
```
SreedharAvvari, line 61: [P1] AgentFlow does not pass this prop, so the production stickyNote node type still renders editable notes in view mode. BaseCanvas disables dragging/deleting for view mode, but selected sticky notes can still enter edit mode because readOnly stays false.

SreedharAvvari: Suggested change:
Pass `readOnly={mode !== "design"}` from the `AgentFlow` stickyNote node renderer and add `mode` to that `useMemo` dependency list.

```tsx
<StickyNoteNode
  {...props}
  readOnly={mode !== "design"}
  placeholder={
    (stickyNoteNodeTranslations ?? DefaultStickyNoteNodeTranslations).placeholder
  }
  renderPlaceholderOnSelect={true}
/>
```

That makes the new state apply to AgentFlow, not only direct consumers and Storybook.

packages/apollo-react/src/canvas/components/StickyNoteNode/StickyNoteNode.tsx
```diff
@@ -291,53 +302,57 @@ const StickyNoteNodeComponent = ({
...
+              minHeight={minHeight}
+              onResizeStart={handleResizeStart}
+              onResizeEnd={handleResizeEnd}
+            >
+              <ResizeHandle selected={selected} cursor="nwse-resize" />
+            </NodeResizeControl>
+          </>
+        )}
 
         <StickyNoteContainer
```
SreedharAvvari, line 357: [P1] The read-only node still owns pointer events through `StickyNoteContainer`, so an overlapping note can continue to swallow clicks before stages underneath receive them. That leaves MST-9292's "stages not clickable" case unfixed even if the edit UI is hidden.

SreedharAvvari: Suggested change:
Carry `readOnly` into the container styles and let the non-interactive sticky-note surface click through while preserving explicit link targets.

```tsx
<StickyNoteContainer
  backgroundColor={colorWithAlpha}
  borderColor={color}
  isEditing={isEditing}
  selected={selected}
  readOnly={readOnly}
  onDoubleClick={handleDoubleClick}
>
```

```ts
cursor: ${(props) =>
  props.readOnly ? "default" : props.isEditing ? "text" : "move"};
pointer-events: ${(props) => (props.readOnly ? "none" : "auto")};
```

Please verify the read-only story still opens markdown links and that clicks reach a stage covered by a note.

packages/apollo-react/src/canvas/components/StickyNoteNode/StickyNoteNode.tsx
```diff
@@ -103,24 +105,32 @@ const StickyNoteNodeComponent = ({
     }
   }, [selected, isResizing]);
 
+  useEffect(() => {
+    if (readOnly) {
```
SreedharAvvari, line 109: [P2] Read-only rendering is not passive yet: the autofocus effect just above this still calls `updateNodeData(id, { autoFocus: false })` even when `readOnly` is true. A consumer that renders a read-only preview over editable data would lose the pending autofocus state just by mounting the preview.

SreedharAvvari: Suggested change:
Gate the autofocus cleanup on editability and include `readOnly` in the effect dependencies.

```ts
if (!readOnly && data.autoFocus) {
  updateNodeData(id, { autoFocus: false });
}
```

That keeps read-only render passive while preserving the current editable behavior.

SreedharAvvari (approved):

PR #655, opened 2026-05-07
feat(apollo-vertex): extend data-table with plain option

tarasiuk-o:
- added `plain` prop to render table without `GLASS_CLASSES`

### DIFF
--- a/apps/apollo-vertex/registry/data-table/data-table.tsx
+++ b/apps/apollo-vertex/registry/data-table/data-table.tsx
@@ -88,6 +88,7 @@ interface DataTableProps<TData, TValue> {
   renderExpandedRow?: (row: Row<TData>) => React.ReactNode;
   getRowClassName?: (row: TData) => string;
   skeletonColumnWidths?: string[];
+  plain?: boolean;
 }
 
 function DataTable<TData, TValue>({
@@ -123,6 +124,7 @@ function DataTable<TData, TValue>({
   renderExpandedRow,
   getRowClassName,
   skeletonColumnWidths,
+  plain = false,
 }: DataTableProps<TData, TValue>) {
   const { t } = useTranslation();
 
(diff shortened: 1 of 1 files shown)


apps/apollo-vertex/registry/data-table/data-table.tsx
```diff
@@ -207,7 +209,7 @@ function DataTable<TData, TValue>({
       <div
         ref={containerRef}
         className={cn(
-          GLASS_CLASSES,
+          plain ? "" : GLASS_CLASSES,
```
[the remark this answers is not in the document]
0xr3ngar: ai has a point

apps/apollo-vertex/registry/data-table/data-table.tsx
```diff
@@ -123,6 +124,7 @@ function DataTable<TData, TValue>({
   renderExpandedRow,
   getRowClassName,
   skeletonColumnWidths,
+  plain = false,
```
0xr3ngar, line 127: what does plain mean, can we give it a better name?

0xr3ngar (changes requested):

apps/apollo-vertex/registry/data-table/data-table.tsx
tarasiuk-o: @0xr3ngar  If the constant is named `GLASS_CLASSES`, we can make the option name clearer — for example, a boolean `glass`, a union `variant: "glass" | "plain"`, `glassClasses`, or `transparent`.

The core issue is that we need a way to render the data table without `GLASS_CLASSES`. WDYT?

ruudandriessen: Why do we want to render it without? This is the current design no?

sdignum: When data tables are the main item in a page (eg our loans table view), glass class is appropriate. Sometimes we need a data table without the glass class when it's inside a another container in a page (and having the glass classes would be too much). 
image

ruudandriessen (approved):

0xr3ngar (approved):

PR #663, opened 2026-05-07
fix(apollo-react): support nested loop outer handle insertion

SreedharAvvari:
## Summary

This PR tackles two loop-handle issues:

- Support nested loop outer handle insertion: dragging from a nested loop outer output now resolves the add-node preview inside the parent loop sequence. The preview code uses an `allowedContainerId` guard so nested loop inserts stay scoped to the parent container.
- Error handle support for loop: LoopNode now honors runtime `data.handleConfigurations`, while inheriting manifest boundary and custom position metadata. It also passes `nodeId` into handle resolution, so dynamic handles such as error outputs can render and behave consistently with loop/container handles.

## Companion Flow Workbench Change

- Paired with Flow Workbench PR https://github.com/UiPath/flow-workbench/pull/1481, which preserves container metadata (`boundary` and `customPositionAndOffsets`) when Flow Workbench resolves manifest handles into runtime `data.handleConfigurations`.

## Flow

```mermaid
flowchart LR
  A[Drag from loop handle] --> B{Nested loop outer handle?}
  B -- Yes --> C[Scope preview to parent loop]
  B -- No parent --> D[Keep top-level outer container ignored]
  C --> E{Existing next node?}
  E -- Yes --> F[Insert preview into sequence]
  E -- No --> G[Append to parent continuation]
  H[Loop runtime handle config] --> I[Merge manifest boundary and offsets]
  I --> J[Resolve normal and error handles]
```

## Validation

`pnpm --filter @uipath/apollo-react exec vitest --run src/canvas/components/LoopNode/LoopNode.helpers.test.ts`

Result: 1 file passed, 15 tests passed.

### DIFF
--- a/packages/apollo-react/src/canvas/components/LoopNode/LoopNode.helpers.test.ts
+++ b/packages/apollo-react/src/canvas/components/LoopNode/LoopNode.helpers.test.ts
@@ -325,6 +325,99 @@ describe('resolveContainerAddNodePreview', () => {
     expect(previewOverrides).toBeNull();
   });
 
+  it('uses the parent container sequence for nested container outer outputs', () => {
+    const innerLoopNode: Node = {
+      id: 'inner-loop',
+      type: 'loop',
+      parentId: loopNode.id,
+      position: { x: 160, y: 112 },
+      style: { width: 560, height: 320 },
+      data: {},
+    };
+    const targetNode: Node = {
+      id: 'task-1',
+      type: 'task',
+      parentId: loopNode.id,
+      position: { x: 800, y: 112 },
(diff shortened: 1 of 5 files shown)


CalinaCristian (approved):

CalinaCristian (commented):
would be great to have storybook updates reflecting this.

SreedharAvvari:
> would be great to have storybook updates reflecting this.

Done, added story in https://github.com/UiPath/apollo-ui/pull/663/commits/561e475544bb8e7d370d4957a89e9cde8424e3da

PR #673, opened 2026-05-11
feat(apollo-vertex): persist rows per page in data table [AGVSOL-3072]

angeloaltamiranom:
## Context
This comes from a request by Medtronic to be able to persist the page size selection. So in the same way we persist the column selection im thinking we should also persist the page size.

## SS
image

### DIFF
--- a/apps/apollo-vertex/registry/use-data-table/useDataTable.tsx
+++ b/apps/apollo-vertex/registry/use-data-table/useDataTable.tsx
@@ -3,13 +3,13 @@
 import type {
   ColumnDef,
   ColumnFiltersState,
-  PaginationState,
   RowSelectionState,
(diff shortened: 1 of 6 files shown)


apps/apollo-vertex/registry/use-data-table/useDataTable.tsx
0xr3ngar, line 1: you need to include the `usePersistedPageSize` hook to the registry dep files as well

0xr3ngar (changes requested):

alincadariu (commented):
@angeloaltamiranom the registry ci check is failing 👀

frankkluijtmans (approved):

0xr3ngar (approved):

PR #702, opened 2026-05-14
feat(apollo-react): add localization to stage nodes strings [MST-9646]

ayush578:
added localization to strings in stage nodes
You can switch to pseudo language from controls and check the effect on strings

https://github.com/user-attachments/assets/ed2c08b8-3b5b-433e-bf1f-7e6fe3f53f95

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.tsx
@@ -11,6 +11,7 @@ import { StageChip, StageHeader } from './StageNode.styles';
 import type { StageNodeProps, StageSlaIcon, StageStatus } from './StageNode.types';
 import { StageHeaderChipType } from './StageNode.types';
 import { StageTitleInput } from './StageTitleInput';
+import { useStageNodeLabels } from './useStageNodeLabels';
 
 const SLA_ICON_CONFIG: Record<StageSlaIcon, { icon: string; iconColor: string }> = {
   warning: {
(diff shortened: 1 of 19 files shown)


packages/apollo-react/src/canvas/components/Toolbox/Toolbox.tsx
```diff
@@ -154,6 +159,7 @@ export function Toolbox<T>({
   fullHeight = false,
   quickActions,
   renderEmptyState,
+  searchPlaceholder = 'Search',
```
kittyyueli: default to nothing

packages/apollo-react/src/canvas/components/Toolbox/Toolbox.tsx
```diff
@@ -647,7 +653,7 @@ export function Toolbox<T>({
           value={search}
           onChange={handleSearch}
           clear={clearSearch}
-          placeholder="Search"
+          placeholder={searchPlaceholder}
```
kittyyueli, line 653:
```suggestion
          placeholder={searchPlaceholder || translatedString}
```

packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.tsx
```diff
@@ -44,11 +45,12 @@ const StageNodeHeaderInner = ({
   status?: StageStatus;
   handleTaskAddClick: (event: React.MouseEvent) => void;
 }) => {
+  const labels = useStageNodeLabels();
   const {
     id,
     stageDetails,
     execution,
-    addTaskLabel = 'Add task',
+    addTaskLabel = labels.addTask,
```
kittyyueli: I think its better to reemove the addTaskLabel prop completely. I think we are generally always using the same label

packages/apollo-react/src/canvas/components/Toolbox/Toolbox.tsx
```diff
@@ -154,6 +159,7 @@ export function Toolbox<T>({
   fullHeight = false,
   quickActions,
   renderEmptyState,
+  searchPlaceholder,
```
kittyyueli: same, I don't think we should be passing this in~ we should use the labels hook

kittyyueli (approved):

PR #703, opened 2026-05-14
fix(apollo-react): stage title render [MST-9623] [MST-9699] [MST-9131]

ayush578:
fixed the overall rendering of stageTitle input.
Major focus is to get clickable area only equal to the text not the total width

https://github.com/user-attachments/assets/110b5d96-8c6a-4a02-a94d-32e823a9f2a0

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.tsx
@@ -1,15 +1,16 @@
 import { Icon, Padding, Spacing } from '@uipath/apollo-core';
 import { Row } from '@uipath/apollo-react/canvas/layouts';
-import { Button, cn } from '@uipath/apollo-wind';
-import { memo, useCallback, useEffect, useRef, useState } from 'react';
+import { Button } from '@uipath/apollo-wind';
+import { memo } from 'react';
 import { EntryConditionIcon, ExitConditionIcon, ReturnToOriginIcon } from '../../icons';
 import { CanvasIcon } from '../../utils/icon-registry';
 import { CanvasTooltip } from '../CanvasTooltip';
 import { ExecutionStatusIcon } from '../ExecutionStatusIcon';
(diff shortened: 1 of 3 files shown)


packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.tsx
```diff
@@ -137,20 +131,30 @@ const StageNodeHeaderInner = ({
...
+                    value={label}
+                    ref={stageTitleRef}
+                    isEditing={isStageTitleEditing}
+                    onInput={handleStageTitleChange}
+                    onKeyDown={handleStageTitleKeyDown}
+                    onBlur={handleStageTitleBlurToSave}
+                  />
+                ) : (
+                  <span
+                    className={cn(
```
kittyyueli: why are we creating a span? I think all the changes should be in the StageTitleInput component itself~ Also, right now, the drag is happening when its being edited~

kittyyueli:
there's also a tiny layout shift when we edit

packages/apollo-react/src/canvas/components/StageNode/StageTitleInput.tsx
```diff
@@ -0,0 +1,103 @@
...
+  }, [isEditing]);
+
+  const handleChange = useCallback((e: React.FormEvent<HTMLInputElement>) => {
+    setLabel((e.target as HTMLInputElement).value);
+  }, []);
+
+  const saveStageTitle = useCallback(() => {
+    setIsEditing(false);
+    if (!onChange) return;
+    const finalLabel = label.trim() === '' ? 'Untitled Stage' : label;
```
kittyyueli, line 41: lets translate the string here

kittyyueli (approved):

packages/apollo-react/src/canvas/components/StageNode/StageTitleInput.tsx
ayush578: Will do it in localization one
thanks

PR #707, opened 2026-05-14
feat(apollo-wind): add maxRef to slider for dynamic max from another field

Holychung:
## Summary

`SliderFieldMetadata.max` was a static number — fine for fixed ranges (e.g. `temperature: 0..1`) but a poor fit when the upper bound depends on another field. The motivating case: an agent's `maxTokenPerResponse` slider should cap at the selected model's `maxTokens`, but the available models (and their token caps) only exist at runtime.

This PR adds an additive, back-compat `maxRef` option that lets the slider read its `max` from another form field at render time.

## Changes

- **`form-schema.ts`** — Adds `FieldRef` interface and `SliderFieldMetadata.maxRef?: FieldRef`. Existing static `max?: number` is untouched, so consumers using the plain numeric form keep working with no migration.
- **`field-renderer.tsx`** — Extracts the `case 'slider'` body into a `SliderField` sub-component (slider needs hooks now). It calls `useFormContext().watch(maxRef.fromField)` and resolves max as: watched value if finite and positive → `maxRef.fallback` → static `max` → 100. If the resolved max drops below the current form value (e.g. user switches to a model with a lower cap), the value is clamped via `onChange`.
- **`field-renderer.test.tsx`** — 6 new tests covering static max, default fallback, maxRef resolution, watched-field-missing fallback, watched-field-non-numeric fallback, and maxRef precedence over static max.

## Usage

```ts
{
  name: 'inputs.maxTokenPerResponse',
  type: 'slider',
  label: 'Max. tokens per response',
  min: 0,
  step: 1,
  // Watches `inputs.modelMaxTokens` — caller wires this hidden field via
  // setValue() whenever the model picker selection changes.
  maxRef: { fromField: 'inputs.modelMaxTokens', fallback: 16384 },
}
```

## Test plan

- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm test` (apollo-wind) — 909 passed (was 903; +6 new)
- [x] `pnpm lint` — no new diagnostics introduced
- [ ] Reviewer sanity check: static `max: 100` slider still renders identically (covered by existing tests but worth a manual eyeball in Storybook)

### DIFF
--- a/packages/apollo-wind/src/components/forms/field-renderer.tsx
+++ b/packages/apollo-wind/src/components/forms/field-renderer.tsx
@@ -660,9 +660,7 @@ function SliderField({ field, formField, error, disabled, required }: SliderFiel
     <FormField>
       <div className="flex justify-between">
         <FormLabel required={required}>{field.label}</FormLabel>
-        <span className="text-sm text-muted-foreground">
-          {displayValue as React.ReactNode}
-        </span>
+        <span className="text-sm text-muted-foreground">{displayValue as React.ReactNode}</span>
       </div>
       <Slider
         value={[(displayValue as number) ?? field.min ?? 0]}
--- a/packages/apollo-wind/src/components/forms/field-renderer.tsx
+++ b/packages/apollo-wind/src/components/forms/field-renderer.tsx
@@ -632,11 +632,8 @@ interface SliderFieldProps {
 
 function SliderField({ field, formField, error, disabled, required }: SliderFieldProps) {
   const { watch } = useFormContext();
-  // watch() requires a defined name; pass a dummy `__none__` path when
-  // maxRef is absent and ignore the value. react-hook-form returns undefined
-  // for unknown fields, which is what resolveSliderMax expects.
-  const watchedMax = watch(field.maxRef?.fromField ?? '__apolloWindSliderMaxRefNone__');
-  const resolvedMax = resolveSliderMax(field, field.maxRef ? watchedMax : undefined);
+  const watchedMax = field.maxRef ? watch(field.maxRef.fromField) : undefined;
+  const resolvedMax = resolveSliderMax(field, watchedMax);
 
   // Clamp the form value if the resolved max drops below it (e.g. user
   // switched to a model with a lower token cap).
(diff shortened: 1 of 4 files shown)


snuziale (approved):

CalinaCristian (approved):

packages/apollo-wind/src/components/forms/field-renderer.tsx
```diff
@@ -619,3 +609,87 @@ function FormError({ children }: { children?: React.ReactNode }) {
...
+  // switched to a model with a lower token cap).
+  const prevMaxRef = useRef(resolvedMax);
+  useEffect(() => {
+    if (resolvedMax === prevMaxRef.current) return;
+    prevMaxRef.current = resolvedMax;
+    const v = formField.value;
+    if (typeof v === 'number' && v > resolvedMax) {
+      formField.onChange(resolvedMax);
+    }
+  }, [resolvedMax, formField]);
```
david-rios-uipath: nit: could we be more granular w/ the `formField` dep since it's likely a new object on each render?
e.g. `[resolvedMax, formfField.value, formField.onChange]`

packages/apollo-wind/src/components/forms/field-renderer.tsx
```diff
@@ -619,3 +609,87 @@ function FormError({ children }: { children?: React.ReactNode }) {
...
+  disabled: boolean;
+  required: boolean;
+}
+
+function SliderField({ field, formField, error, disabled, required }: SliderFieldProps) {
+  const { watch } = useFormContext();
+  // watch() requires a defined name; pass a dummy `__none__` path when
+  // maxRef is absent and ignore the value. react-hook-form returns undefined
+  // for unknown fields, which is what resolveSliderMax expects.
+  const watchedMax = watch(field.maxRef?.fromField ?? '__apolloWindSliderMaxRefNone__');
```
david-rios-uipath: does `watch('__apolloWindSliderMaxRefNone__')` do anything since this field doesn't exist? why not just do something like `field.maxRef ? watch(field.maxRef.fromField) : undefined`?

david-rios-uipath (approved):
a couple nits

packages/apollo-wind/src/components/forms/field-renderer.tsx
Holychung: Good call

packages/apollo-wind/src/components/forms/field-renderer.tsx
Holychung: Switched to [resolvedMax, formField.value, formField.onChange].

david-rios-uipath (approved):

PR #717, opened 2026-05-18
fix(apollo-react): remove automcomplete and add unique stage name

StephenHanzlik:
## Summary

The stage title input on every `StageNode` shared the literal name `"Stage Title"` and had no `autoComplete` opt-out. As a result, browsers (and password managers) treated every stage title across the canvas as the same field, surfacing autofill suggestions and remembering previously-typed stage names — which leaked across stages and across cases.

## Root Cause

`StageTitleInput.tsx` rendered the underlying `<input>` with a hard-coded `name="Stage Title"` and no `autoComplete` attribute. Because every stage rendered an input with the identical `name`, browser heuristics grouped them as the same logical field, enabling autofill/history dropdowns on the canvas.

## Demo

Before
image

After
Screenshot 2026-05-18 at 10 31 33 AM

## Fix

- Added a required `stageId: string` prop to `StageTitleInput` and threaded the existing stage `id` through from `StageNodeHeader`.
- Replaced the hard-coded `name="Stage Title"` with `name={\`stage-title-\${stageId}\`}`, giving every stage input a unique field name derived from its stage id.
- Set `autoComplete="off"` on the input so browsers do not offer history-based suggestions even when a stage id happens to repeat across documents.

## Validation

- Added a new `StageTitleInput - input attributes` describe block in `StageNode.test.tsx` with three tests that enter edit mode and assert:
  - `autocomplete="off"` is present on the rendered input
  - `name` equals `stage-title-<stageId>` for the default stage id
  - rendering two different stage ids produces two different `name` values, confirming the per-stage uniqueness guarantee
- Manually verified in the canvas playground that browser autofill no longer offers prior stage-name suggestions and that password managers no longer attach to the field.

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/StageTitleInput.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageTitleInput.tsx
@@ -74,6 +74,7 @@ export const StageTitleInput = ({
             <input
               ref={inputRef}
               name={`stage-title-${stageId}`}
+              aria-label="Stage title"
               autoComplete="off"
               value={label}
               onInput={handleChange}
--- a/packages/apollo-react/src/canvas/components/StageNode/StageTitleInput.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageTitleInput.tsx
@@ -74,7 +74,7 @@ export const StageTitleInput = ({
             <input
               ref={inputRef}
               name={`stage-title-${stageId}`}
-              aria-label="Stage title"
+              aria-label={`Stage title: ${label}`}
               autoComplete="off"
               value={label}
               onInput={handleChange}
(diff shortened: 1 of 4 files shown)


kittyyueli (approved):

packages/apollo-react/src/canvas/components/StageNode/StageTitleInput.tsx
```diff
@@ -69,7 +71,9 @@ export const StageTitleInput = ({
           {isEditing ? (
             <input
               ref={inputRef}
-              name="Stage Title"
+              name={`stage-title-${stageId}`}
+              aria-label="Stage title"
```
kittyyueli:
```suggestion
              aria-label={`Stage title: ${label}`}
```

kittyyueli (approved):

kittyyueli (approved):

PR #721, opened 2026-05-19
chore(apollo-react): expose ./i18n subpath export

StephenHanzlik:
Follow up to: https://github.com/UiPath/apollo-ui/pull/702 so that we can use `<ApI18nProvider>` in PO.Frontend.

To be consumed by: https://github.com/UiPath/PO.Frontend/pull/5462

### DIFF
--- a/packages/apollo-react/src/i18n/ApI18nProvider.tsx
+++ b/packages/apollo-react/src/i18n/ApI18nProvider.tsx
@@ -24,8 +24,10 @@ export type SupportedLocale = (typeof SUPPORTED_LOCALES)[number];
 
 export interface ApI18nProviderProps {
   /**
-   * Component path relative to src/ where locales are stored.
-   * Example: 'material/components/ap-chat'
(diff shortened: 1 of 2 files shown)


packages/apollo-react/src/i18n/ApI18nProvider.tsx
```diff
@@ -40,22 +42,21 @@ export interface ApI18nProviderProps {
...
  *
- * function MyComponent() {
- *   return (
- *     <ApI18nProvider component="material/components/ap-chat">
- *       <ApChat chatServiceInstance={chatService} />
- *     </ApI18nProvider>
- *   );
- * }
+ * <ApI18nProvider component="canvas" locale="en">
+ *   <MaestroDesigner ... />
```
kittyyueli: Let's keep the example the same with the chat

StephenHanzlik: Done

kittyyueli (approved):

PR #722, opened 2026-05-20
feat(apollo-react): export Api18nProvider

ayush578:
Need to export i18n to get Api18nProvider
We are using translation from apollo-ui now which needs this provider otherwise it is showing error

before
error if provider is not added

Screenshot 2026-05-20 at 12 31 29 PM

after

https://github.com/user-attachments/assets/1cf85b53-27a9-4b43-b32f-d670b169427d

### DIFF
--- a/packages/apollo-react/src/index.ts
+++ b/packages/apollo-react/src/index.ts
@@ -1,5 +1,6 @@
 // Core
 export * from './core';
-
+// i18n utilities
+export * from './i18n';
 // Theme utilities
 export * from './material';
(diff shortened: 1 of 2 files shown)


packages/apollo-react/src/index.ts
```diff
@@ -1,5 +1,6 @@
 // Core
 export * from './core';
-
+// i18n utilities
+export * from './i18n';
```
CalinaCristian, line 4: let's not do this, only the other one

ayush578:
closing this as already done in other PR

PR #724, opened 2026-05-20
feat(apollo-react): add CanvasModeToolbar and CanvasZoomControls

dbacomputer:
## Summary

- **CanvasModeToolbar** — mode switcher (design / debug / evaluate) with undo/redo controls, debug run controls (step, continue, restart, stop), upgrade-node action badge, and an optional add-panel slot. All Flow-app-specific dependencies (Zustand store, internal hooks, telemetry, i18n) replaced with explicit props and inline string defaults.
- **CanvasZoomControls** — zoom in / zoom out / fit-to-screen / tidy-up buttons. Supports `vertical` (default) and `horizontal` orientations. `forwardRef` added for external DOM access (e.g. overlap detection). Removes `useTranslation`/`CANVAS_NAMESPACE`.
- Both components exported from `canvas/components/index.ts` with Storybook stories covering all key states.
- **`useElementsOverlap`** hook added and exported from `canvas/hooks` — ResizeObserver-based detection of whether two DOM elements overlap, intended for wiring `CanvasModeToolbar` ↔ `CanvasZoomControls` auto-orientation.
- Fixed `isTerminalDebugStatus` to use the correct `DebugStatus` values (`'success' | 'error'`); previous values (`'completed' | 'failed' | 'cancelled'`) would have left the canvas locked after debug completion and broken the Restart button.
- Removed dead `initializing` prop (was never referenced in the component body).

## Storybook stories

| Component | Stories |
|---|---|
| `CanvasModeToolbar` | Default, WithEvaluateToggle, DebugRunning, DebugPaused, WithUpgradeAction, ReadOnly |
| `CanvasZoomControls` | Vertical, Horizontal, WithOrganize |

## Test plan

- [ ] Open Storybook → Canvas / Components / CanvasModeToolbar — verify all 6 stories render and mode toggle / debug controls are interactive
- [ ] Open Storybook → Canvas / Components / CanvasZoomControls — verify all 3 stories render and zoom/fit buttons work against a live ReactFlow canvas
- [ ] TypeScript: `pnpm --filter @uipath/apollo-react tsc --noEmit` passes
- [ ] No new Emotion/MUI imports introduced — all styling via Tailwind + `@uipath/apollo-wind`
- [ ] `pnpm --filter @uipath/apollo-react run format:check` passes


packages/apollo-react/src/canvas/components/CanvasModeToolbar/CanvasModeToolbar.tsx
```diff
@@ -0,0 +1,491 @@
...
+                className={`relative ${TOOLBAR_ICON_BUTTON_CLASS}`}
+                onClick={onUndo}
+                disabled={!canUndo}
+              >
+                <Undo2 />
+                <CountBadge count={undoCount} />
+              </Button>
+            </TooltipTrigger>
+            <TooltipContent>Undo (⌘Z)</TooltipContent>
+          </Tooltip>
```
BenGSchulz: Seems like we could possibly pull out a "ToolbarButton" shared between here and the zoom controls that reduces some of the duplication/repetition.

packages/apollo-react/src/canvas/components/CanvasModeToolbar/CanvasModeToolbar.tsx
```diff
@@ -0,0 +1,491 @@
...
+                variant="ghost"
+                size="xs"
+                icon
+                className="relative"
+                onClick={upgradeNodeAction.onClick}
+              >
+                <CircleArrowUp />
+                <span className="absolute -top-1 -right-1 min-w-[16px] h-4 px-1 text-[10px] font-medium bg-destructive text-destructive-foreground rounded-full flex items-center justify-center">
+                  {upgradeNodeAction.count}
+                </span>
```
BenGSchulz: Should this use the CountBadge?

packages/apollo-react/src/canvas/components/CanvasModeToolbar/CanvasModeToolbar.tsx
```diff
@@ -0,0 +1,491 @@
...
+          <ToggleGroupItem
+            data-testid="canvas-mode-evaluate"
+            value="evaluate"
+            className="flex h-8 items-center gap-2 rounded-[10px] px-3 text-sm font-medium text-foreground-muted transition hover:bg-transparent hover:text-foreground data-[state=on]:bg-primary/10 data-[state=on]:text-foreground data-[state=on]:shadow-xs"
+          >
+            <ListChecks
+              strokeWidth={1.75}
+              className={mode === 'evaluate' ? 'text-foreground-accent' : ''}
+            />
+            Evaluate
```
BenGSchulz: I believe we are adding i18n to this repo. Might want to make sure these new components are using it if available.

dbacomputer:
Good feedback @BenGSchulz, all three addressed:

ToolbarButton — extracted a shared ToolbarButton component (canvas/components/ToolbarButton.tsx) that wraps the Tooltip → Button pattern. Both CanvasModeToolbar and CanvasZoomControls now import it. Each file keeps its own icon-size class constant so the sizing stays context-specific.

CountBadge — added a variant prop ('default' | 'destructive'). The upgrade node badge now uses <CountBadge variant="destructive" /> instead of the inline span — same visual output, consistent with how undo/redo badges are handled.

i18n — wired useSafeLingui into both components and wrapped all user-visible strings. Used the same pattern as Toolbox and StageNode. The hook falls back to the English message string when no provider is mounted so there's no risk of regression for consumers who haven't added ApI18nProvider yet.

packages/apollo-react/src/canvas/components/CanvasModeToolbar/CanvasModeToolbar.tsx
```diff
@@ -0,0 +1,422 @@
...
+            onClick={onAddStickyNote}
+            disabled={!canModifyCanvas}
+            className={TOOLBAR_ICON_BUTTON_CLASS}
+          >
+            <StickyNote />
+          </ToolbarButton>
+        </>
+      )}
+
+      {canModifyCanvas && upgradeNodeAction && upgradeCount > 0 && (
```
BenGSchulz: Thanks for addressing my earlier comments. Sorry I missed raising this one earlier. I think we are moving away from the upgrade UX in this toolbar.

But want to confirm with @soomin-uipath first since he has more context on the upgrade UX.

BenGSchulz: Yeah, confirmed with Soomin that it has been removed via https://github.com/UiPath/flow-workbench/pull/1697. So we can remove this I think.

It actually raises another point to consider though... These components lock the UI library into flow-implementation-specific behavior (i.e. Build vs Evaluate modes and having the add sticky/node actions, and debug status handling). I wonder if it might make sense to have these components act more as a standard formatter / shell for any XYZ actions/modes that a consumer might want to add. So rather than accepting explicit props like `showEvaluateToggle` or `onAddNode`, we just render children that the consumer provides like the mode toggle (which can be its own helper component) or ToolbarButtons. The consumer is responsible for figuring out how they want to expose and handle modes / actions while we handle the presentation.

dbacomputer: @BenGSchulz two updates completed here to address your feedback. Thank you

1. Removed: upgradeNodeAction prop, the button block, CircleArrowUp import, and the WithUpgradeAction story are all gone.

2. Children-based shell: CanvasModeToolbar is now ~65 lines — a styled Row container that accepts children and an optional className. All flow-specific props, internal state, and handlers have been removed. CountBadge and TOOLBAR_ICON_BUTTON_CLASS are now exported building blocks. The 4 remaining stories (Default, DebugRunning, DebugPaused, ReadOnly) show how consumers compose their own toolbar using ToolbarButton + Separator + CountBadge. The bot's unused Button import is also gone since the component was rewritten.

packages/apollo-react/src/canvas/components/CanvasModeToolbar/CanvasModeToolbar.stories.tsx
```diff
@@ -0,0 +1,225 @@
...
+  Square,
+  StepForward,
+  StickyNote,
+  Undo2,
+} from 'lucide-react';
+import { ToolbarButton } from '../ToolbarButton';
+import { CanvasModeToolbar, CountBadge, TOOLBAR_ICON_BUTTON_CLASS } from './CanvasModeToolbar';
+
+const meta: Meta<typeof CanvasModeToolbar> = {
+  title: 'Components/CanvasModeToolbar',
```
BenGSchulz: Should we update this (and the zoom toolbar) to follow new folder structure?

BenGSchulz (approved):

packages/apollo-react/src/canvas/components/CanvasModeToolbar/CanvasModeToolbar.stories.tsx
dbacomputer: Updated: Both story titles are now Components/Controls/CanvasModeToolbar and Components/Controls/CanvasZoomControls, matching the new IA (same group as NodeToolbar, Breadcrumb, and Toolbox on main).

BenGSchulz (approved):

PR #730, opened 2026-05-21
feat(apollo-vertex): dashboard template foundation — registry, data layer, shell

petervachon:
## Summary

Sets up the scaffolding for the modular dashboard template. This PR doesn't render any visible UI yet — it establishes the routing shell (TanStack Router with a memory history so the preview is self-contained), the data model that every dashboard component will read from, and the two preview routes at `/preview/dashboard` and `/preview/dashboard-minimal`. The `DashboardContent` is a stub that renders nothing; the full implementation comes in PR 8. Reviewers can confirm the routes exist and the page renders without errors.

## This is a stacked PR

PRs 2–10 are open as drafts and build on this one. Review and merge in order.

## Test plan
- [ ] Visit `/preview/dashboard` — page renders without errors, no content yet
- [ ] Visit `/preview/dashboard-minimal` — same
- [ ] No TypeScript errors in the dashboard template files


apps/apollo-vertex/registry/card/card.tsx
```diff
@@ -10,7 +10,7 @@ export const GLASS_CLASSES = [
   "dark:shadow-[0_2px_24px_2px_rgba(0,0,0,0.12),inset_0_1px_0_0_color-mix(in_srgb,var(--sidebar)_5%,transparent)]",
 ] as const;
 
-const cardVariants = cva("flex flex-col text-card-foreground", {
+const cardVariants = cva("flex flex-col gap-6 py-6 text-card-foreground", {
```
0xr3ngar: `apps/apollo-vertex/registry/metric-card/metric-card.tsx:36-42` already has inner p-6, so metric cards now get extra vertical padding

this looks unrelated to the dashboard foundation that is in this PR and is a likely visual regression

0xr3ngar (changes requested):

apps/apollo-vertex/registry/card/card.tsx
petervachon: Good catch — scoped to a `dashboard` variant in b41cf9d6 so other Card consumers aren't affected. The base classes are back to the pre-PR state; dashboard cards can opt in with `variant="dashboard"` when they need the `gap-6 py-6` rhythm.

0xr3ngar (approved):

PR #731, opened 2026-05-21
feat(apollo-vertex): dashboard config layer — glow-config, DashboardCards, AutopilotInsight

petervachon:
## Summary

Introduces the design token layer (`glow-config.ts`) that controls card backgrounds, gradients, opacity, layout spacing, and the glow animation — essentially the "knobs" the rest of the dashboard reads from. Also adds two standalone display components with no dependencies on the rest of the dashboard: `DashboardCards` (KPI tiles, invoice tables, activity feeds, pipeline and compliance bars) and `AutopilotInsight` (the AI insight panel that slides in from the right). No wiring yet — these components can be reviewed in isolation as pure presentational pieces.

## Stacked on
#730

## Test plan
- [ ] No TypeScript errors in the new files
- [ ] `glow-config.ts` exports `GlowConfig`, `CardConfig`, `LayoutConfig` and related types and defaults


apps/apollo-vertex/templates/dashboard/DashboardRoutes.tsx
frankkluijtmans, line 9: The way this is set up seems like it's a self contained dashboard app rather than a template. Can you explain a bit what the idea is here? 🤔  I don't expect a router inside of a template.

apps/apollo-vertex/templates/dashboard/AutopilotInsight.tsx
```diff
@@ -0,0 +1,76 @@
...
+export function AutopilotInsight({
+  onClose,
+  sourceCardTitle,
+}: AutopilotInsightProps) {
+  return (
+    <Card
+      variant="glass"
+      className="!bg-white/90 dark:!bg-card h-full overflow-hidden relative"
+    >
+      {/* Close button */}
```
frankkluijtmans, line 22: This comment isn't really needed. Also; don't we have a `Button` component that also takes care of accessible labels and such? I'm missing that right now.

apps/apollo-vertex/templates/dashboard/AutopilotInsight.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <img
+            src="/Autopilot_dark.svg"
+            alt="Autopilot"
+            className="size-5 block dark:hidden"
+          />
+          <img
+            src="/Autopilot_light.svg"
+            alt="Autopilot"
+            className="size-5 hidden dark:block"
+          />
```
frankkluijtmans, line 56: If this is gonna work, we have to make sure these images get pulled from the registry and we provide the correct path. Maybe it's better to create icon components for them with inline svg.

apps/apollo-vertex/app/preview/dashboard/page.tsx
frankkluijtmans, line 3: Why did we add this? Shouldn't we just add a new page.mdx like we do for the other templates?

apps/apollo-vertex/templates/dashboard/DashboardShellWrapper.tsx
frankkluijtmans, line 53: We shouldn't have to provide this. Apps using Apollo Vertex should already have the shell, router, nav and everything set up. This is not something that should be handled here.

frankkluijtmans (changes requested):

apps/apollo-vertex/templates/dashboard/glow-config.ts
```diff
@@ -0,0 +1,294 @@
+export interface GlowConfig {
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed in eebd7be2 — added `import type { CSSProperties } from "react"` and use it directly instead of the global `React` namespace.

apps/apollo-vertex/templates/dashboard/DashboardLoading.tsx
```diff
@@ -0,0 +1,135 @@
+"use client";
+
+import { useEffect, useState } from "react";
+
+type Phase = "logo" | "skeleton" | "done";
+
+interface DashboardLoadingProps {
+  children: React.ReactNode;
+  triggerReplay?: number;
+}
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed in eebd7be2 — `children` is now typed as `ReactNode` via an explicit `import { type ReactNode } from "react"`.

apps/apollo-vertex/templates/dashboard/DashboardLoading.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    if (phase === "skeleton") {
+      const timer = setTimeout(() => {
+        setExiting(true);
+        setTimeout(() => {
+          setPhase("done");
+        }, 500);
+      }, 1000);
+      return () => clearTimeout(timer);
+    }
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed in eebd7be2 — the nested exit timer is now tracked in `exitTimer` and cleared alongside the enter timer in the effect cleanup, so no state updates fire after unmount.

frankkluijtmans (approved):

alincadariu (approved):

PR #747, opened 2026-05-25
feat(apollo-react): add getTaskContextMenuItems in stageProps [MST-7548]

ayush578:
add general getTaskContextMenuItems for adding more menu items

https://github.com/user-attachments/assets/2fc26738-44be-43ff-a519-11455864a257

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNode.stories.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNode.stories.tsx
@@ -1218,6 +1218,13 @@ const initialTasksForAddReplace: StageTaskItem[][] = [
   ],
   [{ id: 'task-2', label: 'Document Review', icon: <DocumentIcon /> }],
   [{ id: 'task-3', label: 'Process Validation', icon: <ProcessIcon /> }],
+  [
+    {
+      id: 'task-with-custom-action',
+      label: 'Task with Custom Menu Action',
+      icon: <ProcessIcon />,
+    },
+  ],
 ];
 
 const availableTaskOptions: ListItem[] = [
@@ -1504,6 +1511,16 @@ const AddAndReplaceTasksStory = ({ locale }: { locale?: SupportedLocale }) => {
                 onTaskGroupModification: handleTaskGroupModification,
                 onTaskReorder: handleTaskReorder,
                 onTaskClick: handleTaskClick,
+                getTaskContextMenuItems: ({ task }: { task: StageTaskItem }) =>
+                  task.id === 'task-with-custom-action'
+                    ? [
+                        {
+                          id: 'go-to-definition',
+                          label: 'Go to definition',
+                          onClick: () => alert(`Go to definition clicked for task: ${task.id}`),
+                        },
+                      ]
+                    : undefined,
               },
             }
           : node
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNode.test.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNode.test.tsx
@@ -1071,3 +1071,77 @@ describe('StageTitleInput - input attributes', () => {
     expect(input).toHaveAttribute('name', 'stage-title-stage-b');
   });
 });
+
+describe('StageNode - getTaskContextMenuItems', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  const renderWithCallback = (
+    getTaskContextMenuItems?: StageNodeProps['getTaskContextMenuItems']
+  ) =>
+    renderStageNode({
+      stageDetails: {
+        label: 'Test Stage',
+        tasks: [[{ id: 'task-1', label: 'Task 1' }]],
+      },
+      getTaskContextMenuItems,
+    });
+
+  it('appends items returned by getTaskContextMenuItems to the task menu', async () => {
+    const user = userEvent.setup();
+    const onCustom = vi.fn();
+    renderWithCallback(() => [{ id: 'custom-action', label: 'Custom Action', onClick: onCustom }]);
+
+    await user.click(screen.getByTestId('task-menu-button-task-1'));
+
+    expect(screen.getByTestId('menu-item-task-1-custom-action')).toBeInTheDocument();
+    expect(screen.getByTestId('menu-item-task-1-custom-action')).toHaveTextContent('Custom Action');
+  });
+
+  it('passes the task and group context to the callback', async () => {
+    const user = userEvent.setup();
+    const getTaskContextMenuItems = vi.fn().mockReturnValue([]);
+    renderStageNode({
+      stageDetails: {
+        label: 'Test Stage',
+        tasks: [[{ id: 'task-a', label: 'Task A' }]],
+      },
+      onReplaceTaskFromToolbox: vi.fn(), // ensure the menu renders
+      getTaskContextMenuItems,
+    });
+
+    await user.click(screen.getByTestId('task-menu-button-task-a'));
+
+    expect(getTaskContextMenuItems).toHaveBeenCalledWith({
+      task: expect.objectContaining({ id: 'task-a', label: 'Task A' }),
+      taskGroupType: 'sequential',
+      isParallel: false,
+    });
+  });
+
+  it('does not render the task menu trigger when the callback returns an empty array and no other actions are provided', () => {
+    renderWithCallback(() => []);
+
+    expect(screen.queryByTestId('task-menu-button-task-1')).not.toBeInTheDocument();
+  });
+
+  it('does not render the task menu trigger when the callback returns undefined and no other actions are provided', () => {
+    renderWithCallback(() => undefined);
+
+    expect(screen.queryByTestId('task-menu-button-task-1')).not.toBeInTheDocument();
+  });
+
+  it('invokes the item onClick when the menu item is clicked', async () => {
+    const user = userEvent.setup();
+    const onCustom = vi.fn();
+    renderWithCallback(() => [
+      { id: 'go-to-definition', label: 'Go to definition', onClick: onCustom },
+    ]);
+
+    await user.click(screen.getByTestId('task-menu-button-task-1'));
+    await user.click(screen.getByTestId('menu-item-task-1-go-to-definition'));
+
+    expect(onCustom).toHaveBeenCalledTimes(1);
+  });
+});
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNode.types.ts
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNode.types.ts
@@ -30,6 +30,12 @@ export interface StageTaskItem {
   hasEntryCondition?: boolean;
 }
 
+export interface StageTaskContextMenuArgs {
+  task: StageTaskItem;
+  taskGroupType: 'sequential' | 'event-driven' | 'adhoc';
+  isParallel: boolean;
+}
+
 export enum StageHeaderChipType {
   Entry = 'entry',
   Exit = 'exit',
@@ -86,6 +92,7 @@ export interface StageNodeBaseProps {
   onTaskReorder?: (reorderedTasks: StageTaskItem[][]) => void;
   onReplaceTaskFromToolbox?: (newTask: ListItem, groupIndex: number, taskIndex: number) => void;
   onTaskPlay?: (taskId: string) => Promise<void>;
+  getTaskContextMenuItems?: (args: StageTaskContextMenuArgs) => NodeMenuItem[] | undefined;
   hideParallelOptions?: boolean;
   loadingTaskIds?: ReadonlySet<string>;
 }
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNodeAdhocTaskGroups.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNodeAdhocTaskGroups.tsx
@@ -6,7 +6,7 @@ import {
   StageAdditionalTasksSection,
   StageTaskList,
 } from './StageNode.styles';
-import type { StageNodeProps, StageTaskGroup } from './StageNode.types';
+import type { StageNodeProps, StageTaskGroup, StageTaskItem } from './StageNode.types';
 import { useStageNodeLabels } from './useStageNodeLabels';
 
 const StageNodeAdhocTaskGroupsInner = ({
@@ -37,28 +37,34 @@ const StageNodeAdhocTaskGroupsInner = ({
     onReplaceTaskFromToolbox,
     onTaskPlay,
     loadingTaskIds,
+    getTaskContextMenuItems,
   } = props;
+  const hasBuiltInTaskActions = !!(onReplaceTaskFromToolbox || onTaskGroupModification);
   const labels = useStageNodeLabels();
 
   /** Lazily builds context menu items for a task. Called only when the menu opens,
    * avoiding object allocation on every render for every task. */
   const getAdhocContextMenuItems = useCallback(
-    (taskId: string): NodeMenuItem[] => {
+    (task: StageTaskItem): NodeMenuItem[] => {
       const items: NodeMenuItem[] = [];
 
-      const replaceTaskMenuItem = generateReplaceTaskMenuItemForTask(taskId, false);
+      const replaceTaskMenuItem = generateReplaceTaskMenuItemForTask(task.id, false);
       if (replaceTaskMenuItem) {
         items.push(replaceTaskMenuItem);
       }
 
-      const deleteTaskMenuItem = generateDeleteTaskMenuItemForTask(taskId);
+      const additionalMenuItems =
+        getTaskContextMenuItems?.({ task, taskGroupType: 'adhoc', isParallel: false }) ?? [];
+      items.push(...additionalMenuItems);
+
+      const deleteTaskMenuItem = generateDeleteTaskMenuItemForTask(task.id);
       if (deleteTaskMenuItem) {
         items.push(deleteTaskMenuItem);
       }
 
       return items;
     },
-    [generateReplaceTaskMenuItemForTask, generateDeleteTaskMenuItemForTask]
+    [generateReplaceTaskMenuItemForTask, getTaskContextMenuItems, generateDeleteTaskMenuItemForTask]
   );
 
   if (adhocTasks.length === 0) {
@@ -72,6 +78,15 @@ const StageNodeAdhocTaskGroupsInner = ({
       <StageTaskList>
         {adhocTasks.map(({ task }) => {
           const taskExecution = execution?.taskStatus?.[task.id];
+          const customItems =
+            !isReadOnly && !hasBuiltInTaskActions
+              ? (getTaskContextMenuItems?.({
+                  task,
+                  taskGroupType: 'adhoc',
+                  isParallel: false,
+                }) ?? [])
+              : [];
+          const hasMenu = !isReadOnly && (hasBuiltInTaskActions || customItems.length > 0);
           return (
             <AdhocTaskItem
               key={task.id}
@@ -81,10 +96,9 @@ const StageNodeAdhocTaskGroupsInner = ({
               onTaskClick={handleTaskClick}
               onTaskPlay={onTaskPlay}
               isTaskLoading={loadingTaskIds?.has(task.id)}
-              {...((onTaskGroupModification || onReplaceTaskFromToolbox) &&
-                !isReadOnly && {
-                  getContextMenuItems: () => getAdhocContextMenuItems(task.id),
-                })}
+              {...(hasMenu && {
+                getContextMenuItems: () => getAdhocContextMenuItems(task),
+              })}
             />
           );
         })}
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNodeAllTaskGroups.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNodeAllTaskGroups.tsx
@@ -67,8 +67,6 @@ const StageNodeAllTaskGroupsInner = ({
     [onTaskReorder, eventDrivenTaskGroups, adhocTaskGroups]
   );
 
-  const hasContextMenu = !!(onReplaceTaskFromToolbox || onTaskGroupModification);
-
   const handleTaskClick = useCallback(
     (e: React.MouseEvent, taskElementId: string) => {
       e.stopPropagation();
@@ -177,7 +175,6 @@ const StageNodeAllTaskGroupsInner = ({
             isReadOnly={isReadOnly}
             selectedTaskId={selectedTaskId}
             taskWidthStyle={taskWidthStyle}
-            hasContextMenu={hasContextMenu}
             handleTaskClick={handleTaskClick}
             handleReorderSequentialTasks={handleReorderSequentialTasks}
             allTasks={allTasks}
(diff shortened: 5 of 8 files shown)


SreedharAvvari (changes requested):
Recommended change: make this a generic task-menu extension point rather than baking the product-specific action into `StageNode`.

The intent makes sense: PO.Frontend needs a way to show something like "Go to definition" for only some task rows. But `isInSolutionProject` and `onGoToDefinition` make Apollo aware of consumer/product semantics, and we may end up adding one Apollo prop per consumer-specific menu action over time.

A more reusable API would be something like:

```ts
type StageTaskContextMenuArgs = {
  task: StageTaskItem;
  groupIndex: number;
  taskIndex: number;
  isParallel: boolean;
  taskGroupType: "sequential" | "event-driven" | "adhoc";
};

getTaskContextMenuItems?: (args: StageTaskContextMenuArgs) => NodeMenuItem[] | undefined;
```

Then Apollo can merge those consumer-provided items into the existing task menu, while still owning ordering, separators, and rendering. For example, the consumer could return:

```ts
getTaskContextMenuItems={({ task }) => {
  if (!isTaskInSolutionProject(task.id)) {
    return [];
  }

  return [
    {
      id: "go-to-definition",
      label: "Go to definition",
      onClick: () => openDefinition(task.id),
    },
  ];
}}
```

That keeps Apollo responsible for the generic StageNode task-menu surface, while PO.Frontend owns what "Go to definition" means and when it should appear.

uipath-utkarsh (commented):
## Review summary

The functional change is sound and the new flag/callback are wired through all three task-group types. Three structural issues dominate the review:

1. **Eager per-render context-menu evaluation** across `Sequential / Adhoc / EventDriven` TaskGroups breaks the explicit lazy contract called out in their JSDocs *and* defeats `React.memo` on every child task component. This is the most impactful behavior change in the PR.
2. **`allTasks.flat().find(...)`** turns the new menu-generation pipeline O(N²) per render. A `useMemo`-built `Map<taskId, task>` solves it and would also let us replace the existing nested `for…of entries()` walks (which incidentally hide a latent inner-`break` correctness bug for duplicate task ids).
3. **Adding a required `goToDefinition` field to the exported `StageNodeLabels` interface** is a public-API breaking change. Either mark it optional with a fallback default in the hook, or bump the package as a major release.

Also worth addressing before merge:

- `isInSolutionProject` exposes UiPath product semantics through the open-source surface (cleaner: `enableGoToDefinition`, or drop the flag and let presence of `onGoToDefinition` be the gate).
- The menu-trigger visibility contract silently changed from “callback present” to “items non-empty” — please document or add a regression test for the transient-task case.
- The new tests cover only the sequential path; adhoc and event-driven paths (also touched here) are not tested, and nothing asserts menu-item ordering or the implicit divider-placement fix in `StageNodeSequentialTaskGroups.tsx`.

Inline comments below.

packages/apollo-react/src/canvas/components/StageNode/useStageNodeLabels.ts
```diff
@@ -8,6 +8,7 @@ export interface StageNodeLabels {
   addTask: string;
   replaceTask: string;
   deleteTask: string;
+  goToDefinition: string;
```
uipath-utkarsh: **Breaking change to a public interface.**

```ts
export interface StageNodeLabels {
  ...
  goToDefinition: string; // required
}
```

`StageNodeLabels` is part of the package's exported surface. Adding a *required* field is a TypeScript breaking change for any downstream consumer that satisfies this interface explicitly (custom translation provider, typed mock, label override wrapper) — they will get `Property 'goToDefinition' is missing in type ...`.

Two viable options:

1. Mark it `goToDefinition?: string` and provide a fallback inside `useStageNodeLabels` so the consumer's existing labels object still type-checks.
2. Treat this as a `major` bump (rather than `feat`/`minor`) in the changeset.

Given that no other field has ever needed to be added to this interface since it shipped, option (1) is probably the safer default going forward.

packages/apollo-react/src/canvas/components/StageNode/StageNode.types.ts
```diff
@@ -28,6 +28,7 @@ export interface StageTaskItem {
   isPlaceholder?: boolean;
   taskGroupType?: 'sequential' | 'event-driven' | 'adhoc';
   hasEntryCondition?: boolean;
+  isInSolutionProject?: boolean;
```
uipath-utkarsh: **`isInSolutionProject` leaks product semantics into the public design-system API, and its coupling to `onGoToDefinition` is implicit.**

```ts
isInSolutionProject?: boolean;
```

Two concerns:

1. **Naming.** The other flags on `StageTaskItem` (`isPlaceholder`, `hasEntryCondition`, `taskGroupType`) are structural/visual. *“Solution Project”* is a UiPath Studio concept — an external consumer of this open-source design system has no way to interpret it from the type alone. Either rename to a behavior-describing flag (e.g. `enableGoToDefinition?: boolean`) or drop the flag entirely and let the consumer decide per task by returning early from `onGoToDefinition` for unsupported tasks.
2. **Implicit coupling.** The menu item only appears when *both* `isInSolutionProject` is true **and** `onGoToDefinition` is set on `StageNodeBaseProps`. Neither field's TSDoc points at the other, so a consumer who sets just one will see nothing in the UI with no compile-time hint. Please add cross-referencing TSDoc on both, e.g.:

```ts
/**
 * When true, the task's context menu shows a "Go to definition" entry,
 * provided `StageNodeBaseProps.onGoToDefinition` is also supplied.
 */
isInSolutionProject?: boolean;
```

packages/apollo-react/src/canvas/components/StageNode/StageNodeSequentialTaskGroups.tsx
```diff
@@ -211,6 +217,11 @@ export const StageNodeSequentialTaskGroups = ({
                     )}
                     {taskGroup.map((task, taskIndex) => {
                       const taskExecution = execution?.taskStatus?.[task.id];
+                      const contextMenuItems = buildContextMenuItems(
```
uipath-utkarsh: **Lazy-build contract broken + `React.memo` on `DraggableTask` defeated.**

```ts
const contextMenuItems = buildContextMenuItems(groupIndex, taskIndex, task.id);
...
{...(contextMenuItems.length > 0 &&
  !isReadOnly && {
    getContextMenuItems: () => contextMenuItems,
  })}
```

Two regressions ride together here:

1. The pre-existing JSDoc on `buildContextMenuItems` (line 112) explicitly says *“Called only when the menu opens, avoiding object allocation on every render for every task.”* That contract is now violated — `buildContextMenuItems` runs for every sequential task on every render. Internally each call walks `allTasks.entries()` twice and the new `generateGoToDefinitionMenuItemForTask` does `allTasks.flat().find(...)` (see other comment). For a stage with N tasks this is O(N²) work per render even when no menu is ever opened.
2. `DraggableTask` is wrapped in `memo()`. Pre-PR it received the stable `useCallback` reference `buildContextMenuItems` — memo skipped re-renders. Post-PR it receives `() => contextMenuItems`, a fresh arrow every render → memo always invalidates → every draggable task in the stage re-renders whenever any unrelated parent state changes (hover, zoom, drag projection, sibling status update).

**Suggested shape:** keep `buildContextMenuItems` as a stable `useCallback` (passed by reference) and gate trigger visibility on a cheap `taskHasMenuItems(taskId): boolean` predicate. Don’t precompute the full array per render just to read its `.length`.

If you keep the eager shape, please at least update the JSDoc on line 112 — it’s now misleading.

packages/apollo-react/src/canvas/components/StageNode/StageNodeAdhocTaskGroups.tsx
```diff
@@ -72,6 +77,7 @@ const StageNodeAdhocTaskGroupsInner = ({
       <StageTaskList>
         {adhocTasks.map(({ task }) => {
           const taskExecution = execution?.taskStatus?.[task.id];
+          const contextMenuItems = getAdhocContextMenuItems(task.id);
```
uipath-utkarsh: Same regression as in `StageNodeSequentialTaskGroups.tsx` (see the longer comment there):

1. `getAdhocContextMenuItems(task.id)` is now invoked eagerly during the `.map()` render of every adhoc task — the JSDoc on line 39–40 still claims it’s *“Called only when the menu opens.”*
2. `() => contextMenuItems` is a brand-new arrow each render, breaking `memo()` on `AdhocTaskItem` and forcing re-renders even when the task’s own props are unchanged.

Fix in tandem with the sequential file: keep the producer lazy + use a cheap `hasMenuItems(taskId)` predicate to gate visibility. If you keep the eager shape, please update the JSDoc to reflect the new semantics.

packages/apollo-react/src/canvas/components/StageNode/StageNodeEventDrivenTaskGroups.tsx
```diff
@@ -66,6 +77,7 @@ const StageNodeEventDrivenTaskGroupsInner = ({
       <StageTaskList>
         {eventDrivenTasks.map(({ task }) => {
           const taskExecution = execution?.taskStatus?.[task.id];
+          const contextMenuItems = getEventDrivenContextMenuItems(task.id);
```
uipath-utkarsh: Same regression as in `StageNodeSequentialTaskGroups.tsx` and `StageNodeAdhocTaskGroups.tsx` (see longer comments on those files):

1. `getEventDrivenContextMenuItems(task.id)` is now invoked eagerly during render for every event-driven task, contradicting the *“Called only when the menu opens”* JSDoc on line 39–40.
2. `() => contextMenuItems` is a fresh arrow each render, defeating `memo()` on `EventDrivenTaskItem`.

Apply the same fix (cheap `hasMenuItems(taskId)` predicate + keep the producer lazy).

packages/apollo-react/src/canvas/components/StageNode/StageNodeAdhocTaskGroups.tsx
```diff
@@ -81,9 +87,9 @@ const StageNodeAdhocTaskGroupsInner = ({
               onTaskClick={handleTaskClick}
               onTaskPlay={onTaskPlay}
               isTaskLoading={loadingTaskIds?.has(task.id)}
-              {...((onTaskGroupModification || onReplaceTaskFromToolbox) &&
+              {...(contextMenuItems.length > 0 &&
```
uipath-utkarsh: **Subtle menu-trigger visibility contract change.**

Pre-PR the spread gate was:

```ts
(onTaskGroupModification || onReplaceTaskFromToolbox) && !isReadOnly
```

i.e. *“if a callback is wired up, render the trigger.”* Post-PR it is:

```ts
contextMenuItems.length > 0 && !isReadOnly
```

i.e. *“only if at least one item can be built right now.”*

Concrete failure scenario: a consumer wires `onTaskGroupModification` and renders a task whose `id` is momentarily not in `stageDetails.tasks` (placeholder, optimistic update lag, transient state during reorder/replace). Every generator returns `undefined`, `contextMenuItems.length === 0`, and the kebab silently disappears — breaking consumer E2E selectors that key off the trigger button being stable.

Same applies in `StageNodeEventDrivenTaskGroups.tsx` and `StageNodeSequentialTaskGroups.tsx`. If item-driven visibility is the intended behavior, please document the contract change in a changeset note and add a regression test for the “task transiently absent from `allTasks`” case.

packages/apollo-react/src/canvas/components/StageNode/StageNodeAllTaskGroups.tsx
```diff
@@ -147,6 +146,20 @@ const StageNodeAllTaskGroupsInner = ({
     [allTasks, onTaskGroupModification, labels.deleteTask]
   );
 
+  const generateGoToDefinitionMenuItemForTask = useCallback(
+    (taskId: string) => {
+      if (!onGoToDefinition) {
+        return undefined;
+      }
+      const task = allTasks.flat().find((t) => t.id === taskId);
```
uipath-utkarsh: **Quadratic per-render work + redundant lookup + inherits a latent inner-`break` bug.**

```ts
const task = allTasks.flat().find((t) => t.id === taskId);
```

Three problems folded into this line:

1. `allTasks.flat()` allocates a fresh flattened array on every call. Combined with the eager call sites in the three child task-group files, this becomes O(N²) work + N `flat()` allocations per stage render.
2. **Redundant lookup** — every caller (Sequential / Adhoc / EventDriven map callback) already holds the `task` object directly. The generator re-finds it from `allTasks`.
3. The sibling generators (`generateReplaceTaskMenuItemForTask`, `generateDeleteTaskMenuItemForTask`) use the older nested `for…of entries()` pattern; their inner `break` only exits the inner loop — if `allTasks` ever holds a duplicate `task.id` across groups (mid-drag clone, undo restoring stale state), the indices get silently overwritten by the *last* occurrence.

**Suggested fix — single change resolves all three:**

```ts
const tasksById = useMemo(() => {
  const m = new Map<string, { task: StageTaskItem; groupIndex: number; taskIndex: number }>();
  allTasks.forEach((group, gi) =>
    group.forEach((t, ti) => m.set(t.id, { task: t, groupIndex: gi, taskIndex: ti }))
  );
  return m;
}, [allTasks]);
```

Then each generator becomes an O(1) lookup, the duplicate-id ambiguity becomes “first occurrence wins” deterministically, and `allTasks` can drop out of the inner `useCallback` dep arrays in favor of `tasksById`.

Alternatively, since the call site already has the `task`, pass it through directly — `generateGoToDefinitionMenuItemForTask(taskId, isInSolutionProject)` — and skip the lookup entirely for this generator (mirrors the existing `generateReplaceTaskMenuItemForTask(taskId, isParallel)` shape).

packages/apollo-react/src/canvas/components/StageNode/StageNode.test.tsx
```diff
@@ -1071,3 +1071,56 @@ describe('StageTitleInput - input attributes', () => {
     expect(input).toHaveAttribute('name', 'stage-title-stage-b');
   });
 });
+
+describe('StageNode - Go to Definition', () => {
```
uipath-utkarsh: **Coverage gaps in the new tests.**

The new `describe` block only exercises sequential tasks via `tasks: [[{ id: 'task-1', ... }]]`. A few cases that the PR actually changes are not covered:

1. **Adhoc and event-driven paths** — identical-shape logic lives in `StageNodeAdhocTaskGroups.tsx` and `StageNodeEventDrivenTaskGroups.tsx`, but neither path is tested. A future change that, say, forgets to thread `generateGoToDefinitionMenuItemForTask` into the `useCallback` deps of `getAdhocContextMenuItems` / `getEventDrivenContextMenuItems` (causing a stale-closure bug for `onGoToDefinition`) would land green.
2. **Menu-item ordering** — `[Replace, Go to definition, Delete]` vs sequential `[Replace, Go to definition, ─divider─, …regroup options]`. Today only presence is asserted; assert with `screen.getAllByRole('menuitem').map(el => el.textContent)` so a swap is caught.
3. **Trailing-divider fix in sequential** — `StageNodeSequentialTaskGroups.tsx:159` implicitly fixes a pre-existing trailing-divider bug (`onReplaceTaskFromToolbox` set but `onTaskGroupModification` absent → no dangling divider). Without a test it can silently come back.
4. The test `'does not render the task menu trigger when task.isInSolutionProject is false and no other menu actions are provided'` depends on `renderStageNode` *not* injecting default values for `onTaskGroupModification` / `onReplaceTaskFromToolbox`. Worth asserting that explicitly, or pinning the helper, so a future defaults change doesn't silently invalidate this case.

packages/apollo-react/src/canvas/components/StageNode/StageNode.stories.tsx
```diff
@@ -1504,6 +1512,9 @@ const AddAndReplaceTasksStory = ({ locale }: { locale?: SupportedLocale }) => {
                 onTaskGroupModification: handleTaskGroupModification,
                 onTaskReorder: handleTaskReorder,
                 onTaskClick: handleTaskClick,
+                onGoToDefinition: (taskId: string) => {
```
uipath-utkarsh: Prefer `action('onGoToDefinition')` from `@storybook/addon-actions` (already used elsewhere in this file's stories) over `alert(...)`. `alert` blocks the main thread and forces dismissal between every click, which is awkward when validating across multiple tasks — and Storybook's actions panel is the normal place to inspect callback invocations.

packages/apollo-react/src/canvas/components/StageNode/StageNodeSequentialTaskGroups.tsx
```diff
@@ -226,9 +239,12 @@ export const StageNodeSequentialTaskGroups = ({
                           }
                           isDragDisabled={!onTaskReorder || isReadOnly}
                           isTaskLoading={loadingTaskIds?.has(task.id)}
-                          {...(hasContextMenu &&
+                          {...(hasMenu &&
                             !isReadOnly && {
-                              getContextMenuItems: buildContextMenuItems,
+                              getContextMenuItems: () => [
```
uipath-utkarsh: **Two regressions vs `main` ride together here, both in this file and mirrored in `StageNodeAdhocTaskGroups.tsx` (lines 77-92) and `StageNodeEventDrivenTaskGroups.tsx` (lines 76-96).**

```ts
const additionalMenuItems =
  getTaskContextMenuItems?.({ task, taskGroupType: 'sequential', isParallel }) ?? [];
const hasMenu = hasBuiltInTaskActions || additionalMenuItems.length > 0;
...
getContextMenuItems: () => [
  ...buildContextMenuItems(groupIndex, taskIndex, task.id),
  ...additionalMenuItems,
],
```

1. **Consumer callback runs eagerly per task per render.** `getTaskContextMenuItems` is invoked on every render of every task just to compute `additionalMenuItems.length` for the visibility predicate. A consumer who naturally writes `getTaskContextMenuItems={({task}) => task.id === 'x' ? [...] : undefined}` pays N invocations per parent render even when no menu is opened. Non-trivial consumer callbacks (state/selector lookups, permission checks, label i18n) will be felt at scale. Consider either splitting into `hasTaskContextMenuItems(args): boolean` + `getTaskContextMenuItems(args): NodeMenuItem[]` so the trigger predicate is cheap, or documenting in TSDoc that the callback **must** be cheap and stable.

2. **`React.memo` defeat.** `getContextMenuItems: () => [...]` is a fresh inline arrow every render. `DraggableTask`, `AdhocTaskItem`, and `EventDrivenTaskItem` are all wrapped in `memo()`. Before this PR they received the stable `useCallback` reference `buildContextMenuItems` (or its equivalent) — now every parent re-render passes a new function ref, busting the shallow prop comparison and forcing every task row to re-render whenever any unrelated parent state changes (selection, drag projection, execution status of a sibling). Wrapping the arrow in `useCallback` with `[buildContextMenuItems, additionalMenuItems, groupIndex, taskIndex, task.id]` deps is one fix; restructuring `DraggableTask`'s prop to accept the additional items separately and merge internally is another.

packages/apollo-react/src/canvas/components/StageNode/StageNode.types.ts
```diff
@@ -30,6 +30,12 @@ export interface StageTaskItem {
   hasEntryCondition?: boolean;
 }
 
+export interface StageTaskContextMenuArgs {
+  task: StageTaskItem;
+  taskGroupType: 'sequential' | 'event-driven' | 'adhoc';
+  isParallel: boolean;
```
uipath-utkarsh, line 36: **New public API needs TSDoc, and the args shape has two small modelling issues.**

```ts
export interface StageTaskContextMenuArgs {
  task: StageTaskItem;
  taskGroupType: 'sequential' | 'event-driven' | 'adhoc';
  isParallel: boolean;
}

// on StageNodeBaseProps:
getTaskContextMenuItems?: (args: StageTaskContextMenuArgs) => NodeMenuItem[] | undefined;
```

1. **No TSDoc on either type.** This is package-exported public API (re-exported via `StageNode/index.ts`). IDE hover won't tell consumers (a) that the callback runs on every render and so must be stable/cheap, (b) how the returned items are merged with built-in actions — currently `[...builtIn, ...additional]` with no divider between regroup options and additional items, so consumer items abut the regroup section without separation, or (c) that `isParallel` is meaningful only when `taskGroupType === 'sequential'`.

2. **`isParallel` is a dead axis for adhoc/event-driven** — always hardcoded `false` at the call sites in `StageNodeAdhocTaskGroups.tsx:78` and `StageNodeEventDrivenTaskGroups.tsx:80`. A consumer can write code that branches on `isParallel` for an adhoc task and it will silently "work" today only because the value is always `false`. Discriminated-union models this honestly:

   ```ts
   export type StageTaskContextMenuArgs =
     | { task: StageTaskItem; taskGroupType: 'sequential'; isParallel: boolean }
     | { task: StageTaskItem; taskGroupType: 'event-driven' | 'adhoc' };
   ```

3. **`taskGroupType` duplicates `StageTaskItem.taskGroupType`** (line 29 — same union). Worth extracting a shared `type StageTaskGroupType = 'sequential' | 'event-driven' | 'adhoc'` and referencing it from both, or dropping the field from the args (consumers can read `args.task.taskGroupType`). If the two are ever permitted to differ (rendering bucket vs intrinsic type), TSDoc should spell out which is authoritative — currently it's ambiguous.

uipath-utkarsh (commented):
## Follow-up review on 8475875c

Thanks for the refactor — moving from `isInSolutionProject` + `onGoToDefinition` to a single generic `getTaskContextMenuItems(args)` callback resolves the prior concerns about leaky UiPath-specific naming, the breaking change to `StageNodeLabels`, and the missing locale fallback, and the `hasBuiltInTaskActions` predicate restores the menu-visibility contract. Two remaining issues worth addressing — see inline.

uipath-utkarsh (approved):

packages/apollo-react/src/canvas/components/StageNode/StageNodeSequentialTaskGroups.tsx
```diff
@@ -211,6 +217,15 @@ export const StageNodeSequentialTaskGroups = ({
                     )}
                     {taskGroup.map((task, taskIndex) => {
                       const taskExecution = execution?.taskStatus?.[task.id];
+                      const additionalMenuItems = isReadOnly
```
SreedharAvvari: One follow-up on the generic API implementation: `getTaskContextMenuItems` is evaluated during render for every task so we can decide whether to render the task menu trigger, and then each task receives a fresh inline `getContextMenuItems` arrow. This keeps the callback on the hot render path and also weakens the memoization on task rows. The same pattern exists in the event-driven and adhoc task group files.

Suggested change: keep full menu construction lazy and only use the custom callback as a render-time probe when there are no built-in task actions. For example:

```ts
const customItems = !isReadOnly && !hasBuiltInTaskActions
  ? (getTaskContextMenuItems?.({ task, taskGroupType: 'sequential', isParallel }) ?? [])
  : [];

const hasMenu = !isReadOnly && (hasBuiltInTaskActions || customItems.length > 0);

// When opening the menu, build the full list lazily:
getContextMenuItems: () => [
  ...(getTaskContextMenuItems?.({ task, taskGroupType: 'sequential', isParallel }) ?? []),
  ...buildContextMenuItems(groupIndex, taskIndex, task.id),
]
```

Alternatively, if we expect consumers to provide non-trivial permission/lookup logic, split the API into a cheap `hasTaskContextMenuItems(args)` predicate plus the lazy `getTaskContextMenuItems(args)` builder. That keeps Apollo generic while avoiding a per-task/per-render callback cost.

SreedharAvvari (approved):

packages/apollo-react/src/canvas/components/StageNode/StageNodeSequentialTaskGroups.tsx
ayush578: This looks good to me

ayushya (approved):

PR #751, opened 2026-05-26
feat(apollo-react): add stage and task status tooltip [MST-10211]

ayush578:
added general locale prop to affectively check the localization effect.
added localization to some sticky note left out strings
added stage and tasks status icons tooltip

https://github.com/user-attachments/assets/8ae33bfb-88be-4851-8726-8fee6a31a9bd

### DIFF
--- a/apps/storybook/.storybook/preview.tsx
+++ b/apps/storybook/.storybook/preview.tsx
@@ -1,6 +1,7 @@
 import CssBaseline from '@mui/material/CssBaseline';
 import { ThemeProvider } from '@mui/material/styles';
 import type { Preview } from '@storybook/react';
+import { SUPPORTED_LOCALES } from '@uipath/apollo-react/i18n';
 import {
   apolloMaterialUiThemeDark,
   apolloMaterialUiThemeDarkHC,
@@ -61,6 +62,23 @@ const allThemes: ThemeMode[] = [
   'canvas',
 ];
 
+// Human-readable display names for the locale toolbar.
+const LOCALE_LABELS: Record<(typeof SUPPORTED_LOCALES)[number], string> = {
+  en: 'English',
+  es: 'Español',
+  pt: 'Português',
+  de: 'Deutsch',
+  fr: 'Français',
+  ja: '日本語',
+  ko: '한국어',
+  ru: 'Русский',
+  tr: 'Türkçe',
+  'zh-CN': '中文 (简体)',
+  'zh-TW': '中文 (繁體)',
+  'pt-BR': 'Português (Brasil)',
+  'es-MX': 'Español (México)',
+};
+
 // Map every theme to its closest MUI equivalent (used by stories with `parameters.material`)
 const muiThemeMap: Record<ThemeMode, typeof apolloMaterialUiThemeLight> = {
   light: apolloMaterialUiThemeLight,
@@ -107,6 +125,7 @@ const preview: Preview = {
   initialGlobals: {
     theme: 'future-dark',
     reactScan: 'off',
+    locale: 'en',
   },
   parameters: {
     backgrounds: { disable: true },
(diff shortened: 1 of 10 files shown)


apps/storybook/.storybook/preview.tsx
```diff
@@ -234,6 +265,14 @@ const preview: Preview = {
         }
       }, [reactScanEnabled]);
 
+      // Apply <html lang> so ApI18nProvider's fallback detection picks up the
+      // current locale (it reads document.documentElement.lang on mount). The
+      // `key={locale}` below remounts the story subtree on switch so the
+      // ApI18nProvider effect re-runs with the new fallback.
+      useEffect(() => {
+        document.documentElement.lang = locale;
```
uipath-utkarsh, line 295: **Locale toolbar has a one-switch lag because `<html lang>` is written in `useEffect` but `ApI18nProvider` reads it synchronously during render.**

```ts
useEffect(() => {
  document.documentElement.lang = locale;
}, [locale]);
```

`ApI18nProvider` reads `document.documentElement.lang` synchronously inside its render body (see [`ApI18nProvider.tsx:63`](https://github.com/UiPath/apollo-ui/blob/0c812fb16e11ec58186532870bd52137e8a8c8be/packages/apollo-react/src/i18n/ApI18nProvider.tsx#L62-L66)). On a locale switch `en → fr`:

1. Decorator re-renders with `locale='fr'` and schedules a new `useEffect`.
2. `<div key='fr'>` unmounts the previous story subtree and mounts a fresh one.
3. The new `ApI18nProvider` inside the story runs its render body — `document.documentElement.lang` is **still `'en'`** because the parent's `useEffect` hasn't committed yet. It activates `'en'`.
4. Effects fire bottom-up after commit, so the child's effects run first; only then does the decorator's effect set `lang = 'fr'`. No re-render is triggered, so the provider stays on `'en'`.

Result: every locale switch lags by one toolbar pick. Either write `document.documentElement.lang = locale` synchronously during the decorator's render (safe for a Storybook iframe), or thread the locale into `ApI18nProvider` via an explicit `locale` prop / context instead of relying on the `<html lang>` side-channel. The current comment on line 269-271 ("`key={locale}` below remounts the story subtree on switch so the `ApI18nProvider` effect re-runs with the new fallback") is incorrect — `ApI18nProvider` reads `lang` during render, not in an effect.

packages/apollo-react/src/canvas/components/StageNode/TaskContent.tsx
```diff
@@ -141,21 +147,18 @@ export const TaskContent = memo(
...
-              ) : (
-                <ExecutionStatusIcon status={taskExecution.status} />
-              ))}
+            {hasExecutionStatus && (
+              <CanvasTooltip content={taskStatusTooltip} placement="top">
+                <Button
+                  variant="ghost"
+                  size="icon"
+                  className="h-4 w-4"
+                  aria-label={taskStatusFallbackName}
```
uipath-utkarsh: **Screen-reader users lose the host's error detail that sighted users still see in the tooltip.**

```ts
const taskStatusTooltip = taskExecution?.message || taskStatusFallbackName;
...
<Button
  variant="ghost"
  size="icon"
  className="h-4 w-4"
  aria-label={taskStatusFallbackName}   // <-- only the status name
>
```

The visible tooltip prioritizes `taskExecution.message` (e.g., `"Activity X threw NullReferenceException"`), but `aria-label` is hardcoded to the generic status name (e.g., `"Failed"`). A screen-reader user focusing the icon hears only "Failed, button" while sighted users see the full error. The pre-PR code used `aria-label={taskExecution.message}` when the message was supplied — this PR removes that path.

The same pattern lives at [`StageNodeHeader.tsx:87`](https://github.com/UiPath/apollo-ui/blob/0c812fb16e11ec58186532870bd52137e8a8c8be/packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.tsx#L85-L89) (stage-level status icon). I see the new tests in `TaskContent.test.tsx` and `StageNode.test.tsx` explicitly codify this as intentional — but it's a WCAG 1.1.1 / 4.1.2 concern: the accessible name should carry equivalent information to the visible label. Two options that keep the status-name-as-fallback but preserve detail when available:

```ts
// Option A: include both — status as context + message as detail
aria-label={taskExecution?.message
  ? `${taskStatusFallbackName}: ${taskExecution.message}`
  : taskStatusFallbackName}

// Option B: use aria-describedby pointing at the tooltip surface,
// keeping aria-label as the short status name
```

Worth deciding whether the tests are enforcing the right contract here before merging.

uipath-utkarsh (commented):
## Code review

Two findings worth addressing. Both surfaced from independent review passes and verified against the referenced code.

apps/storybook/.storybook/preview.tsx
```diff
@@ -61,6 +62,23 @@ const allThemes: ThemeMode[] = [
   'canvas',
 ];
 
+// Human-readable display names for the locale toolbar.
+const LOCALE_LABELS: Record<(typeof SUPPORTED_LOCALES)[number], string> = {
```
kittyyueli, line 66: Its missing romanian~

kittyyueli (approved):

apps/storybook/.storybook/preview.tsx
ayush578: romanian is also missing from actual tranlations.
created a separate ticket for it.
https://uipath.atlassian.net/browse/MST-10307

uipath-utkarsh (approved):

PR #752, opened 2026-05-26
fix(apollo-react): addNodePanel scrolling [MST-9993]

ayush578:
Make addNodePanel scrollable. 
Also blocked the outer scroll i.e. FloatingCanvasPanel scroll only for addnodePanel

before

https://github.com/user-attachments/assets/16804bd2-3788-415a-9897-49bba10969aa

after

https://github.com/user-attachments/assets/725da5a3-756e-4024-beb4-2d7290045f7f

### DIFF
--- a/packages/apollo-react/src/canvas/components/AddNodePanel/AddNodeManager.tsx
+++ b/packages/apollo-react/src/canvas/components/AddNodePanel/AddNodeManager.tsx
@@ -267,6 +267,9 @@ export const AddNodeManager: React.FC<AddNodeManagerProps> = ({
       nodeId={PREVIEW_NODE_ID}
       placement="right-start"
       offset={FLOATING_CANVAS_PANEL_OFFSET}
+      // The default AddNodePanel (Toolbox) manages its own virtualized scroll,
+      // so the chrome's outer scroll is disabled to avoid a redundant scrollbar.
+      scrollableContent={!!CustomPanel}
     >
       {CustomPanel ? (
         <CustomPanel onNodeSelect={(item) => handleNodeSelect(item)} onClose={handleClose} />
--- a/packages/apollo-react/src/canvas/components/FloatingCanvasPanel/FloatingCanvasPanel.tsx
+++ b/packages/apollo-react/src/canvas/components/FloatingCanvasPanel/FloatingCanvasPanel.tsx
@@ -1,7 +1,7 @@
 import type { Placement } from '@floating-ui/react';
 import { ViewportPortal } from '@uipath/apollo-react/canvas/xyflow/react';
 import { cn } from '@uipath/apollo-wind';
-import type { ReactNode } from 'react';
+import type { CSSProperties, ReactNode } from 'react';
 import { useMemo } from 'react';
 import { createPortal } from 'react-dom';
 import { CanvasPortal } from './CanvasPortal';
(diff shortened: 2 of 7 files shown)


packages/apollo-react/src/canvas/components/FloatingCanvasPanel/FloatingCanvasPanel.tsx
```diff
@@ -150,6 +178,7 @@ export function FloatingCanvasPanel({
         style={{
           position: 'fixed',
           ...screenPosition,
+          ...sizingStyle,
```
uipath-utkarsh, line 193: **`...sizingStyle` is dead in the `useFixedPosition` branch — fixed-position panels never get the viewport cap.**

`sizingStyle` is only non-empty once `availableHeight` is populated, and `availableHeight` is produced by the `size` middleware in `useFloatingPosition`. That middleware only runs when floating-ui has both a reference and a floating element registered. In this fixed branch, neither is wired up:

- `refs.setFloating` is attached only on the non-fixed `panelContent` div (line 208).
- the reference element (`mergedReferenceRef`) is rendered only in the non-fixed return (line 239).

So in fixed mode `whileElementsMounted: autoUpdate` never starts, `size.apply` never fires, `availableHeight` stays `null`, and this `...sizingStyle` spread is always `{}`. Fixed-position panels keep only the static `max-h-[600px]` class and are never clamped to the viewport.

This is latent today (only `Flow.stories` uses `useFixedPosition`), but it's a silent gap: either wire the floating-ui refs into the fixed branch so `size` can measure, or drop the spread here and document that fixed panels aren't viewport-capped. As written it reads as if the cap applies when it can't.

packages/apollo-react/src/canvas/components/FloatingCanvasPanel/FloatingCanvasPanel.tsx
```diff
@@ -77,23 +92,36 @@ export function FloatingCanvasPanel({
...
     [isPinned]
   );
 
+  // Viewport-aware ceiling derived from the `size` middleware. The inline
+  // `maxHeight` overrides the `max-h-[600px]` class, so it's clamped to the
+  // design ceiling to keep the panel from exceeding 600px on tall viewports.
+  const sizingStyle = useMemo<CSSProperties>(() => {
+    if (isPinned || availableHeight == null) return {};
+    return {
+      maxHeight: `${Math.min(PANEL_FLOATING_MAX_HEIGHT, availableHeight)}px`,
```
uipath-utkarsh: **Two things about this `sizingStyle` block:**

1. **No lower floor on the clamp.** `Math.min(PANEL_FLOATING_MAX_HEIGHT, availableHeight)` is only an upper bound, and `availableHeight` is floored to `0` in the `size` middleware (`Math.max(0, Math.floor(ah))`). When the anchor node sits hard against a viewport edge on a short screen, `availableHeight` can resolve to a few px (or `0`), so `maxHeight` becomes e.g. `0px` while the panel is still `open` — collapsing it to an invisible/unusable sliver with no way to reach the close affordance. Consider a floor, e.g. `Math.max(MIN_PANEL_HEIGHT, Math.min(PANEL_FLOATING_MAX_HEIGHT, availableHeight))`.

2. **The CSS var exposes the *unclamped* value (line 121).** The panel caps itself at `min(600, availableHeight)`, but `--floating-available-height` is handed to children as the raw `availableHeight`, and it doesn't subtract any `PanelChrome` header height. It's benign today (`TOOLBOX_HEIGHT` 440 < 600, and the AddNodePanel renders no panel header), but a child that trusts the var as "the height the panel actually occupies" would oversize past the 600 ceiling or under a header. Worth either clamping the var to match the panel's own `maxHeight` or documenting that it's pre-clamp.

packages/apollo-react/src/canvas/components/AddNodePanel/AddNodeManager.tsx
```diff
@@ -267,6 +267,9 @@ export const AddNodeManager: React.FC<AddNodeManagerProps> = ({
       nodeId={PREVIEW_NODE_ID}
       placement="right-start"
       offset={FLOATING_CANVAS_PANEL_OFFSET}
+      // The default AddNodePanel (Toolbox) manages its own virtualized scroll,
+      // so the chrome's outer scroll is disabled to avoid a redundant scrollbar.
+      scrollableContent={!!CustomPanel}
```
uipath-utkarsh, line 272: **The chrome-scroll opt-out lives at this one call site, but the same self-scrolling Toolbox is hosted elsewhere without it.**

`scrollableContent={!!CustomPanel}` correctly disables the chrome's outer `overflow-y:auto` for the default (self-scrolling) `AddNodePanel`/`Toolbox`. But the same component is rendered inside a `FloatingCanvasPanel` at other call sites that did **not** get this prop and therefore keep the `true` default:

- `BlankCanvasNode.tsx:47` — `<AddNodePanel>` in a panel
- `StageNode.tsx:161,173` — `<Toolbox>` for Add/Replace task
- `Flow.stories.tsx:371`

Because the Toolbox now self-caps via `--floating-available-height`, this doesn't produce a *visible* double scrollbar today (those sites pass `title` to the Toolbox, not to the panel, so there's no `PanelChrome` header and `PanelContent` ends up flush with the Toolbox). But the opt-out is really a property of "this panel hosts a child that manages its own scroll," so it belongs at the `AddNodePanel`/`Toolbox` layer (or as the default for those panels) rather than being duplicated per caller. As-is, a future Toolbox layout change could silently reintroduce the redundant outer scrollbar at the sites that forgot the prop.

uipath-utkarsh (commented):
## Code review

The scrolling fix and the viewport-aware `size` cap are well put together (nice touches: the design-ceiling clamp, the reset-on-close to avoid a stale frame, the kept-in-sync comments). Three things worth addressing before merge — see inline. None block the AddNodeManager happy path; they're about the shared `FloatingCanvasPanel` contract and edge cases.

uipath-utkarsh (approved):

PR #753, opened 2026-05-26
feat(apollo-react): add LoopNodeExecutionCount component

dbacomputer:
## Summary

- Extracts the unified segmented pill from LoopNode V2 proposals into a standalone exported component: `LoopNodeExecutionCount`
- Adds `LoopNodeExecutionCountState` and `iterationPillState` as a first-class prop on `LoopNode` — the component derives the size tier from its measured width automatically, no overlay wrappers needed
- Exports `LoopNodeExecutionCount` and `LoopNodeExecutionCountState` from the LoopNode module index
- Adds a Wind-style **Execution Count** doc page under `Components/Nodes/LoopNode` with:
  - Live canvas preview with four stacked nodes: default All state, full tier, compact tier, minimal tier
  - Expand/collapse overlay (90vw × 90vh)
  - **Anatomy** section (collapsible): default All-state pill, three responsive tier cards with live interactive pills, breakpoint spec table, Count Range (1–999) visual
  - **Header Layout** section (collapsible): live header mockup with annotated A/B/C callouts and region spec table
  - **How to use** section (collapsible): copy-ready usage snippet
  - Expand all / Collapse all toggle above the three sections
- Adds unit tests (`LoopNodeExecutionCount.test.tsx`) covering all key behaviors: navigation, All toggle, inline edit, jump-to-failed, size tiers, and status color mapping

## Component: `LoopNodeExecutionCount`

**Props:**
- `state: LoopNodeExecutionCountState` — active index, total, navigation callbacks, per-iteration statuses, overall status
- `size?: 'full' | 'compact' | 'minimal'` — responsive tier (auto-computed by `LoopNode` from its measured width)

**Size tiers:**
| Node width | Tier | Controls |
|---|---|---|
| ≥ 400 px | `full` | All toggle · ‹ prev · k/N (click-to-type) · next › · jump-to-failed ⊕ |
| 260–399 px | `compact` | All toggle · k/N (click-to-type) · jump-to-failed ⊕ |
| < 260 px | `minimal` | Read-only count chip |

**Default state:** pill opens in the `All` view showing aggregate execution summary (✓ completed / ✗ failed). Clicking **All** toggles into individual iteration navigation.

## Test plan
- [ ] Open Storybook → Components / Nodes / LoopNode / Execution Count
- [ ] Verify four canvas nodes render stacked vertically with correct tier sizes and default All state on the top node
- [ ] Interact: All toggle, prev/next navigation, click-to-type jump, jump-to-failed crosshair
- [ ] Expand canvas preview to 90vw × 90vh and back — single canvas instance, no ReactFlow errors
- [ ] Open/close each collapsible section individually; verify Expand all / Collapse all button
- [ ] Verify Count Range row shows pills from 1/1 through 999/999 without overflow
- [ ] Interact with the Header Layout visual mockup pill
- [ ] `pnpm --filter @uipath/apollo-react run format:check` passes
- [ ] Unit tests pass: `37 tests` covering navigation, All toggle, inline edit, jump-to-failed, size tiers, status colors


packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,315 @@
...
+    case 'Failed':
+      return '#ef4444';
+    case 'InProgress':
+      return '#f59e0b';
+    case 'Paused':
+      return '#a855f7';
+    case 'Cancelled':
+      return '#94a3b8';
+    default:
+      return 'currentColor';
```
CalinaCristian: I don't like hardcoded colors here, we should use wind tokens. Especially since this is not theme friendly.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,315 @@
...
+import { useRef, useState } from 'react';
+import type { ElementStatusValues } from '../../types/execution';
+import { CanvasIcon } from '../../utils/icon-registry';
+import type { LoopNodeExecutionCountState } from './LoopNode.types';
+
+function stopEvent(e: React.SyntheticEvent) {
+  e.stopPropagation();
+}
+
+export function getIterationStatusColor(status: string | undefined): string {
```
CalinaCristian: we should have an enum for status IMO

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+  const currentStatus = iterationStatuses?.get(activeIndex);
+  const firstFailedIndex = iterationStatuses
+    ? [...iterationStatuses.entries()].find(([, s]) => s === 'Failed')?.[0]
+    : undefined;
+  const completedCount = iterationStatuses
+    ? [...iterationStatuses.values()].filter((s) => s === 'Completed').length
+    : undefined;
+  const failedCount = iterationStatuses
+    ? [...iterationStatuses.values()].filter((s) => s === 'Failed').length
```
CalinaCristian: enum could be used here then as well

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,315 @@
...
+        className="nodrag nopan pointer-events-auto flex items-center"
+        onPointerDown={stopEvent}
+        onMouseDown={stopEvent}
+      >
+        <div className="flex h-6 items-center gap-0.5 rounded-full border border-border bg-surface px-2 text-[11px] font-semibold leading-none shadow-sm">
+          {isAll ? (
+            completedCount !== undefined ? (
+              <>
+                <span style={{ color: getIterationStatusColor('Completed') }}>
+                  ✓{completedCount}
```
CalinaCristian, line 131: can we use a lucide icon instead of `✓` ?

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,315 @@
...
+                </span>
+                {failedCount > 0 && (
+                  <span style={{ color: getIterationStatusColor('Failed') }}> ✗{failedCount}</span>
```
CalinaCristian: all those references to `getIterationStatusColor` would benefit from the enum as well

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,315 @@
...
+        {/* Right segment — aggregate or navigation */}
+        {isAll ? (
+          <button
+            type="button"
+            onClick={toggleAll}
+            onPointerDown={stopEvent}
+            onMouseDown={stopEvent}
+            className="nodrag nopan flex items-center gap-1.5 px-2.5 text-[11px] font-semibold leading-none transition-colors hover:bg-surface-overlay"
+            aria-label="Return to individual iteration view"
+            title="Click to return to individual iteration view"
```
CalinaCristian: we should use i18n for those

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,315 @@
...
+        canInteract &&
+        overallStatus !== ('Failed' as ElementStatusValues) && (
+          <button
+            type="button"
+            className="nodrag nopan inline-flex h-6 w-6 items-center justify-center rounded-full border border-border bg-surface shadow-sm transition-colors hover:border-red-400"
+            onClick={handleJumpToFailed}
+            onPointerDown={stopEvent}
+            onMouseDown={stopEvent}
+            aria-label="Jump to first failed iteration"
+            title={`Jump to iteration ${firstFailedIndex + 1} (failed)`}
```
CalinaCristian: same, everything that is consumer facing needs to be translated.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Agreed. Noting here that Cristian and I discussed this. Added to the backlog to fix this in a future PR and leaving this unresolved so we have as reference.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Noting here that Cristian and I discussed this. Unicode character was selected to be used due to size. For now we still stick with this route. Leaving this open and unresolved in case we want to come back and reference for a future update.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Updated: IterationStatus enum — exported const object with all 5 statuses; switch cases in getIterationStatusColor and all s === 'Failed' / s === 'Completed' comparisons throughout the component now reference IterationStatus.Failed / IterationStatus.Completed

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Updated: and included the info above.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Updated: i18n — useSafeLingui wired in, all 8 consumer-facing strings wrapped with _() and proper IDs: the All toggle, return-to-individual view, prev/next iteration, click-to-jump, jump-to-failed, and the dynamic jump-to-iteration title

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Updated: and included the info above.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Updated: and included the info above.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,355 @@
...
+  } = state;
+
+  const { _ } = useSafeLingui();
+  const [isEditing, setIsEditing] = useState(false);
+  const [inputValue, setInputValue] = useState('');
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const canInteract = !disabled && typeof onActiveIndexChange === 'function';
+  const visibleIndex = activeIndex + 1;
+  const clampToRange = (v: number) => Math.max(1, Math.min(total, v));
```
SreedharAvvari: We have clamp utility available in https://github.com/UiPath/apollo-ui/blob/main/packages/apollo-react/src/canvas/utils/NodeUtils.ts#L111

packages/apollo-react/src/canvas/components/LoopNode/LoopNode.tsx
```diff
@@ -384,6 +390,8 @@ function LoopNodeComponent(props: LoopNodeProps) {
         loading={isLoading}
         isParallel={isParallel}
         iterationState={iterationStateProp}
+        iterationPillState={iterationPillStateProp}
+        nodeWidth={width}
```
SreedharAvvari: Issue: The pill tier is derived from the raw React Flow width, but the node renders with a fallback containerWidth when width is 0. Initial/unmeasured renders can show the minimal pill inside a default-width node.

SreedharAvvari: Suggested change: pass containerWidth into Header, or compute the execution-count tier from the same fallback width used by the node style.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,355 @@
...
+  };
+
+  const handleNext = (e: React.MouseEvent) => {
+    e.stopPropagation();
+    if (canInteract && !isAll && activeIndex < total - 1) onActiveIndexChange?.(activeIndex + 1);
+  };
+
+  const toggleAll = (e: React.MouseEvent) => {
+    e.stopPropagation();
+    onAllChange(!isAll);
```
SreedharAvvari, line 87: Issue: disabled only gates index changes, but the All toggle still calls onAllChange. A disabled/read-only pill can still switch between aggregate and individual views.

SreedharAvvari: Suggested change: guard toggleAll when disabled, and mark the All/aggregate buttons disabled so the UI state matches the callback behavior.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
SreedharAvvari: Issue: activeIndex and total are used without the validation that IterationNavigator has. Non-finite, zero/negative, or out-of-range values can render invalid counts and emit invalid indexes.

SreedharAvvari: Suggested change: normalize total and activeIndex before rendering, ideally by extracting/reusing the existing IterationNavigator resolution logic, then clamp edits against the normalized total.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,355 @@
+import { cn } from '@uipath/apollo-wind';
+import { useRef, useState } from 'react';
+import { useSafeLingui } from '../../../i18n';
+import type { ElementStatusValues } from '../../types/execution';
+import { CanvasIcon } from '../../utils/icon-registry';
+import type { LoopNodeExecutionCountState } from './LoopNode.types';
+
+export const IterationStatus = {
```
SreedharAvvari: i think we can use [ElementStatusValues](https://github.com/UiPath/apollo-ui/blob/main/packages/apollo-react/src/canvas/types/execution.ts#L1) here

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,355 @@
...
+            )}
+          </div>
+        )}
+      </div>
+
+      {/* Jump-to-failed shortcut — hidden when loop is globally Failed */}
+      {firstFailedIndex !== undefined &&
+        !isAll &&
+        canInteract &&
+        overallStatus !== ('Failed' as ElementStatusValues) && (
```
SreedharAvvari: > Jump-to-failed shortcut — hidden when loop is globally Failed

Just curious behind the reasoning for this, why would we hide the jump shortcut when the overall loop status is failed 🤔

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,355 @@
...
+            className="nodrag nopan inline-flex h-6 w-6 items-center justify-center rounded-full border border-border bg-surface shadow-sm transition-colors hover:border-red-400"
+            onClick={handleJumpToFailed}
+            onPointerDown={stopEvent}
+            onMouseDown={stopEvent}
+            aria-label={_({
+              id: 'loop-node.execution-count.jump-to-failed',
+              message: 'Jump to first failed iteration',
+            })}
+            title={_({
+              id: 'loop-node.execution-count.jump-to-iteration',
```
SreedharAvvari: I believe we also need to add all such string keys/ values to https://github.com/UiPath/apollo-ui/blob/main/packages/apollo-react/src/canvas/locales/en.json#L45 so that they can be picked up by the  localisation pipeline.

packages/apollo-react/src/canvas/components/LoopNode/LoopNode.tsx
```diff
@@ -502,37 +492,42 @@ function Header({
...
       </div>
       <div className="flex shrink-0 items-center gap-2">
-        {iterationState ? <IterationNavigator iterationState={iterationState} /> : null}
+        {iterationPillState ? (
+          <LoopNodeExecutionCount
+            state={iterationPillState}
+            size={nodeWidth >= 400 ? 'full' : nodeWidth >= 260 ? 'compact' : 'minimal'}
+          />
+        ) : iterationState ? (
+          <IterationNavigator iterationState={iterationState} />
```
SreedharAvvari, line 501: Note: we haven’t integrated the existing loop `IterationNavigator` into Flow yet. Given this newer design, I think this might be good opportunity to consolidate around `LoopNodeExecutionCount`  & cleanup the older code path. 
I can handle that cleanup as a follow-up once this PR is merged 👍

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Great catch. This behavior was leftover by accident from the initial prototyping and has been addressed and updated.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Fixed: now imports and uses clamp from NodeUtils.ts — used in both safeActiveIndex computation and commitEdit to clamp the jump input.

packages/apollo-react/src/canvas/components/LoopNode/LoopNode.tsx
dbacomputer: Fixed: changed nodeWidth={width} to nodeWidth={containerWidth} in the Header call. The tier is now derived from the same fallback width the node layout uses, so the pill is never minimal inside a default-size node on first render.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Fixed: toggleAll now early-returns when disabled. Both the All button and the aggregate button have disabled={disabled} and a cursor-not-allowed/opacity-50 style to match the visual state to the callback behavior.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Fixed: safeTotal and safeActiveIndex are computed upfront before any rendering. safeTotal guards against non-finite/zero/negative values (falls back to 1), and safeActiveIndex is clamped to [0, safeTotal - 1]. All navigation, rendering, and input handling use these normalized values.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Fixed: replaced the custom IterationStatus enum with ElementStatusValues from execution.ts. The iterationStatuses map type is now Map<number, ElementStatusValues>, getIterationStatusColor uses ElementStatusValues in its signature and switch cases, and all comparison sites use ElementStatusValues.Failed / ElementStatusValues.Completed.

packages/apollo-react/src/canvas/components/LoopNode/LoopNode.tsx
dbacomputer: Sounds good — happy to leave that cleanup to you as a follow-up once this is merged. Thanks for offering.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Fixed: all 8 i18n string keys have been added to en.json so they are picked up by the localization pipeline.

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
dbacomputer: Good catch — the guard was an artifact left over from early prototyping, not an intentional design decision. Removed: the jump-to-failed button now shows whenever firstFailedIndex is defined, regardless of overall loop status.

packages/apollo-react/src/canvas/components/LoopNode/LoopNode.types.ts
```diff
@@ -20,12 +20,24 @@ export interface LoopIterationState {
...
 
+export interface LoopNodeExecutionCountState {
+  activeIndex: number;
+  total: number;
+  onActiveIndexChange?: (nextIndex: number) => void;
+  disabled?: boolean;
+  isAll: boolean;
+  onAllChange: (isAll: boolean) => void;
+  iterationStatuses?: Map<number, ElementStatusValues>;
+  overallStatus?: ElementStatusValues;
```
SreedharAvvari: i guess this can be removed now - `overallStatus`

packages/apollo-react/src/canvas/components/LoopNode/LoopNodeExecutionCount.tsx
```diff
@@ -0,0 +1,365 @@
...
+          onPointerDown={stopEvent}
+          onMouseDown={stopEvent}
+          aria-label={_({
+            id: 'loop-node.execution-count.jump-to-failed',
+            message: 'Jump to first failed iteration',
+          })}
+          title={_({
+            id: 'loop-node.execution-count.jump-to-iteration',
+            message: `Jump to iteration ${firstFailedIndex + 1} (failed)`,
+          })}
```
SreedharAvvari, line 355:
```suggestion
          title={_({
            id: 'loop-node.execution-count.jump-to-iteration',
            message: 'Jump to iteration {iterationNumber} (failed)',
            values: { iterationNumber: firstFailedIndex + 1 },
          })}
```

packages/apollo-react/src/canvas/locales/en.json
```diff
@@ -1,4 +1,12 @@
 {
+  "loop-node.execution-count.show-aggregate": "Show aggregate across all iterations",
+  "loop-node.execution-count.return-to-individual": "Return to individual iteration view",
+  "loop-node.execution-count.click-to-return": "Click to return to individual iteration view",
+  "loop-node.execution-count.previous-iteration": "Previous iteration",
+  "loop-node.execution-count.next-iteration": "Next iteration",
+  "loop-node.execution-count.click-to-jump": "Click to jump to a specific iteration",
+  "loop-node.execution-count.jump-to-failed": "Jump to first failed iteration",
+  "loop-node.execution-count.jump-to-iteration": "Jump to iteration (failed)",
```
SreedharAvvari:
```suggestion
  "loop-node.execution-count.jump-to-iteration": "Jump to iteration {iterationNumber} (failed)",
```

SreedharAvvari (approved):

SreedharAvvari:
Couple of minor review comments but looks good overall. Thanks for the PR.

dbacomputer:
I assume we'll have a few quick follow-up iterations once this lands, but it's a solid step forward for LoopNodes.

SreedharAvvari (approved):

PR #754, opened 2026-05-27
feat(apollo-vertex): charts insights adapter [AGVSOL-3412]

pieman1313:
ports over the stand alone adapter from apollo-dashboarding to our insights adapter
I tested it locally with a custom page with 2 charts hitting insights, but I`m not adding the test page to this pr / vertex
I can add a temporary commit if anyone wants to test as well


apps/apollo-vertex/lib/api-proxy.ts
```diff
@@ -0,0 +1,57 @@
+import { type NextRequest, NextResponse } from "next/server";
+
+const UPSTREAM_ORIGIN = "https://alpha.uipath.com";
```
alincadariu, line 3: will this always be alpha? 🤔 
shouldn't be configurable for the consumer?

pieman1313: this is only the proxy, consumers will not instal this; either way, we only whitelisted this app towards alpha, and its specific for this vercel app

alincadariu (approved):

PR #779, opened 2026-06-04
feat(apollo-react): route SequenceEdge through edge.data.bendPoints [MST-6513]

KodudulaAshishUiPath:
## Summary
Adds optional bend-point routing to `SequenceEdge`. When an edge carries `data.bendPoints` (interior corners in flow coordinates, produced by an auto-layout pass such as ELK), the edge routes through them instead of the default smooth-step path — so edges can avoid overlapping nodes.

## File Changes
- **`useEdgePath`**: new optional `bendPoints` param. The path is anchored on the live source/target handle positions and threaded through the interior bends with rounded corners. Two handle-alignment corrections keep it clean despite ELK routing from its *port* while React Flow renders the *handle*:
  1. perpendicular snap — first/last bend onto the handle's perpendicular axis (orthogonal exit/entry);
  2. parallel exit-gap — shifts the route so the riser sits a source-offset past the handle instead of on/behind it.
  Falls back to the existing smooth-step path when `bendPoints` is absent; loop/self edges ignore bends.
- **`SequenceEdge`**: forwards `data?.bendPoints` to `useEdgePath`.
- **`types.ts`**: documents `bendPoints` on `AgentFlowDefaultEdgeData` (transient/derived — recompute on layout, not persisted).
- **Storybook**: new `BendPoints` story (routed vs. straight comparison; clears bends on node drag-start so they follow the drag).

## Demo

https://github.com/user-attachments/assets/fa5ab8cc-6925-4237-8e00-a2bc8d9a8c87

## Notes
Purely additive — absent `bendPoints` ⇒ byte-for-byte the current smooth-step behavior, so no breaking change. `data.bendPoints` is consumed only; nothing is persisted.

## Testing
- 29 `useEdgePath` unit tests (routing through bends, perpendicular snap, exit-gap correction, smooth-step fallback, loop-edge ignore, recompute-on-change).
- Full `@uipath/apollo-react` suite (1664 tests) + `build` green.

### DIFF
--- a/packages/apollo-react/src/canvas/components/Edges/SequenceEdge.stories.tsx
+++ b/packages/apollo-react/src/canvas/components/Edges/SequenceEdge.stories.tsx
@@ -6,9 +6,9 @@
  */
 
 import type { Meta, StoryObj } from '@storybook/react';
-import type { Edge, EdgeTypes, Node } from '@uipath/apollo-react/canvas/xyflow/react';
+import type { Edge, EdgeTypes, Node, OnNodeDrag } from '@uipath/apollo-react/canvas/xyflow/react';
 import { Panel, Position } from '@uipath/apollo-react/canvas/xyflow/react';
-import { useMemo } from 'react';
+import { type Dispatch, type SetStateAction, useCallback, useMemo } from 'react';
 import { useCanvasStory, withCanvasProviders } from '../../storybook-utils';
 import { DefaultCanvasTranslations } from '../../types';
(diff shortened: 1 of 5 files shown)


KodudulaAshishUiPath:
Closing with PR as we are going forward with the new approach of Edge segments which re calculate the path on drag of connected nodes instead of fixed bends this PR uses. 
New Approach: https://github.com/UiPath/apollo-ui/pull/804

PR #782, opened 2026-06-05
feat(apollo-vertex): add Solution Tests view + template [AGVSOL-3355, AGVSOL-3792]

frankkluijtmans:
## Summary

The **UI + registry wiring** for the Solution Tests component, on top of the data
layer from #790 (now merged). This targets `main`. Once it lands,
`solution-tests` is installable from the registry.

Contents:

- `SolutionTestsView` (the dumb presentational view) plus the KPI bar, tabs,
  expandable rows, run-details dialog, evaluator results, JSON viewer,
  delete-confirm, run-confirm, and user-message popover.
- `solution-tests.tsx` — the smart `SolutionTests` container that calls the
  collection-backed hooks from #790 and feeds data + handlers into the view.
- A run-confirmation dialog: "Run" and "Run all" both prompt first, warning that
  running re-processes each case and can modify data in connected systems.
- `index.ts` barrel (the public surface) + the `registry.json` entry.
- `SolutionTestsTemplate` demo (in-memory mock, no vs-core) + the **Templates**
  docs page (`page.mdx`) and nav entry.

## Consumer interface

The view ships as a smart/dumb pair. Most consumers use the smart container,
which has three inputs, all on `SolutionTestsProvider` (two required, plus an
optional `config`). Reads, writes, and attachment id resolution are all sourced
from the consumer's vs-core solution; only writes and presentation are explicit
props:

```tsx
import {
  SolutionTests,
  SolutionTestsProvider,
  type SolutionTestsConfig,
} from "@/components/ui/solution-tests";

const config: SolutionTestsConfig = {
  subjectColumns, // the main per-vertical customization
  subjectNoun: { singular: "Loan", plural: "Loans" },
  // passThreshold defaults to 0.9; override per vertical if needed.
};

<SolutionTestsProvider
  config={config}
  // `{base}/{org}/{tenant}/orchestrator_/t/{folderKey}` (no trailing slash)
  triggerBaseUrl={`/${org}/${tenant}/orchestrator_/t/${folderKey}`}
  getToken={() => tokenService.getCurrentToken()}
>
  <SolutionTests />
</SolutionTestsProvider>;
```

A consumer is responsible for exactly three things:

1. **Reads** — ensure the name-keyed `UiPathST*` collections exist on the vs-core
   solution. The hooks query them reactively; no further data wiring. Attachment
   reads (expected / actual output, evaluator results) resolve each entity's
   DataFabric id from `solution.api.entityIds` (vs-core 2.0.6+), so they need no
   extra wiring either.
2. **Writes** — pass `triggerBaseUrl` + `getToken`. The provider builds the whole
   write surface (run, delete, force-stop, adopt / update / remove baseline) from
   them. Toggle-active isn't a trigger and needs no wiring. Mutations use
   `useMutation`, so the container must render under a `QueryClientProvider`
   (vs-core's solution provider supplies one).
3. **Presentation** — pass `config.subjectColumns` (and optionally
   `getSubjectHref` / `subjectNoun` / `passThreshold`).

Everything else (evaluator labels, status labels, the poll interval) is
hard-coded in `constants.ts`. The pass threshold defaults to `0.9` there but is
overridable via `config.passThreshold`. Consumers who want their own data
plumbing can render the dumb `SolutionTestsView` directly, passing all data and
callbacks via props (this is what the demo does). The docs page documents the
full interface.

## Dependency note

Attachment id resolution uses `useSolution().api.entityIds`, which vs-core
introduced in **2.0.6** (the resolved DataFabric id per entity, `__DF_MAPPING__`
override applied). The registry item declares `@uipath/vs-core@^2.0.6`
accordingly. This replaces the earlier consumer-injected `getEntityId` resolver,
which only saw the codegen `id` and missed the deploy-time mapping.

## Testing

`pnpm typecheck`, `pnpm lint`, `pnpm format:check`, and `pnpm registry:build`
all pass.

> Split out of the original combined PR per review feedback: data layer in #790
> (merged), components here.


apps/apollo-vertex/registry/solution-tests/solution-tests-view.tsx
```diff
@@ -0,0 +1,586 @@
...
+    expanded,
+    onExpandedChange,
+  };
+}
+
+export interface SolutionTestsViewProps {
+  tests: SolutionTest[];
+  batchRuns: SolutionTestBatchRun[];
+  allRuns: SolutionTestRun[];
+  loading: boolean;
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: removed the unused `allRuns` prop from `SolutionTestsViewProps` and both call sites. The expanded-run rendering already uses the container's own hook data, not this prop. (b57982c9)

apps/apollo-vertex/registry/solution-tests/solution-tests-view.tsx
```diff
@@ -0,0 +1,586 @@
...
+              )}
+            </Tooltip>
+            <Tooltip>
+              <TooltipTrigger asChild>
+                <span className="inline-flex items-center">
+                  <Switch
+                    checked={test.IsActive !== false}
+                    disabled={isDisabling}
+                    onCheckedChange={(val) => onToggleActive(test.Id, val)}
+                  />
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: added an `aria-label` to the Switch (enabled/disabled, reusing the existing keys that drive the tooltip). (b57982c9)

apps/apollo-vertex/registry/solution-tests/solution-tests.tsx
```diff
@@ -0,0 +1,173 @@
...
+          );
+        }
+        setRunConfirm(null);
+      }}
+      onToggleActive={(testId, newValue) =>
+        toggleTracked(
+          testId,
+          () => toggleActive.mutate(testId, newValue),
+          t("failed_to_run_test"),
+        )
```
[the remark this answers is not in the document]
frankkluijtmans: Good catch, fixed: added a `failed_to_toggle_test` key ("Failed to update test status") and use it instead of `failed_to_run_test`. (b57982c9)

apps/apollo-vertex/registry/solution-tests/result-expanded-content.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { hasAutoPass } from "./utils";
+import { UserMessagesView } from "./user-messages-view";
+import { EvaluatorResultsView } from "./evaluator-results-view";
+
+/** Pretty-print arbitrary attachment data for display in a `<pre>` block. */
+function formatJson(data: unknown): string {
+  if (data == null) return "No data available";
+  if (typeof data === "string") return data;
+  return JSON.stringify(data, null, 2);
+}
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: `formatJson` now returns the `—` placeholder instead of the hard-coded English string, matching the empty-value convention used elsewhere in this folder. (b57982c9)

apps/apollo-vertex/app/templates/solution-tests/page.mdx
```diff
@@ -0,0 +1,179 @@
...
+- **Dialogs** — run details, JSON viewer, and delete confirmation, plus the
+  expected-vs-actual baseline adopt / update / remove flows.
+
+Source: [`templates/SolutionTestsTemplate.tsx`](https://github.com/UiPath/apollo-ui/blob/main/apps/apollo-vertex/templates/SolutionTestsTemplate.tsx)
+
+## Consumer interface
+
+The smart container has three inputs, all on `SolutionTestsProvider`. Reads are
+implicit (the `UiPathST*` collections on your vs-core solution); writes and
+presentation are explicit props:
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: the prose now says four inputs (three required, plus an optional `config`). (b57982c9)

apps/apollo-vertex/app/templates/solution-tests/page.mdx
```diff
@@ -0,0 +1,179 @@
...
+2. **Writes** — pass `triggerBaseUrl` + `getToken`. The provider builds the
+   whole write surface (run, delete, force-stop, adopt / update / remove
+   baseline) from them. Toggle-active isn't a trigger. It goes through the
+   collection's optimistic update and needs no extra wiring.
+3. **Presentation** — pass `config.subjectColumns` (and optionally
+   `getSubjectHref` / `subjectNoun`) to flavor the table for your vertical.
+
+Everything else (evaluator labels, status labels, the `0.9` pass threshold, the
+poll interval) is hard-coded in `constants.ts`. Edit there to retarget a
+deployment. Consumers who want to supply their own data plumbing can instead
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: the docs now state the threshold defaults to 0.9 and is overridable via `config.passThreshold` (corrected in both spots), and `passThreshold` was added to the documented `SolutionTestsConfig`. (b57982c9)

apps/apollo-vertex/registry/solution-tests/delete-confirm-dialog.tsx
```diff
@@ -0,0 +1,50 @@
+"use client";
+
+import { useTranslation } from "react-i18next";
+import { Button } from "@/components/ui/button";
```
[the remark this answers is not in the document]
frankkluijtmans: Not an issue in this repo. `@types/react` declares `export as namespace React`, so `React` is a global type and `React.FC` resolves without an import. Both `pnpm typecheck` and the fresh-project registry-check `tsc --noEmit` pass, and shipped components (e.g. spinner, skeleton, sonner) use bare `React.*` the same way. Resolving as a false positive.

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
```diff
@@ -0,0 +1,51 @@
+"use client";
+
+import { useSolutionTestsConfig } from "./context";
```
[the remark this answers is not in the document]
frankkluijtmans: Not an issue in this repo. `@types/react` declares `export as namespace React`, so `React` is a global type and `React.FC` resolves without an import. Both `pnpm typecheck` and the fresh-project registry-check `tsc --noEmit` pass, and shipped components (e.g. spinner, skeleton, sonner) use bare `React.*` the same way. Resolving as a false positive.

apps/apollo-vertex/registry/solution-tests/expanded-agents.tsx
```diff
@@ -0,0 +1,157 @@
+"use client";
+
+import { useState, useCallback } from "react";
+import { useQuery } from "@tanstack/react-query";
```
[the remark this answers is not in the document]
frankkluijtmans: Not an issue in this repo. `@types/react` declares `export as namespace React`, so `React` is a global type and `React.FC` resolves without an import. Both `pnpm typecheck` and the fresh-project registry-check `tsc --noEmit` pass, and shipped components (e.g. spinner, skeleton, sonner) use bare `React.*` the same way. Resolving as a false positive.

apps/apollo-vertex/registry/solution-tests/json-viewer-dialog.tsx
```diff
@@ -0,0 +1,56 @@
+"use client";
+
+import { Skeleton } from "@/components/ui/skeleton";
```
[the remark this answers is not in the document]
frankkluijtmans: Not an issue in this repo. `@types/react` declares `export as namespace React`, so `React` is a global type and `React.FC` resolves without an import. Both `pnpm typecheck` and the fresh-project registry-check `tsc --noEmit` pass, and shipped components (e.g. spinner, skeleton, sonner) use bare `React.*` the same way. Resolving as a false positive.

apps/apollo-vertex/registry/solution-tests/user-messages-view.tsx
```diff
@@ -0,0 +1,80 @@
+"use client";
+
+import { useTranslation } from "react-i18next";
```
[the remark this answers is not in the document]
frankkluijtmans: Not an issue in this repo. `@types/react` declares `export as namespace React`, so `React` is a global type and `React.FC` resolves without an import. Both `pnpm typecheck` and the fresh-project registry-check `tsc --noEmit` pass, and shipped components (e.g. spinner, skeleton, sonner) use bare `React.*` the same way. Resolving as a false positive.

apps/apollo-vertex/registry/solution-tests/context.tsx
```diff
@@ -0,0 +1,64 @@
+"use client";
+
+/**
+ * Wires the Solution Tests view to its data adapter and presentation config.
+ * Deeply-nested components (expanded rows, dialogs) read from here instead of
+ * importing a concrete service, so the whole tree is portable.
+ */
+
+import { createContext, useContext, useMemo, type ReactNode } from "react";
```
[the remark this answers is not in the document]
frankkluijtmans: Not an issue in this repo. `@types/react` declares `export as namespace React`, so `React` is a global type and `React.FC` resolves without an import. Both `pnpm typecheck` and the fresh-project registry-check `tsc --noEmit` pass, and shipped components (e.g. spinner, skeleton, sonner) use bare `React.*` the same way. Resolving as a false positive.

apps/apollo-vertex/registry/solution-tests/expanded-run-tests.tsx
```diff
@@ -0,0 +1,211 @@
+"use client";
+
+import { useState, useCallback } from "react";
```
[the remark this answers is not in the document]
frankkluijtmans: Not an issue in this repo. `@types/react` declares `export as namespace React`, so `React` is a global type and `React.FC` resolves without an import. Both `pnpm typecheck` and the fresh-project registry-check `tsc --noEmit` pass, and shipped components (e.g. spinner, skeleton, sonner) use bare `React.*` the same way. Resolving as a false positive.

apps/apollo-vertex/registry/solution-tests/kpi-bar.tsx
```diff
@@ -0,0 +1,181 @@
+"use client";
+
+import { useTranslation } from "react-i18next";
```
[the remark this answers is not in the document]
frankkluijtmans: Not an issue in this repo. `@types/react` declares `export as namespace React`, so `React` is a global type and `React.FC` resolves without an import. Both `pnpm typecheck` and the fresh-project registry-check `tsc --noEmit` pass, and shipped components (e.g. spinner, skeleton, sonner) use bare `React.*` the same way. Resolving as a false positive.

apps/apollo-vertex/registry/solution-tests/result-expanded-content.tsx
```diff
@@ -0,0 +1,147 @@
+"use client";
+
+import { useTranslation } from "react-i18next";
```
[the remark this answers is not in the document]
frankkluijtmans: Not an issue in this repo. `@types/react` declares `export as namespace React`, so `React` is a global type and `React.FC` resolves without an import. Both `pnpm typecheck` and the fresh-project registry-check `tsc --noEmit` pass, and shipped components (e.g. spinner, skeleton, sonner) use bare `React.*` the same way. Resolving as a false positive.

apps/apollo-vertex/registry/solution-tests/run-details-dialog.tsx
```diff
@@ -0,0 +1,507 @@
+"use client";
+
+import { useState, useCallback } from "react";
```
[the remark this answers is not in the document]
frankkluijtmans: Not an issue in this repo. `@types/react` declares `export as namespace React`, so `React` is a global type and `React.FC` resolves without an import. Both `pnpm typecheck` and the fresh-project registry-check `tsc --noEmit` pass, and shipped components (e.g. spinner, skeleton, sonner) use bare `React.*` the same way. Resolving as a false positive.

apps/apollo-vertex/registry/solution-tests/solution-tests-view.tsx
```diff
@@ -0,0 +1,518 @@
+"use client";
+
+import { useState } from "react";
```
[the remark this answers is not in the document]
frankkluijtmans: Not an issue in this repo. `@types/react` declares `export as namespace React`, so `React` is a global type and `React.FC` resolves without an import. Both `pnpm typecheck` and the fresh-project registry-check `tsc --noEmit` pass, and shipped components (e.g. spinner, skeleton, sonner) use bare `React.*` the same way. Resolving as a false positive.

apps/apollo-vertex/templates/SolutionTestsTemplate.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+  return (
+    <LocaleProvider>
+      {/* Presentational demo: writes are no-ops, so trigger deps are stubs. */}
+      <SolutionTestsProvider
+        config={config}
+        triggerBaseUrl=""
+        getToken={() => null}
+        getEntityId={() => ""}
+      >
```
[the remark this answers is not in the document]
frankkluijtmans: Leaving as `() => ""`. Both attachment resolvers guard `if (!entityId) return Promise.resolve(null)`, and `""` is falsy, so no fetch with an invalid GUID can occur. Returning `undefined` would trip the repo's `eslint/no-undefined` rule. Resolving.

apps/apollo-vertex/app/templates/solution-tests/page.mdx
```diff
@@ -0,0 +1,179 @@
...
+render the dumb `SolutionTestsView`, which takes all data and callbacks via
+props (this is what the preview above does).
+
+## Installation
+
+```bash
+npx shadcn@latest add @uipath/solution-tests
```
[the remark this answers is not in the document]
frankkluijtmans: Leaving `shadcn@latest` here. It's the convention across ~80 apollo-vertex docs pages, so pinning only this one would be inconsistent. If reproducible install docs are wanted, that's better done repo-wide as a follow-up. Resolving.

apps/apollo-vertex/registry/solution-tests/use-run-results.ts
```diff
@@ -94,10 +94,9 @@ export function useResultAttachment(): AttachmentFetcher<
   [resultId: string, field: ResultAttachmentField]
 > {
   const solution = useSolution();
-  const getEntityId = useSolutionTestsEntityId();
   return {
     fetch: (resultId: string, field: ResultAttachmentField) => {
-      const entityId = getEntityId(ENTITY.runResults);
+      const entityId = solution?.api.entityIds[ENTITY.runResults];
       if (!entityId) return Promise.resolve(null);
       return fetchAttachment(solution, entityId, resultId, field);
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: added optional chaining on `entityIds` (`solution?.api.entityIds?.[ENTITY.runResults]`), keeping the falsy guard, so an older vs-core without `entityIds` degrades gracefully instead of throwing. (c2c81ed9)

apps/apollo-vertex/registry/solution-tests/use-baseline-jobs.ts
```diff
@@ -51,10 +51,9 @@ export function useRemoveJobBaseline(): MutationHookResult<
 /** Expected-output attachment for a baseline job (Agents expansion). */
 export function useJobExpectedOutput(): AttachmentFetcher<[jobId: string]> {
   const solution = useSolution();
-  const getEntityId = useSolutionTestsEntityId();
   return {
     fetch: (jobId: string) => {
-      const entityId = getEntityId(ENTITY.jobs);
+      const entityId = solution?.api.entityIds[ENTITY.jobs];
       if (!entityId) return Promise.resolve(null);
       // The baseline job's expected output lives in the `Output` File field on
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed the same way: `solution?.api.entityIds?.[ENTITY.jobs]`. (c2c81ed9)

apps/apollo-vertex/app/templates/solution-tests/page.mdx
```diff
@@ -0,0 +1,175 @@
...
+```
+
+> **Requires [`@uipath/vs-core`](https://www.npmjs.com/package/@uipath/vs-core)
+> `^2.0.4`.** The container reads the Solution Test collections from the
+> dedicated, name-keyed `solution.api.collections.solutionTests` namespace
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: the Installation callout now requires `@uipath/vs-core@^2.0.6` (the version that added `entityIds`), matching the prose and the registry dep. (c2c81ed9)

apps/apollo-vertex/registry/solution-tests/json-viewer-dialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+  <Dialog open={open} onOpenChange={(isOpen) => !isOpen && onClose()}>
+    <DialogContent
+      className="max-h-[90vh] overflow-hidden"
+      style={{ maxWidth: "75vw", width: "fit-content", minWidth: "30vw" }}
+    >
+      <DialogHeader>
+        <DialogTitle>{title}</DialogTitle>
+        <DialogDescription className="sr-only">
+          {"JSON data viewer"}
+        </DialogDescription>
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: the SR-only description now uses `t("json_viewer_description")` (added `useTranslation` to this component and the key to `en.json`). (c2c81ed9)

apps/apollo-vertex/registry/solution-tests/run-details-dialog-view.tsx
```diff
@@ -0,0 +1,351 @@
...
+                ? t("subject_run_results", {
+                    subject: config.subjectNoun.singular,
+                    id: subjectId,
+                  })
+                : subjectId
+            } — ${t("run_results")}`}
+          </DialogTitle>
+          <DialogDescription className="sr-only">
+            {"Agent-level results for this test run"}
+          </DialogDescription>
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: now `t("run_results_description")`, with the key added to `en.json`. (c2c81ed9)

apps/apollo-vertex/registry/solution-tests/context.tsx
```diff
@@ -29,25 +25,21 @@ interface SolutionTestsProviderProps {
   /** Base URL each action slug is appended to (no trailing slash). */
   triggerBaseUrl: string;
   getToken: () => Promise<string | null> | string | null;
-  /** Resolves a Solution Test entity name to its DataFabric id (GUID); used for attachment reads. */
-  getEntityId: EntityIdResolver;
   children: ReactNode;
```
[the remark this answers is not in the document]
frankkluijtmans: Already addressed: the PR description was rewritten when `getEntityId` was removed. It no longer documents it as a prop, and the snippet/example drop it; it's only mentioned in a Dependency note explaining that `solution.api.entityIds` replaced it. Resolving.

apps/apollo-vertex/registry/solution-tests/result-expanded-content.tsx
```diff
@@ -0,0 +1,151 @@
...
+          <h4 className="mb-2 text-sm font-semibold">
+            {t("output_evaluator_results")}
+          </h4>
+          <EvaluatorResultsView data={data.evaluatorResults} />
+        </div>
+      )}
+
+      {isPassedOrFailed && !isAutoPass && (
+        <div className="flex flex-col gap-4">
+          {(data.expectedInput != null || data.actualInput != null) && (
```
RobinMennens-UiPath: The Expected/Actual **Input** panels are now rendered whenever input data is present. In the original these were gated behind a `debug` flag (`useConfiguration("debug")`), so raw inputs were only visible in debug mode.

Marking this as up for discussion rather than a hard change: we got feedback that looking at the input isn't always relevant, which is why it was hidden behind the debug gate in the first place. So restoring the gate has real value. That said, I acknowledge it's trickier to wire up the same logic generically here (there's no `useConfiguration` equivalent in this component). Let's decide whether to re-introduce a config-driven flag for this or intentionally always show the inputs.

apps/apollo-vertex/registry/solution-tests/solution-tests.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+function getRunsForBatch(
+  batchId: string,
+  allRuns: ReturnType<typeof useSolutionTestRuns>["runs"],
+) {
+  return allRuns.filter((r) => r.RunBatchId === batchId);
+}
+
+/** Runs an action callable, surfacing any thrown error as a toast. */
+function runAction(action: () => unknown, fallback: string): void {
```
RobinMennens-UiPath: Success toasts for the primary actions (run, run all, enable/disable, delete, force-stop) are no longer shown. The original surfaced `test_triggered` / `all_tests_triggered` / `test_enabled|disabled` / `test_deleted` / `force_stop_initiated` on success; only failures toast now. With live collections the table updates on its own, so this may be intentional, but please confirm.

Relatedly, the specific "batch already running" message (`SolutionTest_BatchAlreadyRunning`) is gone, so a concurrent run now produces a generic error instead. Lower priority since the Run buttons are disabled while runs are active.

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
```diff
@@ -0,0 +1,59 @@
...
+  if (entries.length === 0) return null;
+
+  return (
+    <div className="flex flex-col gap-3">
+      {entries.map(([evaluatorId, result]) => {
+        const score = result?.score;
+        const scoreStr = score == null ? "—" : `${Math.round(score * 100)}%`;
+        const scoreColor =
+          score == null
+            ? "text-muted-foreground"
```
RobinMennens-UiPath: A null evaluator score now renders in `text-muted-foreground`; the original colored a null score red (the old ternary fell through to `text-red-500`). Likely an improvement, just flagging the visual change to confirm it's intentional.

apps/apollo-vertex/registry/solution-tests/kpi-bar.tsx
```diff
@@ -0,0 +1,188 @@
...
+        date: date.toLocaleDateString([], {
+          month: "short",
+          day: "numeric",
+        }),
+        timestamp: date.getTime(),
+        score: b.OverallScore == null ? 0 : Math.round(b.OverallScore * 100),
+      };
+    })
+    .toSorted((a, b) => a.timestamp - b.timestamp)
+    .slice(-10);
```
RobinMennens-UiPath: `.slice(-10)` caps the trend chart to the last 10 batches. Consider a named constant (e.g. `MAX_TREND_POINTS`) to document intent.

apps/apollo-vertex/registry/solution-tests/expanded-run-tests.tsx
```diff
@@ -0,0 +1,61 @@
...
+      .mutate(runId)
+      .catch(() => toast.error(t("failed_to_force_stop_run")))
+      .finally(() => setStoppingRunId(null));
+  };
+
+  return (
+    <ExpandedRunTestsView
+      batchRuns={batchRuns}
+      tests={tests}
+      openingRunId={null}
```
RobinMennens-UiPath: `openingRunId={null}` is hardcoded, so the "opening details" spinner branch in `expanded-run-tests-view.tsx` can never render in the app (only the template sets it, also to null). If a loading state for opening the details dialog was intended, it needs to be tracked like `stoppingRunId`; otherwise consider dropping the prop.

RobinMennens-UiPath (changes requested):
Solid move of the FINS Solution Tests dashboard into a generic, config-driven registry component. Genericness checks out (all vertical specifics flow through `SolutionTestsConfig`), and the new run-confirmation flow is wired correctly. A few behaviors changed during the move that are worth confirming or addressing. Details inline.

apps/apollo-vertex/registry/solution-tests/result-expanded-content.tsx
frankkluijtmans: Restored as a config-driven gate. Added `showInputs?: boolean` to `SolutionTestsConfig` (default `false`), so the Expected/Actual input panels are hidden by default (matching the old debug gate) and a vertical opts in via `config.showInputs`. (9cf79fc6)

apps/apollo-vertex/registry/solution-tests/solution-tests.tsx
frankkluijtmans: Restored success toasts for run, run all, delete, and enable/disable (`test_triggered` / `all_tests_triggered` / `test_deleted` / `test_enabled`|`test_disabled`) via a `success` arg on the action helpers. Left force-stop's `force_stop_initiated` and the specific `SolutionTest_BatchAlreadyRunning` message out for now (lower priority; Run is disabled during active runs and the generic error still surfaces) — happy to add those too if you'd like. (9cf79fc6)

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
frankkluijtmans: Intentional — keeping the muted color. A null/absent score means "no data", not a failure, so red would misrepresent it. (9cf79fc6)

apps/apollo-vertex/registry/solution-tests/kpi-bar.tsx
frankkluijtmans: Done — extracted `MAX_TREND_POINTS = 10` in `constants.ts` and use `slice(-MAX_TREND_POINTS)`. (9cf79fc6)

apps/apollo-vertex/registry/solution-tests/expanded-run-tests.tsx
frankkluijtmans: Dropped the prop. The Details button opens the dialog immediately and the dialog shows its own loading skeleton (`useRunResults` fetches inside it), so a button-level opening spinner was redundant. Removed `openingRunId` from the view props + the spinner branch, and the `={null}` passthroughs. (9cf79fc6)

apps/apollo-vertex/registry/solution-tests/expanded-agents.tsx
```diff
@@ -0,0 +1,50 @@
...
+    loading: boolean;
+  } | null>(null);
+
+  const handleViewExpected = useCallback(
+    async (job: SolutionTestJob) => {
+      setViewing({ job, data: null, loading: true });
+      try {
+        const data = await expectedOutput.fetch(job.Id);
+        setViewedJobs((prev) => new Map(prev).set(job.Id, true));
+        setViewing((prev) => (prev ? { ...prev, data, loading: false } : null));
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: `handleViewExpected` now treats a `null`/`undefined` fetch result as "no expected output" — it sets `viewedJobs` to `false` and closes the viewer instead of marking it viewed and opening with null data. (7690eafa)

apps/apollo-vertex/registry/solution-tests/expanded-run-tests.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+interface DetailsTarget {
+  run: SolutionTestRun;
+  subjectId: string;
+}
+
+interface ExpandedRunTestsProps {
+  batchRuns: SolutionTestRun[];
+  tests: SolutionTest[];
+}
```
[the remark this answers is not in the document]
frankkluijtmans: Renamed the misleading prop: `ExpandedRunTests` / `ExpandedRunTestsView` now take `runs: SolutionTestRun[]` (and the loop var is `run`), with call sites updated. `batchRuns` is now used only for actual `SolutionTestBatchRun[]`. (7690eafa)

apps/apollo-vertex/registry/solution-tests/kpi-bar.tsx
```diff
@@ -0,0 +1,188 @@
...
+  batchRuns: SolutionTestBatchRun[];
+  loading?: boolean;
+}
+
+export const KpiBar: React.FC<KpiBarProps> = ({
+  batchRuns,
+  loading = false,
+}) => {
+  const { t } = useTranslation();
+  const { subjectNoun, passThreshold } = useSolutionTestsConfig();
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: moved `scoreTrendChartConfig` inside `KpiBar` so the label uses `t("score")` instead of the hard-coded English string. (7690eafa)

RobinMennens-UiPath (approved):
All review comments addressed, verified against the latest commit:

- result-expanded-content: input panels restored behind a generic `config.showInputs` gate (default off) — clean solution that keeps the original debug-gate behavior while staying vertical-agnostic.
- solution-tests: success toasts restored for run / run all / delete / toggle.
- kpi-bar: `MAX_TREND_POINTS` constant extracted.
- expanded-run-tests: unused `openingRunId` prop removed.
- evaluator null-score muted color: agreed, intentional.

Genericness and the move from the FINS Loan Setup dashboard both check out. LGTM 🚀

apps/apollo-vertex/registry/solution-tests/solution-tests-view.tsx
```diff
@@ -0,0 +1,584 @@
...
+  setRunConfirm: (target: RunConfirmTarget | null) => void;
+  /** Confirms the pending run (the run-confirmation dialog's primary action). */
+  onConfirmRun: () => void;
+  onToggleActive: (testId: string, newValue: boolean) => void;
+  onDeleteTest: (testId: string) => void;
+  onForceStopBatch: (batchId: string) => void;
+  /** Renders the expanded content for a test case row (baseline agents). */
+  renderExpandedTest?: (test: SolutionTest) => ReactNode;
+  /** Renders the expanded content for a batch-run row (per-run results). */
+  renderExpandedRun?: (batch: SolutionTestBatchRun) => ReactNode;
```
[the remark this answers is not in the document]
frankkluijtmans: Agreed, made them required. `renderExpandedTest` and `renderExpandedRun` are no longer optional on `SolutionTestsViewProps`, and the now-unnecessary `?.` at the call sites was dropped. Both callers (the smart container and the demo template) already pass them, so no behavior change — it just prevents a consumer from getting expand affordances that open into empty rows. (ad3cf3e3)

RobinMennens-UiPath (approved):

apps/apollo-vertex/registry/solution-tests/solution-tests.tsx
```diff
@@ -0,0 +1,199 @@
...
+    setPendingId(id);
+    action()
+      .then(() => {
+        if (success) toast.success(success);
+      })
+      .catch((err: unknown) =>
+        toast.error(err instanceof Error ? err.message : fallback),
+      )
+      .finally(() => setPendingId(null));
+  };
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: rewrote `useIdAction`'s runner with async/await — `await action()` catches both a rejected promise and a synchronous throw, and `pendingId` is cleared in a single `finally`, so the row can't get stuck pending. (4d64ec63)

apps/apollo-vertex/registry/solution-tests/run-details-dialog-view.tsx
```diff
@@ -0,0 +1,351 @@
...
+    if (inBaseline) return { label: "✓", className: "" };
+    if (hadBaseline)
+      return { label: t("removed"), className: "text-red-500 font-medium" };
+    return { label: "—", className: "" };
+  })();
+
+  return (
+    <>
+      <TableRow className="cursor-pointer hover:bg-muted/50" onClick={onToggle}>
+        <TableCell className="px-3 py-2">
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: the chevron in the first cell is now a real `Button` (ghost, icon-sm) with `aria-label` (expand/collapse) and `aria-expanded`, matching the main tables' expand control. It's keyboard-focusable and responds to Enter/Space; the row's click handler stays for mouse users (the button stops propagation to avoid a double toggle). (4d64ec63)

apps/apollo-vertex/registry/solution-tests/expanded-run-tests-view.tsx
```diff
@@ -0,0 +1,130 @@
...
+              <TableHead className="px-3 py-2">{t("status")}</TableHead>
+              <TableHead className="w-6 px-1 py-2" />
+              <TableHead className="px-3 py-2">{t("actions")}</TableHead>
+            </TableRow>
+          </TableHeader>
+          <TableBody>
+            {runs.map((run) => {
+              const test = tests.find((x) => x.Id === run.SolutionTestId);
+              const subjectId = test?.SubjectId ?? run.SolutionTestId;
+
```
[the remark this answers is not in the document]
frankkluijtmans: Keeping `tests.find` here — the run/test lists in this dialog are small (the runs within a single batch), so the O(runs×tests) cost is negligible and `find` reads more clearly than threading a Map through. Can revisit if these lists ever grow.

apps/apollo-vertex/registry/solution-tests/delete-confirm-dialog.tsx
```diff
@@ -0,0 +1,50 @@
...
+export const DeleteConfirmDialog: React.FC<DeleteConfirmDialogProps> = ({
+  open,
+  deleting,
+  onConfirm,
+  onCancel,
+}) => {
+  const { t } = useTranslation();
+
+  return (
+    <Dialog open={open} onOpenChange={(isOpen) => !isOpen && onCancel()}>
```
0xr3ngar: why is onCancel() here in the boolean?! `onOpenChange={(isOpen) => !isOpen && onCancel()}`

apps/apollo-vertex/registry/solution-tests/delete-confirm-dialog.tsx
```diff
@@ -0,0 +1,50 @@
...
+  Dialog,
+  DialogContent,
+  DialogHeader,
+  DialogTitle,
+  DialogDescription,
+} from "@/components/ui/dialog";
+
+interface DeleteConfirmDialogProps {
+  open: boolean;
+  deleting: boolean;
```
0xr3ngar, line 16: isOpen,
isDeleting

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
```diff
@@ -0,0 +1,59 @@
...
+    try {
+      parsed = JSON.parse(parsed) as unknown;
+    } catch {
+      return null;
+    }
+  }
+  if (typeof parsed !== "object" || parsed === null) return null;
+  // oxlint-disable-next-line typescript-eslint(no-unsafe-type-assertion) -- parsed evaluator JSON is trusted to match the EvaluatorResult shape
+  const entries = Object.entries(parsed) as [string, EvaluatorResult][];
+  if (entries.length === 0) return null;
```
0xr3ngar: why not use zod instead of this cursed code?

just Zod schema for `z.record(z.string,EvaluatorResult)`

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
```diff
@@ -0,0 +1,59 @@
...
+    <div className="flex flex-col gap-3">
+      {entries.map(([evaluatorId, result]) => {
+        const score = result?.score;
+        const scoreStr = score == null ? "—" : `${Math.round(score * 100)}%`;
+        const scoreColor =
+          score == null
+            ? "text-muted-foreground"
+            : score >= passThreshold
+              ? "text-green-600"
+              : "text-red-500";
```
0xr3ngar: nested ternary

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
```diff
@@ -0,0 +1,59 @@
+"use client";
+
+import { EVALUATOR_LABELS } from "./constants";
+import { useSolutionTestsConfig } from "./context";
+
+interface EvaluatorResult {
+  score?: number;
+  details?: { justification?: string };
+}
```
0xr3ngar: how can a score be undefined or number? or details undefined ?? this is super confusing

Either score and details is there or it isnt

apps/apollo-vertex/registry/solution-tests/expanded-agents-view.tsx
```diff
@@ -0,0 +1,141 @@
...
+  baselines: SolutionTestJob[];
+  isLoading: boolean;
+  /** Jobs whose expected-output fetch resolved (true) or had none (false). */
+  viewedJobs: Map<string, boolean>;
+  viewing: { job: SolutionTestJob; data: unknown; loading: boolean } | null;
+  onViewExpected: (job: SolutionTestJob) => void;
+  onCloseViewer: () => void;
+}
+
+export const ExpandedAgentsView: React.FC<ExpandedAgentsViewProps> = ({
```
0xr3ngar: React.FC? why is this needed? it's inferred if you just do const Component = ({something}: Interface) => {}...

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export const EvaluatorResultsView: React.FC<{ data: unknown }> = ({ data }) => {
```
0xr3ngar: React.FC? why is this needed? it's inferred if you just do const Component = ({something}: Interface) => {}...

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
```diff
@@ -0,0 +1,59 @@
...
+  if (typeof parsed !== "object" || parsed === null) return null;
+  // oxlint-disable-next-line typescript-eslint(no-unsafe-type-assertion) -- parsed evaluator JSON is trusted to match the EvaluatorResult shape
+  const entries = Object.entries(parsed) as [string, EvaluatorResult][];
+  if (entries.length === 0) return null;
+
+  return (
+    <div className="flex flex-col gap-3">
+      {entries.map(([evaluatorId, result]) => {
+        const score = result?.score;
+        const scoreStr = score == null ? "—" : `${Math.round(score * 100)}%`;
```
0xr3ngar: why the - ? if there is no score it should say it clearly

apps/apollo-vertex/registry/solution-tests/expanded-agents-view.tsx
```diff
@@ -0,0 +1,141 @@
...
+import type { SolutionTest, SolutionTestJob } from "./types";
+import { JsonViewerDialog } from "./json-viewer-dialog";
+
+export interface ExpandedAgentsViewProps {
+  test: SolutionTest;
+  baselines: SolutionTestJob[];
+  isLoading: boolean;
+  /** Jobs whose expected-output fetch resolved (true) or had none (false). */
+  viewedJobs: Map<string, boolean>;
+  viewing: { job: SolutionTestJob; data: unknown; loading: boolean } | null;
```
0xr3ngar: viewedJobs ? why is this map needed here? just pass hasOutput from outside, no need for this map to be here just for a `const hasNoOutput = viewedJobs.get(job.Id) === false;`

apps/apollo-vertex/registry/solution-tests/expanded-agents-view.tsx
```diff
@@ -0,0 +1,141 @@
...
+            </TableRow>
+          </TableHeader>
+          <TableBody>
+            {baselines.map((job) => {
+              const hasNoOutput = viewedJobs.get(job.Id) === false;
+              return (
+                <TableRow key={job.Id}>
+                  <TableCell className="px-3 py-2">{job.ProcessName}</TableCell>
+                  <TableCell className="px-3 py-2 text-muted-foreground">
+                    {job.ProcessVersion ?? "-"}
```
0xr3ngar, line 102: again what is `"-"`

apps/apollo-vertex/registry/solution-tests/expanded-agents-view.tsx
```diff
@@ -0,0 +1,141 @@
...
+            })}
+          </TableBody>
+        </Table>
+      </div>
+
+      {viewing && (
+        <JsonViewerDialog
+          open
+          onClose={onCloseViewer}
+          title={`${t("expected_output")} - ${viewing.job.ProcessName}`}
```
0xr3ngar: make this a translation key - expected_output_{{name}}

apps/apollo-vertex/registry/solution-tests/expanded-agents-view.tsx
```diff
@@ -0,0 +1,141 @@
...
+                </TableRow>
+              );
+            })}
+          </TableBody>
+        </Table>
+      </div>
+
+      {viewing && (
+        <JsonViewerDialog
+          open
```
0xr3ngar, line 132: why hardcode open here?

apps/apollo-vertex/registry/solution-tests/expanded-agents.tsx
```diff
@@ -0,0 +1,57 @@
...
+import type { SolutionTest, SolutionTestJob } from "./types";
+import { useBaselineJobs, useJobExpectedOutput } from "./hooks";
+import { ExpandedAgentsView } from "./expanded-agents-view";
+
+interface ExpandedAgentsProps {
+  test: SolutionTest;
+}
+
+/** Smart wrapper: fetches baseline jobs (lazily, on expand) + expected output. */
+export const ExpandedAgents: React.FC<ExpandedAgentsProps> = ({ test }) => {
```
0xr3ngar: React.FC? why is this needed? it's inferred if you just do const Component = ({something}: Interface) => {}...

apps/apollo-vertex/registry/solution-tests/expanded-agents.tsx
```diff
@@ -0,0 +1,57 @@
...
+        }
+        setViewedJobs((prev) => new Map(prev).set(job.Id, true));
+        setViewing((prev) => (prev ? { ...prev, data, loading: false } : null));
+      } catch {
+        setViewedJobs((prev) => new Map(prev).set(job.Id, false));
+        setViewing(null);
+      }
+    },
+    [expectedOutput],
+  );
```
0xr3ngar: why all the manual fetching? just use react query for this, no?

apps/apollo-vertex/registry/solution-tests/expanded-agents.tsx
```diff
@@ -0,0 +1,57 @@
...
+export const ExpandedAgents: React.FC<ExpandedAgentsProps> = ({ test }) => {
+  const { jobs: baselines, isLoading } = useBaselineJobs(test.Id);
+  const expectedOutput = useJobExpectedOutput();
+
+  const [viewedJobs, setViewedJobs] = useState<Map<string, boolean>>(new Map());
+  const [viewing, setViewing] = useState<{
+    job: SolutionTestJob;
+    data: unknown;
+    loading: boolean;
+  } | null>(null);
```
0xr3ngar: this state feels super shady man, I think we can improve this state and infer

0xr3ngar (changes requested):
@frankkluijtmans this PR is too big for me to properly review and say . A 3.4k line addition is just not something I can give a meaningful review on.

Can you please split this up into smaller PRs that make sense on their own? I thought we had discussed this with the design team as well, since PRs like this are really hard to review.

I already spent about 30 minutes trying to work through it, but Im going to have to say no to reviewing it in this form. Please break it down into smaller pieces.

apps/apollo-vertex/registry/solution-tests/delete-confirm-dialog.tsx
frankkluijtmans: I will change this, but as long as we don't enforce this I don't think it's something we should worry about too much.

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
frankkluijtmans: The dash is intentional for now and mirrors the current implementation. If we get feedback about it I'll update it.

apps/apollo-vertex/registry/solution-tests/expanded-agents-view.tsx
frankkluijtmans: The dash is intentional for now and mirrors the current implementation. If we get feedback about it I'll update it.

0xr3ngar (approved):

apps/apollo-vertex/registry/solution-tests/delete-confirm-dialog.tsx
frankkluijtmans: Rewritten for clarity: `onOpenChange={(isOpen) => { if (!isOpen) onCancel(); }}` (applied to all four dialogs for consistency). (aab86595)

apps/apollo-vertex/registry/solution-tests/delete-confirm-dialog.tsx
frankkluijtmans: Kept the prop as `open` — that's the shadcn `Dialog` convention (the underlying `Dialog` takes `open`), and all four solution-tests dialogs use it. Renaming to `isOpen` would diverge from that. Happy to rename `deleting`→`isDeleting` if that's what you meant — let me know.

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
frankkluijtmans: Done — replaced the `as [string, EvaluatorResult][]` cast with a zod schema: `z.record(z.string(), EvaluatorResultSchema)` + `safeParse`, consistent with user-messages.ts. (aab86595)

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
frankkluijtmans: Done — extracted the color logic into a `scoreColorClass(score, passThreshold)` helper, so no nested ternary in the JSX. (aab86595)

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
frankkluijtmans: Now validated by zod. Kept `score`/`details` optional because the upstream evaluator JSON genuinely omits them sometimes, and we're keeping the "—" no-score rendering (per the placeholder convention we use everywhere). zod makes the parse safe even when fields are missing. (aab86595)

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
frankkluijtmans: Done — dropped `React.FC`, now `const EvaluatorResultsView = ({ data }: { data: unknown }) => …`. (aab86595)

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
frankkluijtmans: Keeping `—` here — it's the empty-value placeholder convention used across the whole feature; we're standardizing on it rather than per-spot prose. (aab86595)

apps/apollo-vertex/registry/solution-tests/expanded-agents-view.tsx
frankkluijtmans: Done — dropped `React.FC` for a typed arrow. (aab86595)

apps/apollo-vertex/registry/solution-tests/expanded-agents-view.tsx
frankkluijtmans: Reworked: the fetch now uses React Query and the tri-state `Map<string, boolean>` is gone — the view takes a `noOutputJobIds: Set<string>` instead (clearer: a job is in the set iff its expected-output fetch came back empty). (aab86595)

apps/apollo-vertex/registry/solution-tests/expanded-agents-view.tsx
frankkluijtmans: Keeping `-` — same empty-value placeholder convention used across the feature. (aab86595)

apps/apollo-vertex/registry/solution-tests/expanded-agents-view.tsx
frankkluijtmans: Done — now a translation key with interpolation: `t("expected_output_for", { name })` (`"Expected output: {{name}}"`). (aab86595)

apps/apollo-vertex/registry/solution-tests/expanded-agents-view.tsx
frankkluijtmans: It's intentional: the dialog only mounts inside `{viewing && (…)}`, so `open` is always true when rendered. `open` (shorthand) is correct there. (aab86595)

apps/apollo-vertex/registry/solution-tests/expanded-agents.tsx
frankkluijtmans: Done — dropped `React.FC` for a typed arrow. (aab86595)

apps/apollo-vertex/registry/solution-tests/expanded-agents.tsx
frankkluijtmans: Done — the expected-output fetch is now a React Query `useQuery` keyed by the open job (enabled on demand), instead of the manual async handler. (aab86595)

apps/apollo-vertex/registry/solution-tests/expanded-agents.tsx
frankkluijtmans: Cleaned up: the state is now just `openJob` + a `noOutputJobIds` Set, with the fetch/loading driven by React Query (`data`/`isFetching`). No more manual `viewing` object juggling. (aab86595)

frankkluijtmans:
> @frankkluijtmans this PR is too big for me to properly review and say . A 3.4k line addition is just not something I can give a meaningful review on.
> 
> Can you please split this up into smaller PRs that make sense on their own? I thought we had discussed this with the design team as well, since PRs like this are really hard to review.
> 
> I already spent about 30 minutes trying to work through it, but Im going to have to say no to reviewing it in this form. Please break it down into smaller pieces.

As discussed, you're correct to call this out. Even though this was moving existing code, there we're some issues that should be solved when extracting it. I do believe I tackled those ones and did a smoke test by integrating it in loan setup. In the future I'll do a better split than just data layer + components. Thanks for reviewing!

apps/apollo-vertex/registry/solution-tests/expanded-agents.tsx
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  // The open job has no expected output once its fetch resolves empty.
+  const noOutputJobIds = new Set(
+    openJob && isFetched && data == null ? [openJob.Id] : [],
+  );
+
+  const viewing = openJob
+    ? { job: openJob, data: data ?? null, loading: isFetching }
+    : null;
```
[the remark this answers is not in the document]
frankkluijtmans: Intentionally not persisting this. We simplified `ExpandedAgents` on purpose to only fetch the currently-open job (no kept-mounted queries / no `noOutputJobIds` set across openings) — the maintainer opted for the simpler model where a row's "no output" state isn't retained after the viewer closes. Re-adding the persistence is the exact thing we just removed, so leaving as-is by design.

apps/apollo-vertex/registry/solution-tests/solution-tests-view.tsx
```diff
@@ -0,0 +1,584 @@
...
+      cell: ({ row }) => (
+        <span className="text-sm">
+          {renderValueOrEmptyState(
+            row.original.StartedAt ?? row.original.CreateTime,
+            { type: "datetime" },
+          )}
+        </span>
+      ),
+      enableSorting: true,
+    },
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: the Run date column now uses `accessorFn: (row) => row.StartedAt ?? row.CreateTime` (with `id: "StartedAt"` to keep the default sort), so sorting matches the rendered fallback. (4d16f0dc)

apps/apollo-vertex/templates/SolutionTestsTemplate.tsx
```diff
@@ -0,0 +1,264 @@
...
+                runs={getRunsForBatch(batch.Id, db.runs)}
+                tests={db.tests}
+                stoppingRunId={null}
+                onOpenDetails={(run, subjectId) =>
+                  setDetailsRun({ run, subjectId })
+                }
+                onForceStop={noop}
+                detailsDialog={
+                  detailsRun && (
+                    <RunDetailsDialogView
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: the demo's details dialog is now gated to the owning batch via `detailsRun?.run.RunBatchId === batch.Id`, so it only renders under the batch that owns the open run. (4d16f0dc)

apps/apollo-vertex/app/templates/solution-tests/page.mdx
```diff
@@ -0,0 +1,178 @@
...
+   collection's optimistic update and needs no extra wiring.
+3. **Presentation** — pass `config.subjectColumns` (and optionally
+   `getSubjectHref` / `subjectNoun`) to flavor the table for your vertical.
+
+Everything else (evaluator labels, status labels, the poll interval) is
+hard-coded in `constants.ts`. Edit there to retarget a deployment. The pass
+threshold defaults to `0.9` in `constants.ts` but is overridable per vertical
+via `config.passThreshold`. Consumers who want to supply their own data plumbing can instead
+render the dumb `SolutionTestsView`, which takes all data and callbacks via
+props (this is what the preview above does).
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed the docs: clarified that the dumb `SolutionTestsView` must still be wrapped in `SolutionTestsProvider` because it reads `config` via `useSolutionTestsConfig`. (4d16f0dc)

0xr3ngar (approved):

PR #784, opened 2026-06-05
feat(apollo-wind): add PromptEditor component [MST-10659]

fikewa-olatunji:
**Jira:** [MST-10659](https://uipath.atlassian.net/browse/MST-10659)

## Summary

Ports the Lexical-based **PromptEditor** from flow-workbench into `@uipath/apollo-wind` as a reusable, fully prop-driven component. It renders inline variable token pills (input / output / state / resource), a formatting toolbar, a markdown preview mode, and a `$`-triggered variable autocomplete.

The port deliberately decouples the editor from flow-workbench-specific systems that don't exist in a design system:
- **Autocomplete & token tooltip** rebuilt on apollo-wind primitives (`Command` / `Popover` / `Tooltip`), driven by the `autoCompleteOptions` prop — no flow `VariablePicker` or schema services.
- **i18n** stripped to plain English literals (no `react-i18next`).
- **Theme** detection dropped; token colors map to apollo-wind design tokens (theme-adaptive via CSS).
- React-Flow keyboard shielding dropped.
- Variable drag-and-drop is supported via a decoupled native HTML5 `VariableDropPlugin` + `mapVarDropToToken` prop (the consumer owns the drag source and sets the path on `dataTransfer` under the exported `VARIABLE_DRAG_MIME`) — no dnd-kit dependency, unlike flow-workbench's version.

## Public API

`PromptEditor` + types (`PromptEditorProps`, `PromptEditorRef`, `PromptEditorToken`, `PromptEditorTokenType`, `PromptEditorAutoCompleteOption`, `PromptEditorMode`) exported from the package root. Controlled (`value`/`onChange`) and uncontrolled (`initialValue`) usage, `multiline`, `minRows`/`maxRows`, `placeholder`, `disabled`, `showToolbar`, `mode`/`onModeChange`, `borderless`, `fillHeight`, and an imperative `editorRef`.

## Dependencies

**apollo-wind:** `lexical`, `@lexical/react`, `@lexical/utils`, `@lexical/clipboard` (0.42.0), `marked`, `dompurify`.

**apollo-react:** its existing `lexical` + `@lexical/*` packages are bumped 0.16.0 → 0.42.0 to align with apollo-wind so the repo-wide dependency-version-consistency check passes without an exemption (the only resulting code change is the `LexicalErrorBoundary` named-import fix). No new audit findings.

## Robustness

Token/option props are normalized to arrays so malformed input (e.g. a Storybook "Set object" `{}`) can't crash the editor or preview. `maxRows` caps the visible height and scrolls (clamped so it isn't overridden when `maxRows ≤ minRows`).

## Testing

- `vitest`: 19 unit tests for PromptEditor (rendering, toolbar, preview, mode toggle, autocomplete mount, malformed-input tolerance) — full apollo-wind suite passes (928 tests).
- Storybook stories: Default, SingleLine, WithTokens, WithToolbar, WithAutocomplete, Preview, Disabled, Borderless, Controlled.
- Verified locally: `biome format`/`lint` clean, `test:coverage` passes, full `pnpm build` (all 8 packages) succeeds, frozen lockfile in sync.
- Interactive flows (typing, `$`-autocomplete, chip insertion, light/dark, height capping) verified end-to-end in a headless browser.

[MST-10659]: https://uipath.atlassian.net/browse/MST-10659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/apollo-wind/.storybook/preview-head.html
```diff
@@ -22,4 +22,11 @@
     width: 100%;
     height: 100%;
   }
+
+  /* Theme-match autodocs story previews. By default Storybook renders each story block on the white
+     docs page, which leaves transparent components (e.g. the borderless PromptEditor) white-on-white
+     under a dark theme. Painting the block with the active theme surface keeps them readable. */
+  .sb-story {
+    background-color: var(--color-background);
+  }
```
CalinaCristian: Fixed this on a different PR -- can you check if it works without this change? (after rebase)

packages/apollo-wind/src/components/ui/prompt-editor/components/MarkdownPreview.tsx
```diff
@@ -0,0 +1,200 @@
...
+.prompt-editor-preview a:hover { opacity: 0.8; }
+.prompt-editor-preview hr { margin: 0.75em 0; border: none; border-top: 1px solid var(--color-border); }
+.prompt-editor-preview table { border-collapse: collapse; margin: 0.5em 0; width: 100%; }
+.prompt-editor-preview th, .prompt-editor-preview td { border: 1px solid var(--color-border); padding: 0.35em 0.75em; text-align: left; }
+.prompt-editor-preview th { font-weight: 600; background-color: var(--color-muted); }
+.prompt-editor-preview strong { font-weight: 700; }
+.prompt-editor-preview em { font-style: italic; }
+.prompt-editor-preview .token-pill { display: inline-flex; align-items: center; gap: 3px; height: 20px; padding: 0 4px; border-radius: 4px; font-size: 13px; line-height: 20px; vertical-align: middle; background: var(--color-primary-lighter); color: var(--color-foreground); }
+.prompt-editor-preview .token-pill svg { display: block; flex-shrink: 0; color: var(--color-primary); width: 14px; height: 14px; }
+`;
```
CalinaCristian, line 41: is there an alternative here to injecting styles inline?

packages/apollo-wind/src/components/ui/prompt-editor/components/PromptEditorAutocompleteMenu.tsx
```diff
@@ -0,0 +1,176 @@
...
+  }, [open, anchorEl]);
+
+  // Stop pointer events from bubbling past the menu: a mousedown inside would otherwise transfer
+  // focus and drop the editor's caret, and (when portaled to body inside a Radix Dialog) trip the
+  // dialog's `onInteractOutside`. The search input still receives focus via cmdk's autofocus.
+  useEffect(() => {
+    if (!open) return;
+    const node = containerRef.current;
+    if (!node) return;
+    const handlePointerDown = (e: PointerEvent) => e.stopPropagation();
```
CalinaCristian: ### 🟡 M1 — missing `preventDefault` on container `pointerdown`

The original `VariablePickerAutocompleteMenu` registered both handlers on the container node:
```ts
// original
node.addEventListener('mousedown', (e) => { e.preventDefault(); e.stopPropagation(); });
node.addEventListener('pointerdown', (e) => e.stopPropagation());
```
`mousedown preventDefault` prevented the browser from transferring focus away from the Lexical editor when the user clicked a picker row — keeping Lexical's cursor alive for `commitChip`'s `anchorNode.getKey() !== triggerInfo.nodeKey` guard. The new code only stops propagation.

In practice this is likely fine: `CommandInput` already takes focus on mount (so Lexical focus is already gone by click time), and `commitChip` captures `triggerInfoRef.current` at the top of the function before any state change can clear it. But `pointerdown preventDefault` also suppresses the subsequent `mousedown`, giving the same defence in one call:

```ts
const handlePointerDown = (e: PointerEvent) => {
  e.preventDefault(); // suppress mousedown → no focus transfer risk
  e.stopPropagation(); // keep Radix Dialog unaware
};
```

The cost is zero; it removes a class of host-environment edge cases where a racing Lexical state update could invalidate `triggerInfo.nodeKey` between open and commit.

CalinaCristian (commented):
## PR #784 review — PromptEditor extraction

**Overall: correct, intentional decoupling — not a strict 1-to-1 but a well-executed port.** Core Lexical engine, plugin architecture, serialization, and public API surface are faithfully preserved. Several real bugs from the flow-workbench original were fixed along the way (empty-array controlled value, minRows/maxRows clamping, Lexical state corruption in `insertVariableToken`, `$`-prefix on free-typed chip paths, NodeSelection crash guard, borderless text color). No data-loss or crash bugs found in the new code.

**Severity legend:** 🟡 medium = worth landing before merge · 🔵 low = follow-up is fine

---

### ✅ Bug fixes shipped in this extraction (all genuine improvements)

| Fix | Location |
|---|---|
| `root.append(tokenNode)` → wrap in `$createParagraphNode()` | `insert-token.ts` |
| `!value` → `value === undefined` — empty array now clears a controlled editor | `ValueSyncPlugin.tsx` |
| `isEmpty` init seeds from `initialValue ?? value` — fixes controlled-only usage | `prompt-editor.tsx` |
| `effectiveMinRows = Math.min(minRows, maxRows)` — max-height no longer silently overridden | `prompt-editor.tsx` |
| `borderless ? 'inherit' : 'var(--color-foreground)'` — correct text color in parent-provided chrome | `prompt-editor.tsx` |
| `$isRangeSelection` guard — `insertRawText` can't throw on a NodeSelection | `prompt-editor.tsx` |
| Free-typed chip path: `$vars.foo` → `vars.foo` — matches menu-selected chips for validation | `AutocompletePlugin.tsx` |
| `closeMenu(explicit)` flag — auto-close no longer suppresses re-opening at the same `$` position | `AutocompletePlugin.tsx` |
| `target` removed from DOMPurify allowlist — closes tabnabbing vector in markdown preview | `MarkdownPreview.tsx` |
| CSS inlined in JS — self-contained bundle, no unresolved `.css` import next to published JS | `MarkdownPreview.tsx` |

Inline comments below cover the two 🟡 medium and one 🔵 low observations.

packages/apollo-wind/src/components/ui/prompt-editor/types.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+export type PromptEditorToolbarFormatAction =
+  | 'bold'
+  | 'bulletedList'
+  | 'italic'
+  | 'numberedList'
+  | 'strikethrough';
+
+export interface PromptEditorHighlightLocator {
```
CalinaCristian: ### 🟡 M2 — `PromptEditorHighlightLocator` / `PromptEditorHighlightItem` declared but neither exported nor used

These two interfaces are in `types.ts` but are not re-exported from `index.ts` and have no implementation in this PR (no `HighlightPlugin`, no `highlightItems` prop). Consumers importing from `@uipath/apollo-wind` can't reach them even though the types are public in the subpath.

Options:
1. **Export them now** — add them to the barrel if they're part of the intended public contract for a follow-up highlight feature.
2. **Move them later** — keep the interfaces co-located with their future implementation (`highlight-plugin-types.ts` or similar) and add the barrel export when the feature lands.

Either is fine; the current state is just confusing — the types signal API intent but are unreachable by consumers.

packages/apollo-wind/src/components/ui/prompt-editor/components/TokenPill.tsx
```diff
@@ -0,0 +1,122 @@
...
+  output: SquareFunction,
+  state: Database,
+  resource: Paperclip,
+  text: Variable,
+};
+
+export interface TokenPillProps {
+  value: string;
+  /** Token type — selects the leading icon and (via the wrapper) the tooltip label. */
+  tokenType: PromptEditorTokenType;
```
CalinaCristian, line 21: ### 🔵 L1 — `size='md'` variant not ported

The flow-workbench `TokenPill` had `size?: 'sm' | 'md'` for a taller standalone single-variable field (28px chip, larger text). The port hardcodes `sm` geometry and only accepts `tokenType`.

No current apollo-wind consumer needs the medium size, so this is fine to ship as-is. Flagging for when a standalone single-token field is added to the design system — the variant will need to be re-introduced, ideally as a named `size` prop to keep the API consistent with other apollo-wind components.

package.json
```diff
@@ -14,14 +14,14 @@
     "build:packages": "turbo run build --filter='./packages/*'",
     "build:apps": "turbo run build --filter='./apps/*'",
     "build:analyze": "turbo run build:analyze",
-    "check:dependencies": "check-dependency-version-consistency .",
+    "check:dependencies": "check-dependency-version-consistency . --ignore-dep-pattern \"^(lexical|@lexical/.+)$\"",
```
CalinaCristian: This repo-wide Lexical exemption should not be added.

Changing `check:dependencies` / `fix:dependencies` to ignore `lexical` and every `@lexical/*` package disables dependency consistency checks for the whole Lexical family across the monorepo. That means future accidental version drift in any package would be invisible to the normal guardrail.

If `apollo-react` and `apollo-wind` currently need different Lexical versions, we should resolve that explicitly instead of weakening the repo-wide check: align the Lexical versions if possible, or use a narrowly scoped exception if the tool supports package/path-specific ignores.

Please revert the `--ignore-dep-pattern "^(lexical|@lexical/.+)$"` change from the root scripts.

packages/apollo-wind/src/index.ts
```diff
@@ -391,3 +391,16 @@ export {
   isFileField,
   isCustomField,
 } from './components/forms/form-schema';
+
+// -----------------------------------------------------------------------------
+// Prompt Editor
+// -----------------------------------------------------------------------------
+export { PromptEditor } from './components/ui/prompt-editor';
```
CalinaCristian: Should this also be exported from `packages/apollo-wind/src/components/ui/index.ts`?

The package root exports `PromptEditor`, but the `components/ui` barrel exports every other UI component and currently skips `./prompt-editor`. Consumers importing from `@uipath/apollo-wind/components/ui` will not see it.

Suggested fix:

```ts
export * from './prompt-editor';
```

packages/apollo-wind/src/components/ui/prompt-editor/prompt-editor.stories.tsx
```diff
@@ -0,0 +1,149 @@
...
+      </div>
+    ),
+  ],
+} satisfies Meta<typeof PromptEditor>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+const AUTOCOMPLETE_OPTIONS: PromptEditorAutoCompleteOption[] = [
+  { type: 'input', value: 'vars.firstName' },
```
CalinaCristian, line 38: Can we document the token value convention for `PromptEditorAutoCompleteOption.value` and token values?

The stories/tests use values like `vars.firstName`, while flow-workbench currently uses `$vars.firstName`. That is fine if Apollo’s public API intentionally owns the no-`$` convention, but it should be explicit because serialization/validation depend on exact string equality.

packages/apollo-wind/src/components/ui/prompt-editor/prompt-editor.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+    // edit mode, controlled value
+    expect(() => render(<PromptEditor ariaLabel="Prompt" value={badTokens} />)).not.toThrow();
+    expect(screen.getByRole('textbox', { name: 'Prompt' })).toBeInTheDocument();
+    // preview mode treats the malformed value as empty rather than rendering junk
+    const { getByText } = render(
+      <PromptEditor value={badTokens} initialValue={badTokens} mode="preview" />
+    );
+    expect(getByText('Nothing to preview')).toBeInTheDocument();
+  });
+});
```
CalinaCristian, line 203: The current tests cover rendering/mounting/preview smoke paths, but the extraction-risky flows are mostly untested.

Can we add focused tests for:
- `$` autocomplete selection committing a token
- free-form Enter commit after `$vars.foo`
- invalid-token validation against `autoCompleteOptions`
- rename-on-options-change behavior
- copy/paste token serialization
- toolbar formatting on multiline/backward selections

Those are the places most likely to regress during a design-system extraction.

packages/apollo-wind/src/components/ui/prompt-editor/components/MarkdownPreview.tsx
```diff
@@ -0,0 +1,200 @@
...
+ * the leading icon is chosen by token type to match what the editor renders in edit mode.
+ */
+const tokensToMarkdownWithPills = (tokens: PromptEditorToken[]): string => {
+  let md = '';
+  for (const token of tokens) {
+    if (token.type === 'text') {
+      md += token.value;
+    } else {
+      const iconSvg = buildTokenIconSvgMarkup(token.type);
+      md += `<span class="token-pill">${iconSvg}${escapeHtml(token.value)}</span>`;
```
CalinaCristian, line 165: Flow-workbench preview marks stale/unresolved variables invalid, while this preview always renders token pills as valid.

Can preview either:
1. receive/use `autoCompleteOptions` to mirror `ValidateTokensPlugin`, or
2. explicitly document that preview is visual-only and does not reflect token validity?

Right now edit mode and preview mode can disagree about whether a token is valid.

packages/apollo-wind/.storybook/preview-head.html
fikewa-olatunji: Rebased onto latest `main` and checked: `main`'s `.storybook` (preview-head.html, preview.tsx, manager, theme.ts) has no `.sb-story` / docs-story background rule, so dropping this reintroduces the white-on-white borderless preview in autodocs under the dark theme. Kept it for now — if the fix is on a not-yet-merged PR (or targets the canvas rather than autodocs), happy to remove once it lands. Could you point me to it?

packages/apollo-wind/src/components/ui/prompt-editor/components/MarkdownPreview.tsx
fikewa-olatunji: Moved the rules out of the inline `<style>` into the package stylesheet (`src/styles/tailwind.utilities.css`, scoped under `.prompt-editor-preview`), which consumers already import via `@uipath/apollo-wind/tailwind.css` and Storybook loads through `tailwind.css`. That drops the inline injection while keeping the bundless build clean — a per-component `.css` import doesn't resolve next to the emitted JS. Done in cecde505.

packages/apollo-wind/src/components/ui/prompt-editor/components/PromptEditorAutocompleteMenu.tsx
fikewa-olatunji: Done — added `e.preventDefault()` to the container `pointerdown` (with a comment on why), so the follow-up mousedown can't move focus off the editor before commit. cecde505.

packages/apollo-wind/src/components/ui/prompt-editor/types.ts
fikewa-olatunji: Removed both interfaces — they had no implementation or barrel export in this PR. cecde505.

packages/apollo-wind/src/components/ui/prompt-editor/components/TokenPill.tsx
fikewa-olatunji: Agreed — leaving `sm`-only for now since no apollo-wind consumer needs the medium size. Will reintroduce it as a named `size` prop when a standalone single-token field is added, to keep the API consistent.

package.json
fikewa-olatunji: Narrowed it: instead of ignoring the lexical family repo-wide, the check now uses `--ignore-package @uipath/apollo-wind`, so version drift for lexical (or anything else) in every other package is still caught. apollo-react pins lexical 0.16 and apollo-wind needs 0.42 (React 19 + 0.42-only APIs); the tool has no per-dep-per-package ignore, so this scopes the exception to just the one diverging package rather than weakening the guardrail globally. 368ac018.

packages/apollo-wind/src/index.ts
fikewa-olatunji: Done — added `export * from './prompt-editor'` to `components/ui/index.ts`. cecde505.

packages/apollo-wind/src/components/ui/prompt-editor/prompt-editor.stories.tsx
fikewa-olatunji: Documented it on `PromptEditorToken` and `PromptEditorAutoCompleteOption` in types.ts: apollo-wind owns the no-`$` convention (e.g. `vars.firstName`), and serialization + `ValidateTokensPlugin` compare by exact string equality, so option values and token values must use the same `$`-less form. cecde505.

packages/apollo-wind/src/components/ui/prompt-editor/prompt-editor.test.tsx
fikewa-olatunji: Added focused unit tests for the extraction-risky logic (368ac018): clipboard token serialization round-trip, free-form path type inference + `VARIABLE_PATH_REGEX`, the dismissed-trigger suppression sentinel, and token-array equality. The interaction-level flows ($-keypress→commit, toolbar on multiline/backward selections) stay limited under jsdom — the existing test file notes Lexical decorators don't paint reliably there — so those are covered at the logic layer rather than via simulated editing.

packages/apollo-wind/src/components/ui/prompt-editor/components/MarkdownPreview.tsx
fikewa-olatunji: Documented preview as visual-only: it doesn't receive `autoCompleteOptions`, so every token renders as a pill, and edit mode's `ValidateTokensPlugin` remains the source of truth for validity (note added to the `MarkdownPreview` doc comment, cecde505). I kept it visual-only to stay consistent with the earlier removal of the `data-invalid` path — happy to thread the options through for full parity if you'd prefer.

packages/apollo-wind/.storybook/preview-head.html
CalinaCristian: on current deployed apollo-design there's no white-on-white https://apollo-design.vercel.app/?path=/docs/apollo-wind-components-core-button--docs

package.json
CalinaCristian: this is still a workaround -- we shouldn't ignore packages for the check, makes the check redundant

package.json
```diff
@@ -14,13 +14,13 @@
     "build:packages": "turbo run build --filter='./packages/*'",
     "build:apps": "turbo run build --filter='./apps/*'",
     "build:analyze": "turbo run build:analyze",
-    "check:dependencies": "check-dependency-version-consistency .",
+    "check:dependencies": "check-dependency-version-consistency . --ignore-package @uipath/apollo-wind",
```
CalinaCristian: we shouldn't do this.

package.json
```diff
@@ -14,13 +14,13 @@
...
     "commit": "git-cz",
     "dev": "turbo run dev",
     "dev:chat": "turbo run dev --filter=@uipath/ap-chat",
     "dev:apollo-vertex": "turbo run dev --filter=apollo-vertex",
     "dev:apollo-docs": "turbo run dev --filter=apollo-docs",
-    "fix:dependencies": "check-dependency-version-consistency . --fix",
+    "fix:dependencies": "check-dependency-version-consistency . --ignore-package @uipath/apollo-wind --fix",
```
CalinaCristian: we shouldn't do this.

packages/apollo-wind/src/components/ui/prompt-editor/prompt-editor.stories.tsx
CalinaCristian: can you experiment with publishing a dev package for this (`dev-packages` label and try consume it in flow-workbench to replace existing one and verify there are no regressions?)

packages/apollo-wind/src/components/ui/prompt-editor/components/MarkdownPreview.tsx
CalinaCristian: we need parity -- best test is to try to repalce this using the `dev-packages` label, install the dev version and try to replace the flow-workbench PromptEditor with this one and verify full parity on functionality.

fikewa-olatunji: Ran the swap: dev pkg `@uipath/apollo-wind@2.19.0-pr784.9c7181c` dropped into flow's `PromptEditorField` (replacing the local core editor) — `@flow/canvas` typechecks clean, **0 prop/type mismatches** across all 16 props + shared types. One gap: `mapVarDropToToken` (variable drag-drop) wasn't ported — intentional scope call, needs a decision (port `VariableDropPlugin` or drop drag-drop). Autocomplete UI was rebuilt on apollo-wind primitives (same `autoCompleteOptions` API).

packages/apollo-wind/src/components/ui/prompt-editor/prompt-editor.stories.tsx
fikewa-olatunji: Validated via the dev package + flow-workbench swap — drop-in clean except `mapVarDropToToken` (drag-drop, intentionally not ported). Details in the preview-parity thread.

packages/apollo-wind/src/styles/tailwind.utilities.css
```diff
@@ -60,4 +60,39 @@
...
+.prompt-editor-preview a { color: var(--color-primary); text-decoration: underline; text-underline-offset: 2px; }
+.prompt-editor-preview a:hover { opacity: 0.8; }
+.prompt-editor-preview hr { margin: 0.75em 0; border: none; border-top: 1px solid var(--color-border); }
+.prompt-editor-preview table { border-collapse: collapse; margin: 0.5em 0; width: 100%; }
+.prompt-editor-preview th, .prompt-editor-preview td { border: 1px solid var(--color-border); padding: 0.35em 0.75em; text-align: left; }
+.prompt-editor-preview th { font-weight: 600; background-color: var(--color-muted); }
+.prompt-editor-preview strong { font-weight: 700; }
+.prompt-editor-preview em { font-style: italic; }
+.prompt-editor-preview .token-pill { display: inline-flex; align-items: center; gap: 3px; height: 20px; padding: 0 4px; border-radius: 4px; font-size: 13px; line-height: 20px; vertical-align: middle; background: var(--color-primary-lighter); color: var(--color-foreground); }
+.prompt-editor-preview .token-pill svg { display: block; flex-shrink: 0; color: var(--color-primary); width: 14px; height: 14px; }
```
CalinaCristian: don't love this in this file, it should be a scoped file to the prompt editor IMO

packages/apollo-wind/src/styles/tailwind.consumer.css
```diff
@@ -15,6 +15,11 @@
 
 @import "./tailwind.utilities.css";
 
+/* Component-scoped styles that need element selectors (not Tailwind classes), pulled in here so they
+   ship with the package stylesheet without a per-component JS `.css` import (unsupported by the
+   bundleless build). Kept in `styles/` (same dir) so the same-dir import survives the flat copy to dist. */
+@import "./prompt-editor.css";
```
CalinaCristian: shouldn't this be only imported in the prompt editor component?
I guess inline is better than this TBH -- so we can go back to that

fikewa-olatunji: Both alternatives are actually why it ended up here: a per-component `import './prompt-editor.css'` is the one thing that *doesn't* work, a JS CSS import doesn't resolve next to the emitted JS in the bundleless rslib build (the original bug), and inline `<style>` re-injects the whole stylesheet into the DOM on every preview render. The current setup keeps all the rules in one prompt-editor-scoped file and ships them once via a single `@import` in the stylesheet consumers already load

CalinaCristian: yes, but they get loaded even though the consumer doesn't use the prompt editor component, that's why inline styles in js for the prompt editor (the first approach) I think was better

fikewa-olatunji: Okay. I will revert now

CalinaCristian (approved):

PR #790, opened 2026-06-08
feat(apollo-vertex): add Solution Tests data layer (collection-backed hooks) [AGVSOL-3355]

frankkluijtmans:
## Summary

The **data layer** for the Solution Tests component, split out of #782 so it
can land and be reviewed ahead of the UI. The view depends on this, so the
**components PR (#782) is stacked on this branch** and merges after it.

This layer is collection-backed: reads stream from the vs-core 2.x Solution Test
collections, and writes go through the standard UiPath Solution Test API
triggers. There is no bespoke adapter or local read cache.

Contents (all under `registry/solution-tests/`, plus `en.json` keys):

- **Reads** — the per-query hooks `use-solution-tests`,
  `use-solution-test-batch-runs`, `use-solution-test-runs`, `use-baseline-jobs`,
  `use-run-results` read the `UiPathST*` collections directly from
  `solution.api.collections.solutionTests` via `useLiveQuery`. A dev-only type
  shim in `types/optional-deps.d.ts` types the name-keyed collection map for this
  repo's own typecheck; consumers rely on their own vs-core typings (nothing is
  shipped to override them).
- **Writes** — `actions.ts` (the `SolutionTestsActions` interface; each method
  rejects on failure), `create-actions.ts` (`createSolutionTestActions` builds it
  from `triggerBaseUrl` + `getToken`), and the mutation hooks (`mutations.ts` +
  `use-force-stop.ts`) run actions through TanStack Query's `useMutation`.
  Toggle-active is a direct collection `.update()` in `use-solution-tests`;
  attachment reads use the uipath-typescript SDK (`attachments.ts`).
- **Wiring** — `context.tsx` (`SolutionTestsProvider` + context hooks),
  `config.ts` (`SolutionTestsConfig`), `constants.ts` (collection names, trigger
  slugs/paths, labels, thresholds: the single setup file), `hooks.ts` (barrel).
- **Validation** — `user-messages.ts` validates parsed user-message payloads
  with `zod`.
- `types.ts`, `status-maps.ts`, `utils.ts`, and the `en.json` translation keys.

New runtime dependencies declared on the registry item: `@tanstack/react-query`
(mutations) and `zod` (validation), alongside the existing `@tanstack/react-db`,
`@uipath/vs-core`, and `@uipath/uipath-typescript`.

## Consumer interface

The smart container has four inputs, all on `SolutionTestsProvider`. Reads are
implicit (the `UiPathST*` collections on the consumer's vs-core solution); writes,
attachment resolution, and presentation are explicit props:

```tsx
interface SolutionTestsProviderProps {
  // Required. Base URL write-action slugs are appended to (no trailing slash):
  // `{base}/{org}/{tenant}/orchestrator_/t/{folderKey}`.
  triggerBaseUrl: string;
  // Required. Resolves the caller's current bearer token (refreshed near expiry).
  getToken: () => Promise<string | null> | string | null;
  // Required. Resolves a Solution Test entity name (e.g. `UiPathSTRunResults`) to
  // its DataFabric id (GUID); used for attachment reads. vs-core does not expose
  // this mapping, so the consumer supplies it (e.g. from its generated entities).
  getEntityId: (name: string) => string | undefined;
  // Optional per-vertical presentation config.
  config?: SolutionTestsConfig;
  children: ReactNode;
}

interface SolutionTestsConfig {
  // Columns inserted between Test Name and Version. Usually the only customization.
  subjectColumns?: ColumnDef<SolutionTest>[];
  // Turns the test name into a link to its subject.
  getSubjectHref?: (test: SolutionTest) => string | undefined;
  // Subject noun for labels, e.g. { singular: "Loan", plural: "Loans" }.
  subjectNoun?: { singular: string; plural: string };
  // Score at/above which a result passes (pass color + KPI trend line). Default 0.9.
  passThreshold?: number;
}
```

A consumer is responsible for exactly four things:

1. **Reads** — ensure the name-keyed `UiPathST*` collections exist on the vs-core
   solution (`solution.api.collections.solutionTests`). The hooks query them
   reactively; no further data wiring.
2. **Writes** — pass `triggerBaseUrl` + `getToken`. The provider builds the whole
   write surface (run, delete, force-stop, adopt / update / remove baseline) from
   them. Toggle-active isn't a trigger and needs no wiring. Mutations use
   `useMutation`, so the container must render under a `QueryClientProvider`
   (vs-core's solution provider supplies one).
3. **Attachments** — pass `getEntityId`. DataFabric addresses entities by GUID,
   not name, and vs-core doesn't surface the name-to-GUID mapping, so attachment
   reads (expected / actual output, evaluator results) need the consumer to
   resolve it (typically from its generated entities map).
4. **Presentation** — pass `config.subjectColumns` (and optionally
   `getSubjectHref` / `subjectNoun` / `passThreshold`) to flavor the table for the
   vertical.

Everything else (evaluator labels, status labels, the poll interval) is
hard-coded in `constants.ts`. The pass threshold defaults to `0.9` there but is
overridable via `config.passThreshold`.

## Testing

`pnpm typecheck`, `pnpm lint`, `pnpm format`, and `pnpm registry:build` pass.

👨 Generated with Kluijt Code


apps/apollo-vertex/registry/solution-tests/constants.ts
```diff
@@ -0,0 +1,65 @@
...
+  adoptJob: "/solution-tests/adopt-job",
+  updateBaseline: "/solution-tests/update-baseline",
+  removeJobBaseline: "/solution-tests/remove-job-baseline",
+} as const;
+
+/** Agent `output_data.status` values treated as success. */
+export const AGENT_SUCCESS_STATUSES = new Set(["success", "partial_success"]);
+
+/** Score ≥ threshold renders as a pass and drives the KPI trend reference line. */
+export const PASS_THRESHOLD = 0.9;
```
RobinMennens-UiPath: The `0.9` pass threshold drives pass/fail rendering and the KPI trend reference line. That is a per-vertical business calibration, not a platform constant. Consider moving it into `SolutionTestsConfig` (e.g. `passThreshold?: number`) with `resolveConfig` defaulting to `0.9`, so a vertical can tune its pass bar without forking the data layer.

apps/apollo-vertex/registry/solution-tests/use-run-results.ts
```diff
@@ -0,0 +1,79 @@
...
+  results: SolutionTestRunResult[];
+  isLoading: boolean;
+}
+
+/** Resolve a result's job role, whether the job is embedded or a bare id. */
+function isBaselineEntryPoint(result: SolutionTestRunResult): boolean {
+  const job = result.SolutionTestJobId;
+  if (job != null && typeof job === "object") {
+    return job.JobRole === JobRole.Baseline;
+  }
```
RobinMennens-UiPath: Two concerns here:

1. Naming: `JobRole` is either `EntryPoint` (0) or `Baseline` (1), so `isBaselineEntryPoint` welds together two mutually-exclusive roles and describes neither. It actually means "this result's job is a Baseline job (or the role is unknown, so keep the row)". Suggest renaming to e.g. `isBaselineJobResult` and reconciling the doc comment on line 39 ("drops non-baseline entry-point results").

2. Likely no-op filter: the read is `useLiveQuery(() => collection)` over the raw collection, so `SolutionTestJobId` will usually be a bare id string rather than an expanded job. In that case this returns `true` for every row and the filter drops nothing, contradicting the documented intent. Can you confirm the collection actually embeds the job? If it does not, the filter cannot do its job here and the row-visibility decision needs to happen where the job role is known.

apps/apollo-vertex/registry/solution-tests/create-actions.ts
```diff
@@ -0,0 +1,133 @@
...
+function agentStatus(data: unknown): string | null {
+  if (!isRecord(data)) return null;
+  const output = isRecord(data.output_data) ? data.output_data : data;
+  return typeof output.status === "string" ? output.status : null;
+}
+
+/** App-level failure across both contracts (HTTP stays 200 even on RBAC denial). */
+function detectFailure(data: unknown): string | null {
+  if (!isRecord(data)) return null;
+  if (typeof data.status_code === "number" && data.status_code >= 400) {
```
RobinMennens-UiPath: `agentStatus` unwraps `data.output_data` to read `status`/`message`, but this `status_code >= 400` check only looks at top-level `data.status_code`. If the agent (`run_solution_tests`) contract can report an app/RBAC failure as `output_data.status_code`, this would miss it and treat the 200 as success. Worth confirming the backend contract. If `status_code` can appear nested, run this check against the same unwrapped `output` object that `agentStatus` uses.

apps/apollo-vertex/registry/solution-tests/constants.ts
```diff
@@ -0,0 +1,65 @@
...
+export const ENTITY = {
+  tests: "UiPathSTTests",
+  batchRuns: "UiPathSTBatchRuns",
+  runs: "UiPathSTRuns",
+  jobs: "UiPathSTJobs",
+  runResults: "UiPathSTRunResults",
+} as const;
+
+/** API-trigger slugs appended to the consumer's trigger base URL. */
+export const RUN_SLUG = "run_solution_tests";
```
RobinMennens-UiPath: Naming: repo style prefers longer-but-clear names over abbreviations. Consider `RUN_TESTS_SLUG`, `AUTOMATION_FUNCTIONS_SLUG`, and `AUTOMATION_FUNCTION_PATH` for `FN_PATH`.

apps/apollo-vertex/registry/solution-tests/use-solution-tests.ts
```diff
@@ -0,0 +1,82 @@
...
+> {
+  const solution = useSolution();
+  const { mutateAsync, isPending } = useMutation({
+    mutationFn: async ({
+      testId,
+      isActive,
+    }: {
+      testId: string;
+      isActive: boolean;
+    }) => {
```
RobinMennens-UiPath, line 60: Minor naming: `tx` is an abbreviation; this is a transaction, so `transaction` reads more clearly per the repo's naming preference.

RobinMennens-UiPath (changes requested):
A few changes requested, focused on configurability and one likely-inert filter. Inline comments below. Overall the move-over is clean and domain-neutral (no FINS-specific coupling found); the main asks are around making per-vertical values configurable and confirming the run-results filter actually works against the raw collection.

apps/apollo-vertex/registry/solution-tests/constants.ts
frankkluijtmans: Agreed. Moved it into `SolutionTestsConfig` as `passThreshold?: number`, with `resolveConfig` defaulting to `0.9` (now `DEFAULT_PASS_THRESHOLD` in constants). `kpi-bar` and `evaluator-results-view` read it from `useSolutionTestsConfig()` instead of importing the constant, so a vertical can tune its pass bar via config. (8cdef332)

apps/apollo-vertex/registry/solution-tests/use-run-results.ts
frankkluijtmans: Both fixed. You're right that `SolutionTestJobId` is always a bare id string off the raw collection, so the old filter was a no-op. Renamed to `isBaselineJobResult`, and it now reads the `UiPathSTJobs` collection in the same hook and joins the bare id against a `Map<jobId, job>` to read `JobRole`: keep when the id is null or the job isn't found, otherwise keep only `Baseline` (dropping entry-points). Doc comment reconciled. (8cdef332)

apps/apollo-vertex/registry/solution-tests/create-actions.ts
frankkluijtmans: Confirmed it can be nested. `detectFailure` now unwraps `output_data` once and runs the `status_code >= 400` check (and reads `error`/`message`) against that same unwrapped `output` object that `agentStatus` uses, so both the top-level RPC-router contract and the nested agent contract are covered. (8cdef332)

apps/apollo-vertex/registry/solution-tests/constants.ts
frankkluijtmans: Done: renamed to `RUN_TESTS_SLUG`, `AUTOMATION_FUNCTIONS_SLUG`, and `AUTOMATION_FUNCTION_PATH`. (8cdef332)

apps/apollo-vertex/registry/solution-tests/use-solution-tests.ts
frankkluijtmans: Done, renamed `tx` to `transaction`. (8cdef332)

apps/apollo-vertex/registry/solution-tests/config.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/** Per-vertical presentation config; everything else is hard-coded in `constants`. */
+export interface SolutionTestsConfig {
+  /** Columns inserted between the Test Name and Version columns. */
+  subjectColumns?: ColumnDef<SolutionTest>[];
+  /** When set, the test name links to its subject. */
+  getSubjectHref?: (test: SolutionTest) => string | undefined;
+  subjectNoun?: { singular: string; plural: string };
+  /** Score at/above which a result passes (drives pass color + KPI trend line). Defaults to 0.9. */
+  passThreshold?: number;
```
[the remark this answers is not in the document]
frankkluijtmans: Keeping `passThreshold` (per review feedback it's a per-vertical calibration, not a platform constant). Updated the PR description to document it under `SolutionTestsConfig` and corrected the line that said the 0.9 threshold is hard-coded. It now defaults to 0.9 in constants and is overridable via `config.passThreshold`.

apps/apollo-vertex/registry/solution-tests/context.tsx
```diff
@@ -0,0 +1,79 @@
...
+);
+
+interface SolutionTestsProviderProps {
+  config?: SolutionTestsConfig;
+  /** Base URL each action slug is appended to (no trailing slash). */
+  triggerBaseUrl: string;
+  getToken: () => Promise<string | null> | string | null;
+  /** Resolves a Solution Test entity name to its DataFabric id (GUID); used for attachment reads. */
+  getEntityId: EntityIdResolver;
+  children: ReactNode;
```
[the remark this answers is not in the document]
frankkluijtmans: The description was stale, now fixed. `getEntityId` is genuinely required: DataFabric addresses entities by GUID and vs-core doesn't surface the name-to-GUID mapping, so the consumer must resolve it for attachment reads. Kept it required (compile-time guarantee that attachments are wired) rather than optional-with-no-op, which would let attachments silently return null. The PR description now lists it as a required prop with that rationale.

apps/apollo-vertex/registry/solution-tests/context.tsx
```diff
@@ -0,0 +1,79 @@
...
+  getToken,
+  getEntityId,
+  children,
+}) => {
+  const value = useMemo(
+    () => ({
+      config: resolveConfig(config),
+      actions: createSolutionTestActions({ triggerBaseUrl, getToken }),
+      getEntityId,
+    }),
```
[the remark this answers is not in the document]
frankkluijtmans: As above: `getEntityId` stays required, so there's no optional/absent case to default. The PR description now documents it as required.

RobinMennens-UiPath (approved):
Re-reviewed after the rebase. All four requested changes are addressed:

- `PASS_THRESHOLD` is now `passThreshold?: number` on `SolutionTestsConfig` (defaults via `DEFAULT_PASS_THRESHOLD`).
- `isBaselineJobResult` now joins `UiPathSTJobs` to read `JobRole` off the bare id, so the previously inert entry-point filter actually drops rows now.
- `detectFailure` unwraps `output_data` once and checks `status_code`/`error`/`message` against the same object, covering both contracts.
- Constant/variable renames done (`RUN_TESTS_SLUG`, `AUTOMATION_FUNCTIONS_SLUG`, `AUTOMATION_FUNCTION_PATH`, `transaction`).

Also sanity-checked the broader restructure (react-query `useMutation` adoption, removal of the `collections.ts` boundary helper in favor of typed stubs, entity-id plumbing for attachments). `pnpm typecheck` and `pnpm lint` both pass locally. LGTM.

apps/apollo-vertex/registry/solution-tests/use-solution-tests.ts
```diff
@@ -0,0 +1,82 @@
...
+export interface UseSolutionTestsResult {
+  tests: SolutionTest[];
+  isLoading: boolean;
+}
+
+/** Live list of solution tests. */
+export function useSolutionTests(): UseSolutionTestsResult {
+  const solution = useSolution();
+  const { data, isLoading } = useLiveQuery<SolutionTest>((q) =>
+    q.from({ tests: solution?.api.collections.solutionTests[ENTITY.tests] }),
```
KokoMilev: This live query can run before the collection is ready.

On the first render, `useSolution()` may still be `null`, so this can pass `undefined` into `q.from`:

```ts
q.from({ tests: solution?.api.collections.solutionTests[ENTITY.tests] })
```

That means the hook may create a query before it has a real collection to read from. It also does not pass a dependency array, so the query may not be rebuilt when the collection becomes available later.

Could we guard the query and add the collection as a dependency?

```ts
const collection = solution?.api.collections.solutionTests[ENTITY.tests];

const { data, isLoading } = useLiveQuery(
  (q) => (collection ? q.from({ tests: collection }) : undefined),
  [collection],
);
```

Same pattern applies to the other Solution Tests `useLiveQuery` hooks.

apps/apollo-vertex/registry/solution-tests/use-run-results.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/** Live run results for a run, excluding entry-point jobs (role != Baseline). */
+export function useRunResults(runId: string): UseRunResultsResult {
+  const solution = useSolution();
+  const { data, isLoading } = useLiveQuery<SolutionTestRunResult>((q) =>
+    q.from({
+      results: solution?.api.collections.solutionTests[ENTITY.runResults],
+    }),
+  );
+  const { data: jobsData } = useLiveQuery<SolutionTestJob>((q) =>
```
KokoMilev: `useRunResults` depends on two collections: run results and jobs.

The jobs collection is needed to tell whether each result belongs to a baseline job or an entry-point job. Right now, if run results load before jobs, `jobsById` is empty and unresolved jobs are kept as valid results.

That means entry-point results can briefly appear even though this hook is meant to hide them.

Could we keep the hook in a loading state until both collections are ready, then apply the filter?

KokoMilev (changes requested):

apps/apollo-vertex/registry/solution-tests/use-solution-tests.ts
frankkluijtmans: This problem didn't surface in my test sandbox because of a suspense boundary, but let's make this change to make it a bit more reliable.

frankkluijtmans: Good catch, fixed. All six `useLiveQuery` call sites now resolve the collection first, return a disabled query when it's absent, and pass it as a dependency so the query rebuilds once the collection appears:
```ts
const collection = solution?.api.collections.solutionTests[ENTITY.tests];
const { data, isLoading } = useLiveQuery<SolutionTest>(
  (q) => (collection ? q.from({ tests: collection }) : null),
  [collection],
);
```
Used `: null` rather than `: undefined` because the `no-undefined` lint rule bans the literal; the library treats both as a disabled query. Applied to `use-solution-tests`, `use-solution-test-batch-runs`, `use-solution-test-runs`, `use-baseline-jobs`, and both queries in `use-run-results`. (3dc496a1)

apps/apollo-vertex/registry/solution-tests/use-run-results.ts
frankkluijtmans: Fixed. `useRunResults` now gates on both queries being ready before filtering, returning `{ results: [], isLoading: true }` until then, so entry-point results can no longer flash before the jobs join resolves:
```ts
const { data, isReady: resultsReady } = useLiveQuery<SolutionTestRunResult>(...);
const { data: jobsData, isReady: jobsReady } = useLiveQuery<SolutionTestJob>(...);
if (!resultsReady || !jobsReady) return { results: [], isLoading: true };
```
Used `isReady` rather than OR-ing `isLoading`, since a disabled query reports `isLoading: false` and would defeat the gate. The "keep when job not found" branch stays, but now only covers a genuinely orphaned job rather than the load-order race. (3dc496a1)

KokoMilev (commented):
Copilot woke up 😆 , let me know when you resolve its comments, otherwise it looks good to me.

apps/apollo-vertex/registry/solution-tests/user-messages.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (result.success) items.push(result.data);
+  }
+  return items;
+}
+
+const SEVERITY_ORDER: Record<string, number> = {
+  Error: 2,
+  Warning: 1,
+  Info: 0,
+};
```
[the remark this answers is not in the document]
frankkluijtmans: Done. Typed it as `Record<UserMessageItem["category"], number>`, so a missing category is now a compile error. That also makes the index lookups provably defined, so I dropped the now-dead `?? 0` fallbacks. (0282963e)

apps/apollo-vertex/registry/solution-tests/user-messages.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export function worstSeverity(
+  messages: UserMessageItem[],
+): "Error" | "Warning" | "Info" | null {
+  if (messages.length === 0) return null;
```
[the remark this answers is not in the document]
frankkluijtmans: Done, return type is now `UserMessageItem["category"] | null`. (0282963e)

KokoMilev (approved):

PR #792, opened 2026-06-08
fix(apollo-react): match escalation tool names case-insensitively in agent canvas

akraminakib:
Two related Agent Canvas (`@uipath/apollo-react`) bug fixes surfaced while bumping apollo-react v3 → v4, plus a small type cleanup.

## 1. Escalation nodes/edges not highlighting during execution

On the agent execution view, an escalation node never glowed blue/green and the canvas didn't focus it, even though the run was clearly escalating.

**Root cause:** escalation toolCall spans preserve the resource-name casing (e.g. `escalate_Escalation_1`), but `normalizeToolName` lowercases it (`escalate_escalation_1`). The `toolName` equality in `hasResourceStatus` therefore never matched mixed-case escalations, so the node got no running/success status — and since the edge color is derived from the connected node's status, the path stayed grey. It looked "intermittent" because lowercase-named escalations matched fine while mixed-case ones (the default `Escalation_1`) never did. Regular tools were unaffected (their normalization doesn't lowercase).

**Fix:** compare tool names case-insensitively (lowercase both sides).

## 2. Span parsing hardening

`hasResourceStatus` / `hasModelStatus` / `hasAgentRunning` called `JSON.parse` unguarded and compared `span.Status` strictly. During live streaming a malformed/partial span could throw (aborting node-status computation for the whole render) or arrive before `Status` was populated.

**Fix:** `parseSpanAttributes` guards `JSON.parse` (skips bad spans instead of throwing), and a missing/non-numeric `Status` is treated as `RUNNING` so in-progress highlighting stays stable while spans stream in.

## 3. Breakpoint icon rendered as a hollow circle

After the canvas icon migration, the breakpoint adornment used Lucide's `circle`, which is stroke-only (`fill="none"`), so it showed as a hollow outline instead of the expected solid dot.

**Fix:** forward a `fill` prop through `CanvasIcon` → `getIcon` to the underlying icon, and set it on the breakpoint dot so it renders a solid red circle.

## Tests
Added regression tests in `props-helpers.test.ts` for mixed-case escalation spans, malformed-JSON attributes, and missing-`Status` handling.

## Files
- `canvas/utils/props-helpers.ts` — case-insensitive tool-name match + guarded parsing + default RUNNING status
- `canvas/utils/icon-registry.tsx` — `CanvasIcon`/`getIcon` now forward a `fill` prop
- `canvas/components/AgentCanvas/nodes/ResourceNode.tsx` — filled breakpoint dot
- `canvas/utils/props-helpers.test.ts` — regression tests

### DIFF
--- a/packages/apollo-react/src/canvas/components/AgentCanvas/nodes/ResourceNode.tsx
+++ b/packages/apollo-react/src/canvas/components/AgentCanvas/nodes/ResourceNode.tsx
@@ -355,7 +355,7 @@ export const ResourceNode = memo(
 
     const breakpointAdornment = useMemo((): React.ReactNode => {
       if (!hasBreakpoint) return undefined;
-      return <CanvasIcon icon="circle" size={14} color="#cc3d45" />;
+      return <CanvasIcon icon="circle" size={14} color="#cc3d45" fill="#cc3d45" />;
     }, [hasBreakpoint]);
 
     const statusAdornment = useMemo((): React.ReactNode => {
--- a/packages/apollo-react/src/canvas/utils/icon-registry.tsx
+++ b/packages/apollo-react/src/canvas/utils/icon-registry.tsx
@@ -10,7 +10,12 @@ import { icons } from 'lucide-react';
 import { type JSX, memo, useMemo } from 'react';
 import * as Icons from '../icons';
 
-export type IconComponent = (props: { w?: number; h?: number; color?: string }) => JSX.Element;
+export type IconComponent = (props: {
+  w?: number;
+  h?: number;
+  color?: string;
+  fill?: string;
+}) => JSX.Element;
 
 /**
  * Registry of available icons (UIPath icons only)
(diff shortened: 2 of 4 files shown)


packages/apollo-react/src/canvas/components/AgentCanvas/nodes/ResourceNode.tsx
```diff
@@ -355,7 +355,7 @@ export const ResourceNode = memo(
 
     const breakpointAdornment = useMemo((): React.ReactNode => {
       if (!hasBreakpoint) return undefined;
-      return <CanvasIcon icon="circle" size={14} color="#cc3d45" />;
+      return <CanvasIcon icon="circle" size={14} color="#cc3d45" fill="#cc3d45" />;
```
Holychung, line 358: can we use `--canvas-error-icon`?

Holychung (approved):

packages/apollo-react/src/canvas/components/AgentCanvas/nodes/ResourceNode.tsx
akraminakib: No, we'll keep it same color as form builder, and it has been that

snuziale (approved):

PR #797, opened 2026-06-09
feat(apollo-react): add Optional and Ends case status badges to StageNode header

BenNormann:
## Summary

Adds two non-interactive status badges — **Optional** and **Ends case** — to the `StageNode` canvas header in `@uipath/apollo-react`. They are pure status indicators (no count, no click) driven by new optional `stageDetails` flags, complementing the existing interactive header chips and SLA text. This is **part 1 of a 2-part feature**; part 2 (the domain logic that computes `isOptional` / `endsCase` and supplies localized copy) lands separately in **PO.Frontend**. Relates to **MST-10799**.

## Demo

image

## Changes

- **`StageNode.types.ts`** — added optional, backward-compatible fields to `stageDetails`: `isOptional`, `endsCase`, plus consumer-supplied `optionalLabel` / `optionalTooltip` / `endsCaseLabel` / `endsCaseTooltip` (all default to off/undefined).
- **`StageNodeHeader.tsx`** — new local `StageStatusBadge` (non-interactive `<span>` pill, optional `CanvasTooltip`); renders the badges in the header status row alongside SLA + `headerChips`. The status row now also shows when only badges are present.
- **`useStageNodeLabels.ts`** — localized default labels (`optionalBadge` → "Optional", `endsCaseBadge` → "Ends case") used as fallback when the consumer doesn't supply copy, so PO.Frontend retains i18n ownership.
- Design tokens: Optional = `bg-background-secondary` + `text-foreground-muted`; Ends case = `bg-error-icon` + `text-foreground-inverse` (theme-aware contrast across light/dark/HC).
- **Tests** — new "Status Badges" suite (default-off, default + consumer labels, both badges, coexistence with SLA/chips, tooltip wiring).
- **Stories** — new `WithStatusBadges` story mirroring the Figma frames (Optional / terminal / combined stages).

## Flow

```mermaid
flowchart TD
    A[stageDetails] -->|isOptional / endsCase| B[StageNodeHeader]
    A -->|optionalLabel / endsCaseLabel| B
    H[useStageNodeLabels] -->|localized fallback| B
    B --> C{flag set?}
    C -->|isOptional| D[StageStatusBadge: Optional]
    C -->|endsCase| E[StageStatusBadge: Ends case]
    D --> F[header status row]
    E --> F
```

## Testing

- [x] `pnpm lint` passes (JS + deps)
- [x] `pnpm typecheck` passes
- [x] `pnpm test` passes (apollo-react: 1703 tests)
- [ ] `pnpm test:visual` — no `test:visual` task in apollo-react locally; runs in CI
- [x] Manual testing / Storybook checked (`WithStatusBadges` story)

# Padding Updates

Stage cards weren't bottom-aligning to the canvas 16px grid. Verified this is not introduced here — for any stage without the new badges, the rendered DOM is byte-identical to main.
Root cause: StageNode is content-sized, and its header chrome was built from non-grid pieces (22px chips, the mt-1 status-row margin, the duration line, etc.), so card heights weren't 16px multiples.
Swept every node type: StageNode was the only offender. BaseNode/LoopNode/GroupNode/BlankCanvasNode/TriggerNode/StickyNote are all already grid-aligned by construction (explicit GRID_SPACING multiples or snap-to-grid on resize).
Fix: made every StageNode vertical segment a 16px multiple

Status row and duration line → 32px (min-h-8) grid-clean rows.
Chips/badges → 24px so each wrapped chip-row is a clean 32px pitch (24 + 8px gap); wrapped lines keep an 8px gap instead of touching.
Tasks already step 48px (3×16); trimmed content bottom padding to land totals on the grid.
Result: card heights are now always multiples of 16 (e.g. 128 / 160 / 208 / 256), so bottoms sit on the dot grid in every config — empty, SLA-only, chips, wrapped, and execution/duration cards.

## Demo 2

image

image

image

image

I checked all stage nodes but there are a few ^

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNode.styles.ts
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNode.styles.ts
@@ -255,7 +255,7 @@ export const StageChip = styled.button`
   justify-content: center;
   gap: ${Padding.PadS};
   box-sizing: border-box;
-  height: 24px;
+  height: ${Spacing.SpacingL};
   padding: 0 ${Spacing.SpacingXs};
   border-radius: 10px;
   border: 1px solid var(--canvas-border-de-emp);
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNode.styles.ts
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNode.styles.ts
@@ -78,7 +78,7 @@ export const StageHeader = styled.div<{ isException?: boolean }>`
 `;
 
 export const StageContent = styled.div`
-  padding: 15px ${STAGE_CONTENT_PADDING_X}px ${Spacing.SpacingBase} ${STAGE_CONTENT_PADDING_X}px;
+  padding: 15px ${STAGE_CONTENT_PADDING_X}px ${Spacing.SpacingS} ${STAGE_CONTENT_PADDING_X}px;
   border-radius: 0 0 ${Spacing.SpacingBase} ${Spacing.SpacingBase};
   overflow: hidden;
   flex: 1;
@@ -254,7 +254,9 @@ export const StageChip = styled.button`
   align-items: center;
   justify-content: center;
   gap: ${Padding.PadS};
-  padding: ${Padding.PadXs} ${Spacing.SpacingXs};
+  box-sizing: border-box;
+  height: 24px;
+  padding: 0 ${Spacing.SpacingXs};
   border-radius: 10px;
   border: 1px solid var(--canvas-border-de-emp);
   background: transparent;
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNode.types.ts
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNode.types.ts
@@ -41,11 +41,14 @@ export enum StageHeaderChipType {
   Exit = 'exit',
   Completion = 'completion',
   ReturnToOrigin = 'returnToOrigin',
+  Optional = 'optional',
+  EndsCase = 'endsCase',
 }
 
 export interface StageHeaderChip {
   type: StageHeaderChipType;
   count?: number;
+  label?: string;
   tooltip?: React.ReactNode;
   onClick?: () => void;
 }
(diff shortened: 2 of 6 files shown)


kittyyueli:
Let's have the top and bottom of the stage always align to the 16px grid: 
image

packages/apollo-react/src/canvas/components/StageNode/StageNode.types.ts
```diff
@@ -65,6 +65,12 @@ export interface StageNodeBaseProps {
     tasks: StageTaskItem[][];
     selectedTaskId?: string;
     headerChips?: StageHeaderChip[];
+    isOptional?: boolean;
+    optionalLabel?: string;
+    optionalTooltip?: React.ReactNode;
+    endsCase?: boolean;
+    endsCaseLabel?: string;
+    endsCaseTooltip?: React.ReactNode;
```
kittyyueli: lets also just have them as headerChips rather than new fields here, you can modify the interface for that to make it work

kittyyueli: hmmm...  or just remove the label and tooltip fields we don't want people passing the text in for those

packages/apollo-react/src/canvas/components/StageNode/StageNode.styles.ts
```diff
@@ -254,7 +254,9 @@ export const StageChip = styled.button`
   align-items: center;
   justify-content: center;
   gap: ${Padding.PadS};
-  padding: ${Padding.PadXs} ${Spacing.SpacingXs};
+  box-sizing: border-box;
+  height: 24px;
```
kittyyueli: Is there a spacing token that is 24px?

packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.tsx
```diff
@@ -26,13 +26,64 @@ const SLA_ICON_CONFIG: Record<StageSlaIcon, { icon: string; iconColor: string }>
...
 
-const CHIP_ICONS: Record<StageHeaderChipType, React.ReactElement> = {
+const CHIP_ICONS: Partial<Record<StageHeaderChipType, React.ReactElement>> = {
   [StageHeaderChipType.Entry]: <EntryConditionIcon w={Icon.IconXs} h={Icon.IconXs} />,
   [StageHeaderChipType.Exit]: <ExitConditionIcon w={Icon.IconXs} h={Icon.IconXs} />,
   [StageHeaderChipType.Completion]: <ChecklistIcon size={16} />,
   [StageHeaderChipType.ReturnToOrigin]: <ReturnToOriginIcon w={Icon.IconXs} h={Icon.IconXs} />,
 };
 
+/** Header-chip types that render as non-interactive status pills instead of the interactive {@link StageChip}. */
```
kittyyueli: I think they are interactive. On click, they should take you to the control on the properties panel. For example, clicking the "Ends case" chip should take you to the case complete rule that the chip is associated with.

kittyyueli (approved):

PR #819, opened 2026-06-14
fix(apollo-react): improve canvas overlay performance [MST-11141]

SreedharAvvari:
## Summary

This PR reduces unnecessary React/render work in the canvas overlay and loop-container paths during high-frequency interactions like dragging, resizing, multi-selecting, and geometry-only updates. The main target is avoidable component work around viewport portals, hidden overlays, loop chrome, sticky notes, and edge animation output.

## What is optimized

- **Node viewport overlays subscribe only to geometry.** `NodeViewportOverlay` now reads just `x`, `y`, `width`, and `height` from the React Flow store with a selector and `shallow` equality. That avoids re-rendering overlay portals when unrelated internal node fields change.
- **Hidden portal content stops mounting.** `HandleButton` skips the viewport portal when neither the add button nor its label is visible. `NodeToolbar` also returns `null` for hidden toolbar instances that are rendered through the node overlay portal, so hidden toolbars do not keep portal wrappers and animation structure alive.
- **Loop success handle labels are removed.** The loop `success` handle still exists for edge connectivity, but it no longer renders the visible `Success` label. This removes a portalized connected-handle label that was showing up during nested-loop drag profiling.
- **Loop shell chrome can bail out during drag.** `BodyFrame`, `ResizeControls`, and `ResizeCornerIndicators` are memoized so static loop chrome does not rerender just because the loop shell rerenders. Resize hit targets remain mounted, so unselected loops can still be resized by hovering the resize controls.
- **Storybook-only static chrome can bail out.** `StoryInfoPanel` is memoized so the story description panel does not rerender during canvas drag profiling.
- **Sticky notes skip position-only churn.** `StickyNoteNode` now uses the position-insensitive node props comparator, so moving nodes does not force sticky note markdown, edit state, resize handles, toolbar config, and color picker UI to re-render unnecessarily.
- **Sticky note overlays render only when actionable.** The sticky note toolbar and color picker overlay now share a single visibility guard and close during drag, resize, and multi-select states. That keeps overlay work out of interaction states where the controls are not usable.
- **Execution edge animations recompute only for in-progress edges.** `useExecutionEdge` still applies status colors for all execution and validation states, but only calls `getStatusAnimation` when the edge status is `InProgress`. Completed, failed, and validation-colored edges no longer rebuild animation output on path-only geometry updates.
- **Loop child detection uses React Flow lookup state.** Loop nodes now check `parentLookup` instead of scanning `state.nodes` to determine whether a loop has children, making that check a map lookup rather than a full node-list scan.
- **The shared node comparator is exported.** `nodePropsEqual` is exported from canvas utils so other canvas nodes can reuse the same position-insensitive memoization path.

## Demo

- This PR primarily reduces avoidable re-renders from `NodeViewportOverlay` / viewport portal.
- We still see expected re-renders from React Flow internals like `NodeWrapper` and `EdgeWrapper`.
- `CanvasEdge`, `SequenceEdge`, `EdgePath`, and `EdgeArrow` still rerender when edge positions change because the SVG geometry has to move with the nodes.

Before vs After

https://github.com/user-attachments/assets/f200508f-4bc9-4582-8811-414cabc03988

https://github.com/user-attachments/assets/a89781f9-fe23-42c2-9f0c-71e05d3a581f

## Expected impact

- Less React work while dragging, selecting, resizing, and moving around larger canvases.
- Fewer hidden portal trees in the viewport overlay layer.
- Less nested-loop drag churn from static loop chrome and connected handle labels.
- Less animation recomputation for status-colored edges when only edge geometry changes.
- Lower sticky note render cost during position updates and multi-node interactions.
- Better scaling for loop-heavy canvases when checking child-node presence.

## Validation

- `pnpm --filter @uipath/apollo-react exec vitest run src/canvas/utils/nodePropsEqual.test.tsx`
- `pnpm --filter @uipath/apollo-react exec vitest run src/canvas/components/LoopNode/LoopNode.test.tsx`
- `pnpm --filter @uipath/apollo-react exec tsc --noEmit --project tsconfig.json`

### DIFF
--- a/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
+++ b/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
@@ -710,11 +710,12 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
           <div className="basis-full pt-0.5 min-w-0 overflow-hidden">{displayFooter}</div>
         )}
       </BaseContainer>
-      {toolbarConfig && !dragging && !multipleNodesSelected && (
+      {toolbarConfig && (
         <NodeToolbar
           nodeId={id}
           config={toolbarConfig}
           expanded={selected || isHovered}
+          hidden={dragging || multipleNodesSelected}
           offsetToolbar={offsetToolbar}
         />
       )}
--- a/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
+++ b/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
@@ -710,12 +710,11 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
           <div className="basis-full pt-0.5 min-w-0 overflow-hidden">{displayFooter}</div>
         )}
       </BaseContainer>
-      {toolbarConfig && (
+      {toolbarConfig && !dragging && !multipleNodesSelected && (
         <NodeToolbar
           nodeId={id}
           config={toolbarConfig}
           expanded={selected || isHovered}
-          hidden={dragging || multipleNodesSelected}
           offsetToolbar={offsetToolbar}
         />
       )}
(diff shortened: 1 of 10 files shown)


snuziale:
@SreedharAvvari are you able to show a before vs. after with react-scan please.

SreedharAvvari:
> @SreedharAvvari are you able to show a before vs. after with react-scan please.

Added, thanks!

snuziale (approved):

PR #823, opened 2026-06-15
feat(apollo-vertex): dashboard card renderers — InsightCardBody and typed card content components

frankkluijtmans:
## Summary

Adds the rendering logic that turns raw data into card content. `InsightCardBody` decides whether a card should show a KPI number, a horizontal bar chart, a donut, a sparkline, or an area chart based on the card's `type` and `chartType` fields. This is the layer between the data model and the visual cards.

Rebased onto `main` — the dashboard foundation and config layer it was previously stacked on have since merged, so this now contains only the new renderer logic (`insight-card-renderers.tsx`).

Supersedes #732, which could not be reopened after its head branch was rebased/force-pushed.

## Test plan
- [ ] No TypeScript errors in `insight-card-renderers.tsx`
- [ ] All card `type` and `chartType` combinations are handled

### DIFF
--- a/apps/apollo-vertex/templates/dashboard/insight-card-renderers.tsx
+++ b/apps/apollo-vertex/templates/dashboard/insight-card-renderers.tsx
@@ -110,8 +110,9 @@ function KpiContent({
 
 function DonutContent({ cardData }: { cardData: InsightCardData }) {
   const circumference = 97.39;
-  // donutPercent is a 0–1 fraction; fall back to the sample value.
-  const fraction = cardData.donutPercent ?? 0.47;
+  // donutPercent is a 0–1 fraction; fall back to the sample value and clamp so
+  // out-of-range data can't produce a negative/oversized arc or a wrong label.
+  const fraction = Math.min(1, Math.max(0, cardData.donutPercent ?? 0.47));
   const label = cardData.donutLabel ?? "funded";
   return (
     <div className="flex items-center justify-center flex-1">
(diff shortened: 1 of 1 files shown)


frankkluijtmans:
Addressed the Copilot review in a follow-up commit (`418b94f3`): keyboard-focusable tooltip trigger, NaN guards on all percentage math, stacked-bar totals/keys now keyed by category index (not CSS color, which repeats past 5 categories), and the donut/sparkline/area renderers now read `donutPercent`/`donutLabel`/`points` from the dataset with sample fallbacks.

Two items I intentionally did **not** change, for the record:

- **"KPI cards ignore `chartType`, donut renderer unreachable"** — this is intentional. `kpi`-type cards render the KPI number by design; their `chartType: "donut"` is consumed for layout styling in `glow-config.ts` (`cardClassName`), not as a renderer selector. Routing those cards to the donut renderer would break the intended KPI display.
- **CodeQL `js/call-to-non-callable` on `useDashboardData()`** — this was a side effect of an import that resolved to `undefined` (since fixed in `97a7d338`). The alert no longer appears on the file, so no defensive guard was added.

Note: `donutPercent` is treated as a 0–1 fraction (matching the prior `0.47 → 47%` sample) since no dataset populates it yet — easy to flip to 0–100 if that was the intent.

ruudandriessen (approved):

angeloaltamiranom (approved):

PR #825, opened 2026-06-15
feat(storybook): add system-wide introduction landing page and unify getting started docs

dbacomputer:
## Summary

- **New landing page** (`Introduction/Overview`) — root-level entry point with three package cards (Apollo Wind, Apollo React, Apollo Core), a GitHub repo card, and a decision table helping users pick the right package. Cards link directly to each package's Getting Started story.
- **Apollo Core Getting Started** — new story with three tabs: Overview (what's included), Quick Start (path switcher: Add to existing project / Run locally, plus JS token imports and CSS custom property usage), and Tokens (full 8-card token category grid with links).
- **Apollo Core sidebar restructure** — all token stories (Colors, Typography, Spacing, etc.) moved into a `Theme/` subfolder for visual consistency with the Apollo Wind and Apollo React sidebar structure.
- **Apollo Wind Getting Started** — restructured from "Choose your package / Contributing / CLI" to **Overview | Quick Start | Components**, matching the Apollo React Canvas format. Quick Start uses the same two-path switcher (Add to existing project / Run locally). React and Core package cards removed since the landing page now handles package selection.
- **Apollo React Canvas Getting Started** — Quick Start updated with the same two-path switcher pattern. GitHub inline link removed.
- **Content cleanup** — all em dash characters (` — `) removed from story copy across all three packages and the introduction page. A no-em-dash rule added to `CLAUDE.md` with replacement guidance for future authoring.

## Test plan

- [ ] Visit `/?path=/story/introduction--overview` — landing page loads with three package cards and GitHub card
- [ ] Click "Explore docs →" on each package card — navigates to the correct Getting Started page (not an iframe URL)
- [ ] Apollo Wind: verify Overview / Quick Start / Components tabs render and path switcher toggles correctly
- [ ] Apollo React Canvas: verify Quick Start path switcher shows Add to existing project / Run locally paths
- [ ] Apollo Core: verify Overview / Quick Start / Tokens tabs, token cards link to correct story paths
- [ ] Apollo Core sidebar: confirm Colors, Typography, etc. appear under `Theme/` folder
- [ ] Confirm no ` — ` characters appear in any documentation copy

Screenshot 2026-06-15 at 3 23 32 PM

### DIFF
--- a/apps/storybook/src/introduction/introduction.stories.tsx
+++ b/apps/storybook/src/introduction/introduction.stories.tsx
@@ -1,0 +1,235 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import * as React from 'react';
+
+const meta = {
+  title: 'Introduction',
+  parameters: {
+    layout: 'fullscreen',
+  },
+} satisfies Meta;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+// ============================================================================
+// Primitives
+// ============================================================================
+
+function Divider() {
+  return <div className="my-10 h-px bg-border" />;
+}
+
+function InlineCode({ children }: { children: React.ReactNode }) {
+  return (
+    <code className="rounded bg-muted px-1.5 py-0.5 font-mono text-sm font-medium text-foreground">
+      {children}
+    </code>
+  );
+}
+
+function Tag({ children }: { children: React.ReactNode }) {
+  return (
+    <span className="inline-flex items-center rounded-full border border-border bg-muted px-2.5 py-0.5 text-xs font-medium text-muted-foreground">
+      {children}
+    </span>
+  );
+}
+
+function SectionHeading({ children }: { children: React.ReactNode }) {
+  return <h2 className="mb-2 text-xl font-semibold tracking-tight text-foreground">{children}</h2>;
+}
+
+function SectionDescription({ children }: { children: React.ReactNode }) {
+  return <p className="mb-6 text-base leading-7 text-muted-foreground">{children}</p>;
+}
+
+// ============================================================================
+// Package card
+// ============================================================================
+
+function PackageCard({
+  name,
+  packageName,
+  description,
+  tags,
+  storyPath,
+  accent,
+}: {
+  name: string;
+  packageName: string;
+  description: string;
+  tags: string[];
+  storyPath: string;
+  accent: string;
+}) {
+  return (
+    <div className="flex flex-col gap-4 rounded-xl border border-border bg-card p-6">
+      <div className="flex flex-col gap-2">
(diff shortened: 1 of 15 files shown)


apps/storybook/src/introduction/introduction.stories.tsx
```diff
@@ -0,0 +1,235 @@
...
+        {/* Hero */}
+        <div className="mb-16">
+          <div className="mb-4 flex items-center gap-2">
+            <Tag>Open Source</Tag>
+            <Tag>UiPath</Tag>
+          </div>
+          <h1 className="mb-4 text-4xl font-bold tracking-tight text-foreground">
+            Apollo Design System
+          </h1>
+          <p className="max-w-2xl text-lg leading-8 text-muted-foreground">
```
BenGSchulz: This `max-w-2xl` doesn't appear to align with the rest of the content

Image

BenGSchulz (commented):
Help to have this section and additional organization, thanks!

Wondering if we want to follow-up with aligning all the "Introduction" stories? They are currently a bit inconsistent. Maybe we want to consider using [MDX](https://storybook.js.org/docs/writing-docs/mdx) so that they appear like documentation rather than stories? (can just have top-level "Getting started" "Overview" or "Prototyping" guides.

Image

apps/storybook/src/introduction/introduction.stories.tsx
dbacomputer: Thanks for the feedback @BenGSchulz  
The intent here is that Introduction lives on its own as a top-level landing page and pointer, it's not tied to any specific library. Each library's Introduction/Getting Started is its own self-contained entry point, so teams that only ever touch one package can land there directly without needing the root page.

Both of these are good follow-ups I have noted and will manage in a separate branch 
-- You're right on consistency though as some older Introduction pages in other sections don't match the tab pattern yet. 
-- On MDX: the tab switchers and interactive "Add to existing project / Run locally" path selector are doing real work for non-technical users, so we'd lose that with MDX. The TSX approach keeps the interactivity while still looking like documentation.

BenGSchulz: I am pretty sure we can still achieve the interactivity / on-boarding components while using MDX for treatment as documentation in storybook.

> [MDX](https://mdxjs.com/) files mix Markdown and Javascript/JSX to create rich interactive documentation. You can use Markdown’s readable syntax (such as # heading) for your documentation, include stories defined in [Component Story Format (CSF)](https://storybook.js.org/docs/api/csf), and freely embed JSX component blocks at any point in the file. All at once.

BenGSchulz (approved):

PR #834, opened 2026-06-18
fix(apollo-react): clear node faces so routed edges exit perpendicular [MST-11337]

KodudulaAshishUiPath:
## Summary
Routed (auto-layout) edges on multi-handle nodes rendered with a kink right at the source/target handle — no perpendicular exit offset. This shifts the bend nearest each node face out to a fixed perpendicular `STUB_OFFSET` so edges leave and enter cleanly.

ELK computes bend points from its port positions, which differ from the rendered React Flow handle positions on multi-handle nodes (Decision true/false, Script main/error). The first/last routed bend lands at or behind the handle, so the edge jogs immediately instead of exiting straight.

## Demo
Before:
Screenshot 2026-06-18 at 10 19 46
After:
Screenshot 2026-06-18 at 10 22 01

## Files Changed
- New `clearNodeFace(waypoints, anchor, position)` in `geometry.ts`: shifts the waypoint riser nearest a node face out to `STUB_OFFSET` (lands exactly `STUB_OFFSET` in front for any gap, including bends behind the face). Applied to both source and target faces in `buildPathVertices`.
- Gated behind a new `autoRouted` flag, threaded `buildPathVertices` → `useEdgeGeometry` → `CanvasEdge`, so it only affects router-produced `routedWaypoints`; manual waypoints render exactly where the user placed them.
- Shifting the stored waypoints (ids preserved) survives `consolidateWaypoints`, unlike a derived stub elbow.

## Tests
New `buildPathVertices` / `clearNodeFace` cases: source + target shift, vertical (y-axis) face, behind-face forward-pull, and manual-not-shifted. 29 geometry / 85 Edges tests pass; biome + tsc clean.

### DIFF
--- a/packages/apollo-react/src/canvas/components/Edges/shared/geometry.ts
+++ b/packages/apollo-react/src/canvas/components/Edges/shared/geometry.ts
@@ -154,6 +154,34 @@ function connectorElbow(from: Point, to: Point, incoming: SegmentOrientation | n
 }
 
 /**
+ * Push the waypoint riser nearest a node face (`anchor` = inset source/target
+ * endpoint) out to `EDGE_CONSTANTS.STUB_OFFSET` so the edge keeps a perpendicular
+ * offset. Any riser closer than that is shifted forward, including bends behind
+ * the face (gap < 0) — common on multi-handle nodes where the router's port sits
+ * behind the rendered handle. Shifting the stored waypoint's own vertex (its
+ * `waypointIndex` is preserved) survives consolidation, whereas a derived elbow
+ * (`waypointIndex === -1`) would be collapsed.
+ */
+function clearNodeFace(waypoints: Waypoint[], anchor: Point, position: Position): Waypoint[] {
+  const dir = getDirection(position);
+  const axis = dir.dx !== 0 ? 'x' : 'y';
+  const sign = dir.dx !== 0 ? dir.dx : dir.dy;
+  const nearest = waypoints.reduce(
+    (acc, w) => (sign > 0 ? Math.min(acc, w[axis]) : Math.max(acc, w[axis])),
+    sign > 0 ? Infinity : -Infinity
+  );
+  // Lands the riser exactly STUB_OFFSET in front of the face for any gap < that,
+  // including bends behind the face (gap < 0) — common on multi-handle nodes
+  // where the router's port sits behind the rendered handle.
+  const gap = (nearest - anchor[axis]) * sign;
+  if (gap >= EDGE_CONSTANTS.STUB_OFFSET) return waypoints;
+  const shift = (EDGE_CONSTANTS.STUB_OFFSET - gap) * sign;
+  return waypoints.map((w) =>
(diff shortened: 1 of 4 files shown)


packages/apollo-react/src/canvas/components/Edges/shared/geometry.ts
```diff
@@ -153,6 +153,32 @@ function connectorElbow(from: Point, to: Point, incoming: SegmentOrientation | n
...
+ * survives consolidation where a derived elbow would not.
+ */
+function clearNodeFace(waypoints: Waypoint[], anchor: Point, position: Position): Waypoint[] {
+  const dir = getDirection(position);
+  const axis = dir.dx !== 0 ? 'x' : 'y';
+  const sign = dir.dx !== 0 ? dir.dx : dir.dy;
+  const nearest = waypoints.reduce(
+    (acc, w) => (sign > 0 ? Math.min(acc, w[axis]) : Math.max(acc, w[axis])),
+    sign > 0 ? Infinity : -Infinity
+  );
```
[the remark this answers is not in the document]
KodudulaAshishUiPath: `clearNodeFace` only runs on auto-routed paths (`autoRouted` guard in `buildPathVertices`; manual waypoints render as-is). Those polylines come back from the router as a monotonic staircase from source to target, so the toward-face extremum along the exit axis *is* the face-adjacent riser for each end. We're keeping the current global min/max for now since there's no repro where a non-adjacent bend gets shifted. If routing ever emits non-monotonic auto-paths we'll switch to selecting the riser nearest the anchor along the face axis (and add a multi-bend test). Tracking this as a follow-up rather than changing routing geometry without a failing case.

uipath-utkarsh (approved):

packages/apollo-react/src/canvas/components/Edges/shared/hooks/useEdgeGeometry.ts
```diff
@@ -29,6 +29,11 @@ export type UseEdgeGeometryArgs = {
   targetPosition: Position;
   /** Manual waypoints — only used in `waypoint` routing. */
   waypoints: Waypoint[];
+  /**
+   * True when `waypoints` are router-produced (not user-placed), so the path
+   * builder may shift bends off the node faces. Manual waypoints render as-is.
+   */
+  autoRouted?: boolean;
```
snuziale, line 40: do you even need this `autoRouted` 
can you just check the waypoints length downstream?

KodudulaAshishUiPath: Downstream we only get effectiveWaypoints, which is already `waypoints.length > 0 ? waypoints : routedWaypoints` — so a non-empty array could be either manual (user-placed) or router-produced. A length check downstream can't tell them apart; autoRouted is what carries that distinction (face-clearance should apply only to router waypoints). We could instead pass both arrays down and check manual.length there, but that pushes the merge into the geometry layer.

snuziale (approved):

PR #842, opened 2026-06-23
fix(apollo-vertex): type solution-tests live queries against react-db

frankkluijtmans:
This PR does two things:

- Amend collection types (noticed issues during implementation)
- Throw when collections are undefined instead of a silent fallback

👨 Generated with Kluijt Code

### DIFF
--- a/apps/apollo-vertex/registry/solution-tests/use-run-results.ts
+++ b/apps/apollo-vertex/registry/solution-tests/use-run-results.ts
@@ -20,6 +20,7 @@ import type {
   SolutionTestJob,
   SolutionTestRunResult,
 } from "./types";
+import { useSolutionTestCollection } from "./use-solution-test-collection";
 
 export interface UseRunResultsResult {
   results: SolutionTestRunResult[];
@@ -44,16 +45,14 @@ function isBaselineJobResult(
 
 /** Live run results for a run, excluding entry-point jobs (role != Baseline). */
 export function useRunResults(runId: string): UseRunResultsResult {
-  const solution = useSolution();
-  const resultsCollection =
-    solution?.api.collections.solutionTests[ENTITY.runResults];
-  const jobsCollection = solution?.api.collections.solutionTests[ENTITY.jobs];
-  const { data, isReady: resultsReady } = useLiveQuery<SolutionTestRunResult>(
-    (q) => (resultsCollection ? q.from({ results: resultsCollection }) : null),
+  const resultsCollection = useSolutionTestCollection(ENTITY.runResults);
+  const jobsCollection = useSolutionTestCollection(ENTITY.jobs);
+  const { data, isReady: resultsReady } = useLiveQuery(
+    () => resultsCollection,
     [resultsCollection],
   );
-  const { data: jobsData, isReady: jobsReady } = useLiveQuery<SolutionTestJob>(
-    (q) => (jobsCollection ? q.from({ jobs: jobsCollection }) : null),
+  const { data: jobsData, isReady: jobsReady } = useLiveQuery(
+    () => jobsCollection,
     [jobsCollection],
   );
   // The role filter needs both collections. Until both are ready, stay loading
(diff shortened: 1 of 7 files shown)


apps/apollo-vertex/registry/solution-tests/use-run-results.ts
```diff
@@ -45,15 +45,21 @@ function isBaselineJobResult(
...
-    (q) => (resultsCollection ? q.from({ results: resultsCollection }) : null),
+  const resultsCollection = solution?.api.collections.solutionTests[
+    ENTITY.runResults
+  ] as Collection<SolutionTestRunResult> | undefined;
+  const jobsCollection = solution?.api.collections.solutionTests[ENTITY.jobs] as
+    | Collection<SolutionTestJob>
+    | undefined;
+  if (!resultsCollection || !jobsCollection) {
+    throw new Error("Solution Tests collections are unavailable.");
+  }
```
ruudandriessen: we repeated this multiple times. Perhaps we just make a `useSolutionTestCollection()` hook?

ruudandriessen (changes requested):

apps/apollo-vertex/registry/solution-tests/use-run-results.ts
ruudandriessen: That likely would also allow for more future configuration of the collection instead - for example if we wanna pass some property into the SolutionProvider indicating which collection to use for these tests

angeloaltamiranom (approved):

apps/apollo-vertex/registry/solution-tests/use-run-results.ts
frankkluijtmans: that's a good one, I'm pretty confident we'll end up having more of these cases

frankkluijtmans: Refactored a little bit, the select function argument is there because we might not always want to select a collection from the same namespace. I also added a name prop so we can add the expected entity name when we throw.

apps/apollo-vertex/registry/use-solution-collection/use-solution-collection.ts
```diff
@@ -0,0 +1,33 @@
...
+
+type SolutionCollections = NonNullable<
+  ReturnType<typeof useSolution>
+>["api"]["collections"];
+
+/**
+ * Resolve the collection picked by `select`, typed as `Collection<TRow>`.
+ * Throws when the solution or the selected collection is unavailable.
+ */
+export function useSolutionCollection<TRow extends object>(
```
ruudandriessen: hmm this sort of seems like we accept any row type - but aren't we making assumptions on the *exact schema* we have? So shouldn't this actually be:

```
  select: (collections: SolutionCollections) => Collection<SolutionTestRun>,
```

I think we are now using it for both SolutionTestRun and SolutionTest - but that makes it actually convoluted and less type safe - relying on the caller to pass in the correct collection rather then just making the select function have the right property here.

ruudandriessen: Like it's a bit odd to use it for *everything* since a better interface is just using `solution.collection.<theoneIwant>` rather then have this as a different alternative to the solution interface

ruudandriessen (changes requested):

apps/apollo-vertex/registry/use-solution-collection/use-solution-collection.ts
frankkluijtmans: Hmm, yeah. Let's just go with a `useSolutionTestCollection` hook where we specifically expect one of the solution test collections 👍  

We'd have something like

```
type SolutionTestRow = {
  [ENTITY.tests]: SolutionTest;
  [ENTITY.runs]: SolutionTestRun;
  [ENTITY.batchRuns]: SolutionTestBatchRun;
  [ENTITY.jobs]: SolutionTestJob;
  [ENTITY.runResults]: SolutionTestRunResult;
};

export function useSolutionTestCollection<K extends keyof SolutionTestRow>(
  entity: K,
): Collection<SolutionTestRow[K]> { … }
```

ruudandriessen (approved):

PR #855, opened 2026-06-25
feat(apollo-wind): add shadow DOM CSS injection utilities

david-rios-uipath:
## Summary

Builds on #830 to consolidate the shadow DOM CSS injection pattern into reusable apollo-wind exports. Every shadow DOM consumer of apollo-wind currently reimplements the same three steps: detect existing CSS, inject a `<style>` tag, and call `registerCssPropertyRules`. This PR extracts that into shared utilities.

### New exports

- **`injectTailwindIntoShadowRoot(root, css)`** — injects a `<style data-tailwind-inject>` into the shadow root, calls `registerCssPropertyRules` internally. Idempotent; returns the created element (or `null` if skipped) so callers can clean up on unmount.
- **`hasApolloWindCss(root)`** — two-level detection: fast-path tag check + computed-style probes (`grid`/`flex`/`hidden`/`relative`). Deliberately avoids probing custom properties (they inherit across the shadow boundary and would false-positive).
- **`TAILWIND_INJECT_ATTR`** — the sentinel attribute constant (`data-tailwind-inject`).

### Backward compatibility

`registerCssPropertyRules` is unchanged. The new functions are additive exports only.

## Test plan

- [x] `pnpm --filter @uipath/apollo-wind test` — 923 tests pass (11 new)
- [x] `pnpm --filter @uipath/apollo-wind build` — new exports appear in `dist/index.d.ts`
- [ ] Publish dev version, install in traceview-ui2, replace local `injectTailwindIntoShadowRoot` and `hasApolloWindCss` with imports from `@uipath/apollo-wind`

### DIFF
--- a/packages/apollo-wind/src/index.ts
+++ b/packages/apollo-wind/src/index.ts
@@ -7,6 +7,11 @@
 // -----------------------------------------------------------------------------
 export { cn } from './lib/utils';
 export { registerCssPropertyRules } from './lib/register-shadow-dom-properties';
+export {
+  injectTailwindIntoShadowRoot,
+  hasApolloWindCss,
+  TAILWIND_INJECT_ATTR,
+} from './lib/shadow-dom-css';
 
 // -----------------------------------------------------------------------------
 // Layout Components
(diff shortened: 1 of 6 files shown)


david-rios-uipath:
Merged into #830 — single PR now ships the full shadow DOM injection API.

PR #866, opened 2026-06-29
feat(apollo-vertex): add ShareWithUiPathCheckbox registry component

RobinMennens-UiPath:
## Summary

Adds a `ShareWithUiPathCheckbox` to the apollo-vertex shadcn registry — a labeled opt-in checkbox for the **"Share feedback data with UiPath to help troubleshoot and improve the product"** consent on Platform Feedback submission surfaces.

Multiple verticals repeat this exact opt-in today (FINS QAQC + LoanSetup, each with an inline `Checkbox` + `Label`), so it belongs in the shared registry. This is the source component for the FINS consumer PR ([fins-vertical-solution#574](https://github.com/UiPath/fins-vertical-solution/pull/574)) — **merge this first**.

## Design

- **Composition over duplication** (per AGENTS.md): builds on the existing `@uipath/checkbox` + `@uipath/label` primitives rather than re-implementing.
- **`label` is a required prop** — consumers plug in their own translated, legal-approved string (same pattern as `feedback-vote-widget`).
- **`size` variant** (`sm` → `text-xs`, `default` → `text-sm`) covers the minor typographic differences between the existing call sites.
- **`className`** for wrapper spacing.
- Collapses Radix's `boolean | "indeterminate"` to a plain boolean so consumers don't have to.
- Self-generated `useId` for the label association.

## Validation

- `shadcn build` emits `r/share-with-uipath-checkbox.json` with the right `registryDependencies`.
- `tsc --noEmit` and `oxlint --type-aware` pass.

### DIFF
--- a/apps/apollo-vertex/registry.json
+++ b/apps/apollo-vertex/registry.json
@@ -2076,6 +2076,19 @@
       ]
     },
     {
+      "name": "share-with-uipath-checkbox",
+      "type": "registry:ui",
+      "title": "Share With UiPath Checkbox",
+      "description": "A labeled opt-in checkbox for sharing feedback data with UiPath — the outer-loop consent control on Platform Feedback submission surfaces.",
+      "registryDependencies": ["@uipath/checkbox", "@uipath/label"],
+      "files": [
+        {
+          "path": "registry/share-with-uipath-checkbox/share-with-uipath-checkbox.tsx",
+          "type": "registry:ui"
+        }
+      ]
+    },
+    {
       "name": "sheet",
       "type": "registry:ui",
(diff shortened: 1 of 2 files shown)


frankkluijtmans (changes requested):
This one doesn't need to go in the registry since it's basically just a regular checkbox. Within a specific vertical client you should de-dupe it though.

RobinMennens-UiPath:
Closing per discussion with @RobinMennens-UiPath — we decided not to extract this into the vertex registry. It's just a checkbox + label with no real customization, so we'll consolidate it as a local reusable component within the FINS clients instead.

PR #867, opened 2026-06-29
fix(apollo-react): compute node height from handle count to stop flicker [MST-11677]

KodudulaAshishUiPath:
## Summary

Fixes the Switch-node height flicker (MST-11677): deleting a case from an unconnected Switch made it oscillate between the 2-handle (96px) and 3-handle (128px) sizes. Root cause was a feedback loop where `BaseNode` derived its height from the *measured* `height` prop and wrote it back via `updateNode` inside a `requestAnimationFrame`.

## Changes

- **`computedHeight` is now a pure function of visible handle count + intrinsic footer/default height** (`getIntrinsicHeight`). It never reads the measured `height` prop, so the value written back can't feed into its own input — the loop is structurally gone.
- **A single, synchronous `useEffect` writes `computedHeight` to `node.height`** (no rAF), only when it differs from the *explicit* `getNode(id)?.height`, then calls `updateNodeInternals`. Comparing against the explicit height (not the measured prop) means a lagging ResizeObserver measurement can't retrigger the write. Convergent and idempotent.
- Keeping `node.height` authoritative fixes the downstream model: measured height, edge anchoring, fit-view, selection bounds, and handle spacing all agree with the rendered body (fixes handles overflowing on high-fan-out nodes).
- `animatedViewportManager` now reads `node.measured?.height ?? node.height ?? …`, matching the package's dominant convention (previously `node.height || 100`, which went stale when height wasn't written).
- Removed the old `baseHeightRef`/`syncedHeightRef` bookkeeping and the dead `getContainerHeight` height param.

## Testing

- [x] `pnpm lint` (Biome) passes
- [x] `pnpm typecheck` passes
- [x] Tests pass — rewritten height suite asserts the write-once/converge behavior, the 128→96 deflation settles with no re-inflation, and no write when already correct (full apollo-react suite green, 1885)
- [ ] `pnpm test:visual` — runs in CI
- [x] No `as any` / type suppressions added

### DIFF
--- a/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
+++ b/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
@@ -82,7 +82,7 @@ const getIntrinsicHeight = (
 };
 
 const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
-  const { type, data, selected, id, dragging, width, height, parentId } = props;
+  const { type, data, selected, id, dragging, width, parentId } = props;
 
   // Read runtime configuration from context (provided by parent node components)
   const {
@@ -519,8 +519,9 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
     hovered: isHovered,
     showAddButton: mode === 'design' && !multipleNodesSelected && !isConnecting && !dragging,
     showNotches,
-    nodeWidth: width,
-    nodeHeight: height,
+    // Deterministic geometry (not the measured props) so handles are grid-aligned from first render.
+    nodeWidth: containerWidth,
+    nodeHeight: computedHeight,
     shouldShowAddButtonFn,
     portalActions: !!parentId,
   });
@@ -561,8 +562,8 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
             id={handle.id}
             defaultPosition={defaultPosition as Position}
             handleType={handleVisualType}
-            nodeWidth={width}
-            nodeHeight={height}
+            nodeWidth={containerWidth}
+            nodeHeight={computedHeight}
             label={handle.label}
             showButton={shouldShowButton}
             selected={selected}
@@ -580,8 +581,8 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
     useSmartHandles,
     handleConfigurations,
     shouldShowHandles,
-    width,
-    height,
+    containerWidth,
+    computedHeight,
     mode,
     selected,
     showNotches,
@@ -727,7 +728,7 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
   // Wrap with SmartHandleProvider only when smart handles are enabled
   if (useSmartHandles) {
     return (
-      <SmartHandleProvider nodeWidth={width} nodeHeight={height}>
+      <SmartHandleProvider nodeWidth={containerWidth} nodeHeight={computedHeight}>
         {nodeContent}
       </SmartHandleProvider>
     );
--- a/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
+++ b/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
@@ -63,11 +63,11 @@ const getContainerWidth = (shape: NodeShape | undefined, width: number | undefin
   return defaultWidth;
 };
 
-const getContainerHeight = (
-  height: number | undefined,
+// Intrinsic height: the fixed footer height when a footer is present, else the default.
+const getIntrinsicHeight = (
   hasFooter: boolean,
   footerVariant: FooterVariant | undefined
-): number | 'auto' => {
+): number => {
   if (hasFooter) {
     switch (footerVariant) {
       case 'button':
@@ -76,13 +76,9 @@ const getContainerHeight = (
         return NODE_HEIGHT_FOOTER_SINGLE;
       case 'double':
         return NODE_HEIGHT_FOOTER_DOUBLE;
-      default:
-        return 'auto';
     }
   }
-  // Use `||` rather than `??` — React Flow can pass `height: 0` before measurement,
-  // and we want that to fall through to the default rather than collapse the node.
-  return height || NODE_HEIGHT_DEFAULT;
+  return NODE_HEIGHT_DEFAULT;
 };
 
 const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
@@ -111,22 +107,13 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
   } = useBaseNodeOverrideConfig();
 
   const updateNodeInternals = useUpdateNodeInternals();
-  const { updateNodeData, updateNode } = useReactFlow();
+  const { updateNodeData, updateNode, getNode } = useReactFlow();
   const containerRef = useRef<HTMLDivElement>(null);
   const [isHovered, setIsHovered] = useState(false);
   const [isFocused, setIsFocused] = useState(false);
 
-  // Tracks the height we last synced to React Flow via updateNode, so we can
-  // distinguish our own handle-based writes from external changes (user-specified
-  // height, React Flow measurement, resize).
-  const syncedHeightRef = useRef<number | undefined>(undefined);
-  // The "base" height — what the node height would be without handle inflation.
-  // Updated when height changes from an external source (not our sync).
-  const baseHeightRef = useRef<number>(DEFAULT_NODE_SIZE);
-
-  if (height && height !== syncedHeightRef.current) {
-    baseHeightRef.current = height;
-  }
+  // Height is driven by computedHeight (below), a pure function of handle count/footer.
+  // It never reads the measured height, so writing it back can't loop.
 
   // Get execution status from external source
   const executionState = useNodeExecutionState(id);
@@ -260,10 +247,9 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
     };
   }, [adornmentsProp, statusContext]);
 
-  // Compute height: max of base height (user-specified or measured) and handle minimum.
-  // baseHeightRef is updated above from external height changes; handle inflation
-  // is computed from the current handleConfigurations.
-  // biome-ignore lint/correctness/useExhaustiveDependencies: height updates baseHeightRef above and intentionally retriggers this memo.
+  // Computed node height: max of the handle-count floor and the intrinsic default/footer
+  // height. Pure (never reads the measured `height`); written to node.height below as
+  // the authoritative size, so node content is expected to fit within it.
   const computedHeight = useMemo(() => {
     const leftHandles = handleConfigurations
       .filter((config) => config.position === Position.Left && config.visible !== false)
@@ -282,31 +268,21 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
     const leftRightHandles = Math.max(leftHandles, rightHandles);
 
     // Each handle gets a 2-grid-space lane (32px), plus 2-grid-space padding at top + bottom of node.
-    const minNodeHeight = (leftRightHandles * 2 + 2) * GRID_SPACING;
-    return Math.max(baseHeightRef.current, minNodeHeight);
-  }, [handleConfigurations, height]);
+    const handleFloor = (leftRightHandles * 2 + 2) * GRID_SPACING;
 
-  // biome-ignore lint/correctness/useExhaustiveDependencies: handle configuration changes require React Flow handle recalculation.
+    return Math.max(getIntrinsicHeight(!!footerComponent, footerVariant), handleFloor);
+  }, [handleConfigurations, footerComponent, footerVariant]);
+
+  // Write computedHeight to node.height and recalculate handle positions. Compare
+  // against node.height (not the measured `height` prop) so a lagging measurement
+  // can't retrigger the write.
+  // biome-ignore lint/correctness/useExhaustiveDependencies: handleConfigurations triggers handle recalculation.
   useEffect(() => {
-    updateNodeInternals(id);
-  }, [handleConfigurations, id, updateNodeInternals]);
-
-  // Sync computed height to node when it differs from React Flow's current value
-  useEffect(() => {
-    if (computedHeight !== undefined && computedHeight !== height) {
-      syncedHeightRef.current = computedHeight;
-      const frameId = requestAnimationFrame(() => {
-        updateNode(id, { height: computedHeight });
-        updateNodeInternals(id);
-      });
-
-      return () => {
-        cancelAnimationFrame(frameId);
-      };
+    if (getNode(id)?.height !== computedHeight) {
+      updateNode(id, { height: computedHeight });
     }
-
-    return undefined;
-  }, [computedHeight, height, id, updateNode, updateNodeInternals]);
+    updateNodeInternals(id);
+  }, [handleConfigurations, computedHeight, id, getNode, updateNode, updateNodeInternals]);
 
   const displayLabel = display.label;
   const displayShape = display.shape ?? 'square';
@@ -320,7 +296,6 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
   // Display customization from props (not data)
   const displayLabelTooltip = labelTooltip;
   const displayLabelBackgroundColor = labelBackgroundColor;
-  const displayFooterVariant = footerVariant;
 
   // SubLabel: Component prop takes precedence, then plain string from display.
   const displaySubLabel = useMemo(() => {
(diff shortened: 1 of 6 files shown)


packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
```diff
@@ -291,11 +291,13 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
...
 
-  // Sync computed height to node when it differs from React Flow's current value
+  // Sync computed height to node when it differs from React Flow's current value.
   useEffect(() => {
     if (computedHeight !== undefined && computedHeight !== height) {
-      syncedHeightRef.current = computedHeight;
       const frameId = requestAnimationFrame(() => {
+        // Stamp at commit time, paired with the write, so the guard above never
+        // mistakes an in-flight stale height for an external resize.
+        syncedHeightRef.current = computedHeight;
```
SreedharAvvari: [P1] This can still bounce back to the stale height if another render lands after this rAF stamps `syncedHeightRef` but before React Flow propagates the new height prop. In that window, the guard above treats the stale `height` as an external resize, restores `baseHeightRef` to the old value, and schedules the old height again.

SreedharAvvari: Suggested change:
Keep the previous synced height ignored until that same value is observed through props, or track a separate pending synced target/source pair so stale post-commit renders cannot rewrite `baseHeightRef`. Please also add coverage for a stale rerender after the rAF callback fires and before the new height prop arrives.

snuziale (approved):

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
```diff
@@ -282,31 +272,21 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
...
-    const minNodeHeight = (leftRightHandles * 2 + 2) * GRID_SPACING;
-    return Math.max(baseHeightRef.current, minNodeHeight);
-  }, [handleConfigurations, height]);
+    const handleFloor = (leftRightHandles * 2 + 2) * GRID_SPACING;
+
+    // Intrinsic floor: fixed footer height when a footer is present, else the default.
+    const intrinsic = getContainerHeight(undefined, !!footerComponent, footerVariant);
+    const intrinsicFloor = typeof intrinsic === 'number' ? intrinsic : NODE_HEIGHT_DEFAULT;
 
+    return Math.max(intrinsicFloor, handleFloor);
```
BenGSchulz: It looks like this somehow affects the handle spacing so some scenarios are breaking:
- Handles become offset from the grid
- Handles grow out of node bounds

You can compare the existing storybook with preview storybook to see behavior differences.

https://github.com/user-attachments/assets/e48dffe9-ca66-47c3-a948-9827c06a3071

KodudulaAshishUiPath: Thanks Ben — addressed in ed95b26d.

Root cause: handle spacing is computed from the `nodeHeight` prop, which is React Flow's *measured* height. When the height write-back was removed, that value was no longer tied to the handle-count floor — and because the stories/nodes set an explicit `node.height` (e.g. the Dynamic Handles story's `height: 96`), the measured height never grew to the floor. So the handles were distributed over the wrong height (off-grid, and overflowing the node bounds on high-fan-out nodes like the 3-input/7-output switch).

Fix: `computedHeight` (a pure function of handle count + footer) is now written back to `node.height`, so React Flow's measured height matches the rendered body and the grid math (`calculateGridAlignedHandlePositions`) gets the correct height. No feedback loop is reintroduced, because `computedHeight` never reads the measured height.

Could you re-check against the preview storybook once the new build is up?

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
```diff
@@ -282,31 +268,21 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
...
-        updateNode(id, { height: computedHeight });
-        updateNodeInternals(id);
-      });
-
-      return () => {
-        cancelAnimationFrame(frameId);
-      };
+    if (getNode(id)?.height !== computedHeight) {
+      updateNode(id, { height: computedHeight });
     }
```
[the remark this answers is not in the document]
KodudulaAshishUiPath: Good catch — the PR description was stale and is now updated. The write-back is intentional and required: React Flow's node model (measured height, edge anchoring, fit-view, selection bounds, and handle spacing) all read `node.height`, so we do need to keep it in sync.

The key change from the previous flickering version is that `computedHeight` is a *pure* function of handle count + footer and never reads the measured `height` prop — so writing it back cannot feed into its own input. That's what breaks the old feedback loop, not the removal of the write. The guard compares against the explicit `node.height` (via `getNode`), not the measured prop, so a lagging ResizeObserver measurement can't retrigger the write.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
```diff
@@ -378,7 +353,7 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
 
     return {
       '--node-w': `${containerWidth}px`,
-      '--node-h': typeof containerHeight === 'number' ? `${containerHeight}px` : 'auto',
+      '--node-h': `${containerHeight}px`,
       '--node-radius': nodeRadius,
```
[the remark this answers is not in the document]
KodudulaAshishUiPath: Correct, and the description is updated to match. Height is deliberately handle/footer-driven, not content-driven: `computedHeight` is written to `node.height` as the authoritative size, and `BaseNodeContainer` applies it as a fixed `h-(--node-h)`. The earlier `min-h-(--node-h) h-auto` experiment was reverted for exactly the consistency reason you note — with an authoritative `node.height`, a fixed height keeps the body, the React Flow wrapper, and the measured height in agreement. Node content (icon/label, or fixed-height footer variants) is expected to fit within the computed height by construction.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
```diff
@@ -282,31 +268,21 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
...
-    const minNodeHeight = (leftRightHandles * 2 + 2) * GRID_SPACING;
-    return Math.max(baseHeightRef.current, minNodeHeight);
-  }, [handleConfigurations, height]);
+    const handleFloor = (leftRightHandles * 2 + 2) * GRID_SPACING;
 
-  // biome-ignore lint/correctness/useExhaustiveDependencies: handle configuration changes require React Flow handle recalculation.
-  useEffect(() => {
-    updateNodeInternals(id);
-  }, [handleConfigurations, id, updateNodeInternals]);
+    return Math.max(getIntrinsicHeight(!!footerComponent, footerVariant), handleFloor);
```
SreedharAvvari, line 273: 🟠 Issue: `computedHeight` now ignores the explicit/stored node height and the effect below writes this floor back to React Flow. A `BaseNode` rendered with `height={200}` and two handles now paints at 96px and updates `node.height` to 96, dropping caller-provided sizing instead of only breaking the measured-height feedback loop.

SreedharAvvari: Suggested change:
Restore a base-height distinction: keep explicit/stored heights as the floor, and only deflate when the stored value was produced by the handle sync. The old `baseHeightRef`/`syncedHeightRef` pattern, or equivalent metadata, would let measured-height lag avoid loops without overwriting caller-provided height.

packages/apollo-react/src/canvas/stores/animatedViewportManager.ts
```diff
@@ -237,8 +237,10 @@ class AnimatedViewportManager {
     const nodeBounds: NodeBounds = {
       x: node.position.x,
       y: node.position.y,
-      width: node.width || 200,
-      height: node.height || 100,
+      // Prefer measured dimensions: node width/height are content-driven and may be
+      // unset in the store until React Flow's ResizeObserver reports them.
+      width: node.measured?.width ?? node.width ?? 200,
```
SreedharAvvari: 🟡 Issue: `??` treats `measured.width === 0` or `measured.height === 0` as a valid size. React Flow can briefly report zero dimensions before measurement; those zeros flow into `calculateNodeFocusViewport`, which divides by width/height and centers on a zero-sized top-left bound instead of using the fallback.

SreedharAvvari: Suggested change:
Use positive-dimension fallbacks here and in `getNodeBounds`, e.g. `node.measured?.width || node.width || 200` and `node.measured?.height || node.height || 100`, or centralize a helper that rejects non-positive dimensions before viewport math.

KodudulaAshishUiPath: Good catch — fixed in 4a3a6a6d. Switched both bounds (here and in `getNodeBounds`) to `||` so a transient `0` measured dimension falls through to the explicit size / default instead of feeding a zero-sized bound into the viewport math.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
```diff
@@ -282,31 +268,21 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
...
+  }, [handleConfigurations, footerComponent, footerVariant]);
 
-  // Sync computed height to node when it differs from React Flow's current value
+  // Write computedHeight to node.height and recalculate handle positions. Compare
+  // against node.height (not the measured `height` prop) so a lagging measurement
+  // can't retrigger the write.
+  // biome-ignore lint/correctness/useExhaustiveDependencies: handleConfigurations triggers handle recalculation.
```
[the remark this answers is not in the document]
KodudulaAshishUiPath: Verified this suppression is required. Removing it makes Biome report `useExhaustiveDependencies` — *"This hook specifies more dependencies than necessary: handleConfigurations"* — and its (unsafe) autofix would drop `handleConfigurations` from the array. It's an intentional extra dependency: the effect body doesn't read `handleConfigurations`, but the effect must re-run `updateNodeInternals` on handle changes that don't move `computedHeight` (e.g. top/bottom handles, or a reshuffle that keeps the same max per-side count). Keeping the ignore.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
```diff
@@ -63,11 +63,11 @@ const getContainerWidth = (shape: NodeShape | undefined, width: number | undefin
...
 };
 
-const getContainerHeight = (
-  height: number | undefined,
+// Intrinsic height: the fixed footer height when a footer is present, else the default.
+const getIntrinsicHeight = (
   hasFooter: boolean,
   footerVariant: FooterVariant | undefined
-): number | 'auto' => {
+): number => {
```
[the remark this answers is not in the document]
KodudulaAshishUiPath: The fallthrough to `NODE_HEIGHT_DEFAULT` is intentional — only the sized footer variants (`button`/`single`/`double`) reserve extra height; every other case (`'none'`, `undefined`, or no footer) uses the default. In practice a footer is always paired with a real variant, so the `hasFooter` + `'none'/undefined` combination isn't reached. Can add an explicit `default` case/comment if you'd prefer it spelled out.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
```diff
@@ -282,31 +268,21 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
...
-        cancelAnimationFrame(frameId);
-      };
+    if (getNode(id)?.height !== computedHeight) {
+      updateNode(id, { height: computedHeight });
     }
-
-    return undefined;
-  }, [computedHeight, height, id, updateNode, updateNodeInternals]);
+    updateNodeInternals(id);
+  }, [handleConfigurations, computedHeight, id, getNode, updateNode, updateNodeInternals]);
```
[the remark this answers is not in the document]
KodudulaAshishUiPath: Done in f331804d. Handle placement now uses the deterministic `computedHeight`/`containerWidth` instead of the measured `height`/`width` props at all three sites (`useButtonHandles`, `SmartHandle`, `SmartHandleProvider`), so handles are grid-aligned from the first render — no percentage→grid transition. `height` is no longer read at all (dropped from the destructure), and a test asserts handles receive the computed height even when the measured prop is a stale small value.

KodudulaAshishUiPath:
@BenGSchulz I think this is safe to merge, I've checked for any external providers setting height in both Flow and apollo-ui. Found none. The current stale heights in .flow files would get re-computed on render and get written by `computedHeight`. 

But there is an issue with stories as Sreedhar reported below, we can't render square nodes of any size other than 96 without handles as setting height isn't possible anymore

SreedharAvvari:
https://apollo-design-qxvi3khj6-uipath.vercel.app/?path=/story/apollo-react-canvas-components-nodes-basenode--sizes story looks broken compared to https://apollo-design.vercel.app/?path=/story/apollo-react-canvas-components-nodes-basenode--sizes

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
```diff
@@ -76,17 +76,13 @@ const getContainerHeight = (
...
       case 'double':
         return NODE_HEIGHT_FOOTER_DOUBLE;
-      default:
-        return 'auto';
     }
   }
-  // Use `||` rather than `??` — React Flow can pass `height: 0` before measurement,
-  // and we want that to fall through to the default rather than collapse the node.
-  return height || NODE_HEIGHT_DEFAULT;
+  return NODE_HEIGHT_DEFAULT;
```
SreedharAvvari, line 81: 🟠 Issue: This removes the content-driven fallback for custom footers. A footer whose variant is `undefined` or `none` previously used `auto`; it now falls through to 96px. `footerComponent` and `footerVariant` remain independently optional in the public `BaseNodeOverrideConfig`, so this is still a valid configuration. Such a footer is now forced into a fixed-height node and its overflow-hidden wrapper can clip content while React Flow records incorrect bounds.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
SreedharAvvari, line 273: 🟠 Issue: This overwrites explicit node heights. `computedHeight` ignores the caller or stored height, then the effect writes the handle/footer floor back to `node.height`. The repository Sizes story creates square, circle, and rectangle nodes with explicit heights; for example, a 128×128 circle with at most two side handles becomes 128×96 and renders as an ellipse. No external provider is required. The rewritten test now explicitly asserts that `height={500}` renders at 96px, so green CI codifies the regression.

SreedharAvvari: Suggested change:
Restore a separate caller-provided base height and calculate `max(baseHeight, intrinsicHeight, handleFloor)`. Only deflate a height known to have been produced by handle synchronization, using explicit sync metadata or equivalent ref tracking. Add a regression test where `getNode(id).height` and the NodeProps height both start at a custom value, then confirm the node returns to that base after the handle count drops.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
SreedharAvvari: Suggested change:
Preserve the prior auto-sizing branch for `hasFooter` with an `undefined` or `none` variant, and do not write a fixed 96px `node.height` for that path. If every footer must have a sized variant, make the configuration discriminated and required in a separately documented breaking change. Add coverage for a custom footer without a sized variant.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
KodudulaAshishUiPath: Honoring the caller/stored height as a floor is what reintroduces the bug this PR fixes. A persisted `node.height` is a resolved value from a prior session, indistinguishable from author intent, so `max(stored, floor)` stops a Switch from deflating when a case is removed. We cannot tell an author-set height from our own resolved output via `node.height`, and shape/width heuristics do not help (Switch is a handle-driven square). So height stays purely computed.

The Sizes-story circles are a Storybook demo artifact (product circles use the default 96x96 and stay round). I will adjust the story and rename the height test so it reads as asserting the computed-height contract rather than codifying a regression. If author-set fixed heights become a real requirement, they need a dedicated channel separate from `node.height`, which we can add once a producer actually exists.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
KodudulaAshishUiPath: `footerComponent` and `footerVariant` are independently optional in the type, but in practice they are always set together. AgentNode is the only producer of footer config, and it pairs a real footer with a sized variant (`button`/`single`/`double`), using `none` only when the footer is `null`. So a `footerComponent` with a `none`/undefined variant never reaches BaseNode, and the 96px fallback is only hit by genuinely footerless nodes where 96 is the correct default. Enforcing the pairing at the type level (a discriminated config) is possible but heavyweight for a case nothing constructs, so I have added a JSDoc note documenting the invariant on the two fields instead.

BenGSchulz (approved):
Thanks for digging through for any custom height needs. I would just like to see the sizes story updated or removed so it doesn't cause confusion.

KodudulaAshishUiPath:
> Thanks for digging through for any custom height needs. I would just like to see the sizes story updated or removed so it doesn't cause confusion.

Yeah sure, will update that

uipath-utkarsh (approved):

PR #870, opened 2026-06-30
feat(apollo-vertex): per-evaluator solution-test result views + IXP comparison

RobinMennens-UiPath:
## What

Makes the solution-tests result view **evaluator-dependent**: each evaluator's result is rendered by a component resolved from its **evaluator id** (the key it's stored under in the `EvaluatorResults` attachment — no schema-sniffing). Adds a rich view for the IXP document-extraction evaluator.

## How

- **`evaluators/registry.tsx`** — `evaluator id → component`. `uipath-json-similarity` and `uipath-llm-judge-output-semantic-similarity` are listed explicitly (they reuse the generic card); `uipath-ixp-document-extraction` → the new IXP view; unknown ids fall back to generic.
- **`evaluators/generic-evaluator-result.tsx`** — the prior default (score + justification + raw expected/actual), extracted so each evaluator owns its own rendering.
- **`evaluators/ixp-extraction/`** — taxonomy-**provenance** bar (highlights a version change, e.g. v24 → v26), a verdict **summary**, and a **per-document collapsible** (clean docs collapsed, with a high-level summary on the row) → per-group field tables showing baseline vs new value with an `identical` / `semantically same` / `different` verdict + the judge's reason; identical fields hide behind a toggle.
- `result-expanded-content.tsx` / `evaluator-results-view.tsx` now dispatch through the registry (per evaluator), so IXP shows provenance + per-field instead of the raw JSON dump.
- Reuses `@/lib/renderValueOrEmptyState` for the empty state; adds `@uipath/collapsible` to `registryDependencies`; registers the new files in `registry.json`; adds `ixp_*` keys to `en.json` (other locales fall back to en).

## Verification

`pnpm typecheck`, `pnpm lint` (type-aware oxlint), and `pnpm registry:build` all clean — solution-tests builds with the new files.

## Consumers

Translation strings aren't pulled by shadcn, so consuming apps (qaqc_webapp, loanSetup) add the `ixp_*` (and runtime `SolutionTest_IXP*` / new-document) keys to their own `en.json`. The FINS backend change that populates the IXP evaluator output (incl. provenance) ships separately in the fins repo.

https://claude.ai/code/session_015YPZsrEKNAi4N8kNtTUZoP


apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/document-row.tsx
```diff
@@ -0,0 +1,120 @@
...
+ * preserving first-seen order. */
+function groupFields(fields: IxpField[]): [string, IxpField[]][] {
+  const map = new Map<string, IxpField[]>();
+  for (const f of fields) {
+    const list = map.get(f.group);
+    if (list) list.push(f);
+    else map.set(f.group, [f]);
+  }
+  return [...map.entries()];
+}
```
0xr3ngar: claude touched this.

you can use [Object.groupBy() ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy)

alone or combine it with Object.entries(Object.groupBy(....))

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/document-row.tsx
```diff
@@ -0,0 +1,120 @@
...
+      {semanticallySame > 0 && (
+        <Badge status="warning" variant="secondary">
+          {t("ixp_count_semantically_same", { count: semanticallySame })}
+        </Badge>
+      )}
+      {different === 0 && semanticallySame === 0 && (
+        <Badge status="success" variant="secondary">
+          {t("ixp_doc_match")}
+        </Badge>
+      )}
```
0xr3ngar: extract the booleans so its easier to read
```ts
 const hasDiff = different > 0;
  const hasSemantic = semanticallySame > 0;
  const isPerfectMatch = !hasDiff && !hasSemantic;
  ```

Comment (0xr3ngar) [reply apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/document-row.tsx]: or even better

```tsx
 const hasDiff = different > 0;
  const hasSemantic = semanticallySame > 0;
  const isPerfectMatch = !hasDiff && !hasSemantic;
  
  const metrics = [
    { show: hasDiff, status: "error", key: "ixp_count_different", args: { count: different } },
    { show: hasSemantic, status: "warning", key: "ixp_count_semantically_same", args: { count: semanticallySame } },
    { show: isPerfectMatch, status: "success", key: "ixp_doc_match" }
  ] as const;

  return (
    <div className="flex items-center gap-1.5">
      {metrics
        .filter(m => m.show)
        .map(({ status, key, args }) => (
          <Badge key={key} status={status} variant="secondary">
            {t(key, 'args' in m ? args : undefined)}
          </Badge>
        ))}

      <span className="text-xs text-muted-foreground">
        {t("ixp_identical_ratio", { identical: doc.identical_count, total: doc.total_fields })}
      </span>
    </div>
  );
  ```

File: apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/document-row.tsx [hunk 3]
```diff
@@ -0,0 +1,120 @@
...
+
+/** A collapsible per-document section. Clean (all-identical) documents start
+ * collapsed; anything with a change or a missing/new status starts open. */
+export const DocumentRow = ({ doc }: { doc: IxpDocument }) => {
+  const different = doc.fields.filter(
+    (f) => f.verdict === IxpVerdict.Different,
+  ).length;
+  const semanticallySame = doc.fields.filter(
+    (f) => f.verdict === IxpVerdict.SemanticallySame,
+  ).length;
```
0xr3ngar: no point in doing 2x filters
you can use groupby or reduce
```ts
const groups = Object.groupBy(doc.fields, (f) => f.verdict);

const different = groups[IxpVerdict.Different]?.length ?? 0;
const semanticallySame = groups[IxpVerdict.SemanticallySame]?.length ?? 0;
```

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/field-group.tsx
```diff
@@ -0,0 +1,95 @@
...
+                    <VerdictBadge verdict={f.verdict} />
+                  </TableCell>
+                </TableRow>
+              ))}
+            </TableBody>
+          </Table>
+        </div>
+      )}
+      {identicalCount > 0 && (
+        <button
```
0xr3ngar: use the Button component? why raw html? I think we have a <Button> component

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/verdict-badge.tsx
```diff
@@ -0,0 +1,35 @@
...
+        </Badge>
+      );
+    default:
+      return (
+        <Badge status="info" variant="outline">
+          {verdict}
+        </Badge>
+      );
+  }
+};
```
0xr3ngar, line 36: you can do this
```tsx
const VERDICT_CONFIG = {
  [IxpVerdict.Different]: { status: "error", labelKey: "ixp_verdict_different" },
 ...
};

export const VerdictBadge = ({ verdict }: { verdict: string }) => {
  
  const config = VERDICT_CONFIG[verdict];

  const children = config ? t(config.labelKey) : verdict;
  
  const status = config?.status ?? "info";

  return (
    <Badge status={status} variant="outline">
      {children}
    </Badge>
  );
};
```

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/ixp-extraction-result.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const { t } = useTranslation();
+
+  let raw: unknown = details;
+  if (typeof raw === "string") {
+    try {
+      raw = JSON.parse(raw) as unknown;
+    } catch {
+      return null;
+    }
+  }
```
0xr3ngar: why do we need this? Just pass it directly to safeParse with (Json.parse...) this try catch and if isn't needed

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/ixp-extraction-result.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const parsed = IxpDetailsSchema.safeParse(raw);
+  if (!parsed.success) return null;
+  const d = parsed.data;
```
0xr3ngar: open up the object, no point in having const d = parsed.data;

const { expected_provenance, actual_provenance } = parsed.data

and consume them directly after...

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/ixp-extraction-result.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <div className="flex flex-col gap-4">
+      <ProvenanceBar
+        expected={d.expected_provenance}
+        actual={d.actual_provenance}
```
0xr3ngar: also why are the values in snake case? and not camelCase?

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
```diff
@@ -0,0 +1,66 @@
+import { z } from "zod";
+import { renderValueOrEmptyState } from "@/lib/renderValueOrEmptyState";
+
+/** The three-way per-field verdict the IXP evaluator emits. */
+export const IxpVerdict = {
+  Identical: "identical",
+  SemanticallySame: "semantically_same",
+  Different: "different",
+} as const;
+export type IxpVerdictValue = (typeof IxpVerdict)[keyof typeof IxpVerdict];
```
0xr3ngar, line 9: why don't we just use an ENUM 🤔 instead of having a type that is exported just so we can see a string?

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/** Render a field's value list (multi-valued fields join with " | "). Empty,
+ * absent, and null values all render the same way ("—") — the distinction
+ * between "no field" and "null value" isn't meaningful to a reviewer. */
+export function formatFieldValues(values: unknown[]): string {
+  const present = values.filter((v) => v != null && v !== "").map(String);
+  return present.length > 0
+    ? present.join(" | ")
+    : renderValueOrEmptyState(null);
+}
```
0xr3ngar: why is this in the schema file 🤔

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
```diff
@@ -0,0 +1,66 @@
...
+export type IxpDetails = z.infer<typeof IxpDetailsSchema>;
+export type IxpDocument = z.infer<typeof DocumentSchema>;
+export type IxpField = z.infer<typeof FieldSchema>;
+export type IxpProvenance = z.infer<typeof ProvenanceSchema>;
+
+/** Render a field's value list (multi-valued fields join with " | "). Empty,
+ * absent, and null values all render the same way ("—") — the distinction
+ * between "no field" and "null value" isn't meaningful to a reviewer. */
+export function formatFieldValues(values: unknown[]): string {
+  const present = values.filter((v) => v != null && v !== "").map(String);
```
0xr3ngar: why do we need to do this check here? 🤔 shouldn't the values already be different than null or "" when called here?

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
```diff
@@ -0,0 +1,66 @@
...
+  Different: "different",
+} as const;
+export type IxpVerdictValue = (typeof IxpVerdict)[keyof typeof IxpVerdict];
+
+/** Per-document comparison status. */
+export const IxpDocStatus = {
+  Compared: "compared",
+  MissingInActual: "missing_in_actual",
+  NewInActual: "new_in_actual",
+} as const;
```
0xr3ngar, line 16: cant this be an enum?

apps/apollo-vertex/registry/solution-tests/evaluators/registry.tsx
```diff
@@ -0,0 +1,38 @@
+"use client";
+
+import type { ComponentType } from "react";
+import type { SolutionTestRunResult } from "../types";
+import { GenericEvaluatorResult } from "./generic-evaluator-result";
+import { IxpExtractionResult } from "./ixp-extraction/ixp-extraction-result";
+
+export const IXP_EXTRACTION_EVALUATOR_ID = "uipath-ixp-document-extraction";
```
0xr3ngar: why is this in const when this isn't used anywhere? but other strings are not extracted as const?

apps/apollo-vertex/registry/solution-tests/evaluators/registry.tsx
```diff
@@ -0,0 +1,38 @@
...
+// is listed explicitly (the deterministic ones reuse GenericEvaluatorResult);
+// unknown ids fall back to it too.
+const EVALUATOR_COMPONENTS: Record<
+  string,
+  ComponentType<EvaluatorResultProps>
+> = {
+  "uipath-json-similarity": GenericEvaluatorResult,
+  "uipath-llm-judge-output-semantic-similarity": GenericEvaluatorResult,
+  [IXP_EXTRACTION_EVALUATOR_ID]: IxpExtractionResult,
+};
```
0xr3ngar, line 71: why can't we use an enum here :thinking:

apps/apollo-vertex/registry/solution-tests/result-expanded-content.tsx
```diff
@@ -83,32 +59,15 @@ export const ResultExpandedContent = ({
...
-        <div>
-          <h4 className="mb-2 text-sm font-semibold">
-            {t("output_evaluator_results")}
-          </h4>
-          <EvaluatorResultsView data={data.evaluatorResults} />
-        </div>
-      )}
-
+      {/* Each evaluator renders its own view (e.g. the IXP comparison shows
+          provenance + per-field verdicts and skips the raw outputs). */}
```
0xr3ngar: why is this comment needed?

apps/apollo-vertex/registry/solution-tests/evaluators/registry.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { GenericEvaluatorResult } from "./generic-evaluator-result";
+import { IxpExtractionResult } from "./ixp-extraction/ixp-extraction-result";
+
+export const IXP_EXTRACTION_EVALUATOR_ID = "uipath-ixp-document-extraction";
+
+export interface EvaluatorResultProps {
+  evaluatorId: string;
+  score: number | undefined;
+  /** Evaluator-specific `details` payload — each component validates it. */
+  details: unknown;
```
0xr3ngar: if it is evaluator specific details, then name the variable accordingly please.

apps/apollo-vertex/registry/solution-tests/evaluators/registry.tsx
```diff
@@ -0,0 +1,38 @@
...
+  /** Raw expected/actual outputs (used by the generic component). */
+  expected: unknown;
+  actual: unknown;
```
0xr3ngar: rename to - expectedOutput, actualOutput

0xr3ngar: also why is it of type unknown? If it is used from a generic component, the component should be able to pass down type T so we can ensure type safety for the consumer?

apps/apollo-vertex/registry/solution-tests/evaluators/output-panels.tsx
```diff
@@ -0,0 +1,24 @@
+"use client";
+
+export function formatJson(data: unknown): string {
+  if (data == null) return "—";
+  if (typeof data === "string") return data;
```
0xr3ngar: what if when the data is of type string is not formatted with 2 spaces?

then the user experience can be different - some renders will be 4/8 spaces, others 2.

apps/apollo-vertex/registry/solution-tests/evaluators/output-panels.tsx
```diff
@@ -0,0 +1,24 @@
+"use client";
+
+export function formatJson(data: unknown): string {
```
0xr3ngar, line 3: again, why is this type unknown? 🤔

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
```diff
@@ -0,0 +1,66 @@
...
+  extractor: z.string().nullish(),
+  tag: z.string().nullish(),
+  project_name: z.string().nullish(),
+});
+
+const FieldSchema = z.object({
+  group: z.string(),
+  field: z.string(),
+  expected: z.array(z.unknown()),
+  actual: z.array(z.unknown()),
```
0xr3ngar, line 33: why is this unknown? why can't we allow the consumer to define the types so we ensure that everything is type safe?

0xr3ngar: also the names are confusing expected/actual? what expected? output? input???

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
```diff
@@ -0,0 +1,66 @@
...
+  verdict: z.string(),
+  reason: z.string().optional().default(""),
```
0xr3ngar: verdict?I would call this evaluationVerdict
reason? this still is too generic make it evaluationVerdictReasoning or something that is more expressive

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
```diff
@@ -0,0 +1,66 @@
...
+  Compared: "compared",
+  MissingInActual: "missing_in_actual",
+  NewInActual: "new_in_actual",
+} as const;
+
+const ProvenanceSchema = z.object({
+  resolved_project_version: z.union([z.number(), z.string()]).nullish(),
+  extractor: z.string().nullish(),
+  tag: z.string().nullish(),
+  project_name: z.string().nullish(),
```
0xr3ngar: this nullish() to me is weird. How come there is a project_name: null??? this makes no sense? each evaluation/extraction should have a project name, right? Correct me if I am wrong, maybe I am missing context

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/provenance-bar.tsx
```diff
@@ -0,0 +1,58 @@
...
+    : String(p.resolved_project_version);
+
+/** Informational banner showing which IXP taxonomy version / extractor produced
+ * the baseline vs the new run, highlighting a version change. */
+export const ProvenanceBar = ({
+  expected,
+  actual,
+}: {
+  expected: IxpProvenance | null | undefined;
+  actual: IxpProvenance | null | undefined;
```
0xr3ngar: this typing goes against the zod schemas that define it as nullish(). 

why do we have null or undefined?

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/provenance-bar.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+import { useTranslation } from "react-i18next";
+import { ArrowRight } from "lucide-react";
+import { Badge } from "@/components/ui/badge";
+import type { IxpProvenance } from "../schema";
+
+const version = (p: IxpProvenance | null | undefined) =>
+  p?.resolved_project_version == null
+    ? null
+    : String(p.resolved_project_version);
```
0xr3ngar: I don't think this wrapper is needed

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
```diff
@@ -1,26 +1,32 @@
...
-  passThreshold: number,
-): string {
-  if (score == null) return "text-muted-foreground";
-  return score >= passThreshold ? "text-green-600" : "text-red-500";
+interface EvaluatorResultsViewProps {
+  data: unknown;
+  expected: unknown;
+  actual: unknown;
+  result: SolutionTestRunResult;
 }
```
0xr3ngar, line 42: again, why do we pass unknown? If the idea is to make the components generic please use type T

0xr3ngar (dismissed):

apps/apollo-vertex/registry/solution-tests/evaluators/registry.tsx
RobinMennens-UiPath: Renamed to expectedOutput / actualOutput.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
RobinMennens-UiPath: Moved out — formatFieldValues now lives in field-group.tsx (its only consumer).

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/ixp-extraction-result.tsx
RobinMennens-UiPath: Done — dropped it; details is already an object at this point, so it goes straight to safeParse.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/registry.tsx
RobinMennens-UiPath: Made consistent — all three evaluator ids are named consts now.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/document-row.tsx
RobinMennens-UiPath: Done — extracted hasDiff / hasSemantic / isPerfectMatch.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/field-group.tsx
RobinMennens-UiPath: Done — switched to <Button variant="link">.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/registry.tsx
RobinMennens-UiPath: Renamed to evaluatorDetails.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/output-panels.tsx
RobinMennens-UiPath: Good catch — it now re-parses and re-stringifies JSON strings at 2 spaces; genuinely non-JSON strings pass through unchanged.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/document-row.tsx
RobinMennens-UiPath: Used a Map here instead of Object.groupBy: the consuming apps target the ES2023 lib and Object.groupBy is ES2024, so it would break their build. It's still a single-pass grouping (no double filter).

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/provenance-bar.tsx
RobinMennens-UiPath: Done — inlined.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/document-row.tsx
RobinMennens-UiPath: Done — the summary chips are now driven off a metrics array.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/result-expanded-content.tsx
RobinMennens-UiPath: Removed.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
RobinMennens-UiPath: The filter is needed: the field-value arrays can contain null (IXP emits null for taxonomy fields it did not extract). Kept it, moved into field-group with a comment explaining that.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/verdict-badge.tsx
RobinMennens-UiPath: Done — replaced the switch with a VERDICT_CONFIG lookup.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/document-row.tsx
RobinMennens-UiPath: Done — counts are computed with a single reduce now (no double filter).

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/ixp-extraction-result.tsx
RobinMennens-UiPath: Done — destructured parsed.data.

https://github.com/UiPath/apollo-ui/pull/870/commits/4d34f994629167db244dff1219c548779ba11ce6

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/ixp-extraction-result.tsx
RobinMennens-UiPath: snake case comes from Python producing the json we are parsing. Not a huge problem IMO

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
RobinMennens-UiPath: Using same convention as we already did in apps\apollo-vertex\registry\solution-tests\types.ts

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
RobinMennens-UiPath: See https://github.com/UiPath/apollo-ui/pull/870#discussion_r3505851342

apps/apollo-vertex/registry/solution-tests/evaluators/registry.tsx
RobinMennens-UiPath: https://github.com/UiPath/apollo-ui/pull/870#discussion_r3505851342

Did extract the consts

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
RobinMennens-UiPath: Did the renames. 
The unknown typing is because we have a generic registry where we map evaluator IDs to components that render the evaluator results. Every evaluator has a different output schema. In the future we might even support verticals defining their own custom evaluators & components with their own schema's. So at this level kept it generic.
At the evaluator level we do have:
`const parsed = IxpDetailsSchema.safeParse(evaluatorDetails);`
`const parsed = DetailsSchema.safeParse(evaluatorDetails)`
So the individual schema's are still verified

apps/apollo-vertex/registry/solution-tests/evaluators/output-panels.tsx
RobinMennens-UiPath: This is basically formatting the raw agent output, for which we have no schema. We also don't care that much about the schema, just want to format the json for rendering

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
RobinMennens-UiPath: I think verdict is clear. Renaming reason to verdictReason

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
RobinMennens-UiPath: These are the actual extracted field values (baseline vs new run) — arbitrary per taxonomy: text, number, date, null — so unknown[] is honest; there's no fixed type to enforce.

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
RobinMennens-UiPath: Kept unknown here on purpose: this is the raw EvaluatorResults attachment (untrusted wire data that can come from many different agents for which we don't know the schema), so it's validated at runtime with zod. The per-evaluator details are typed via the registry now. https://github.com/UiPath/apollo-ui/pull/870/commits/33b07993416231dcd5d6b0ace8335a6a933587b1

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/provenance-bar.tsx
RobinMennens-UiPath: Switched the schema from nullish to nullable, so it's T | null (no undefined). Kept it nullable because a baseline can legitimately have no provenance (older baselines, or a no-document run). https://github.com/UiPath/apollo-ui/pull/870/commits/33b07993416231dcd5d6b0ace8335a6a933587b1

apps/apollo-vertex/registry/solution-tests/evaluators/registry.tsx
RobinMennens-UiPath: Reworked: the registry now pairs each evaluator with its zod schema, validates once centrally, and hands each component a typed evaluatorDetails — no unknown at the component. https://github.com/UiPath/apollo-ui/pull/870/commits/33b07993416231dcd5d6b0ace8335a6a933587b1

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
RobinMennens-UiPath: Renamed reason -> verdict_reason (backend + FE). Kept verdict as-is; it's already unambiguous inside a field's result. https://github.com/UiPath/apollo-ui/pull/870/commits/33b07993416231dcd5d6b0ace8335a6a933587b1

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
RobinMennens-UiPath: Tightened project_name to required — the backend always resolves one (env-default fallback). project_id/tag/version/extractor stay nullish, since those can genuinely fail to resolve. https://github.com/UiPath/apollo-ui/pull/870/commits/33b07993416231dcd5d6b0ace8335a6a933587b1

frankkluijtmans: Can you check this one @RobinMennens-UiPath ? I think it's important to have the API contract properly defined.

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
frankkluijtmans: Agreed, would like this one to be handled too.

apps/apollo-vertex/registry/solution-tests/result-expanded-content.tsx
```diff
@@ -83,32 +59,13 @@ export const ResultExpandedContent = ({
...
-                />
-                <JsonPanel title={t("actual_input")} data={data.actualInput} />
-              </div>
-            )}
-          <div className="grid grid-cols-2 gap-4">
-            <JsonPanel title={t("expected_output")} data={data.expected} />
-            <JsonPanel title={t("actual_output")} data={data.actual} />
-          </div>
-        </div>
+        <EvaluatorResultsView
```
frankkluijtmans: Regression: this returns null when evaluatorResults is null/empty/unparseable, and each renderer also returns null when the details schema fails to parse. So a passed/failed result can now render nothing here. We used to always show the expected/actual output panels. Let's keep an output-panel fallback when there are no valid evaluator entries.

apps/apollo-vertex/registry/solution-tests/evaluators/generic-evaluator-result.tsx
```diff
@@ -0,0 +1,61 @@
...
+            className={`text-sm font-semibold ${scoreColorClass(score, passThreshold)}`}
+          >
+            {scoreStr}
+          </span>
+        </div>
+        {justification && (
+          <p className="mt-2 whitespace-pre-wrap text-xs">{justification}</p>
+        )}
+      </div>
+      <div className="grid grid-cols-2 gap-4">
```
frankkluijtmans, line 55: The inputs display is gone now and showInputs no longer has any reader. If that's intended, drop showInputs from config.ts + resolveConfig and remove the orphaned expected_input / actual_input / output_evaluator_results keys from en.json. Otherwise we need to bring the input panels back.

apps/apollo-vertex/registry/solution-tests/evaluators/generic-evaluator-result.tsx
```diff
@@ -0,0 +1,61 @@
...
+export type GenericEvaluatorDetails = z.infer<
+  typeof GenericEvaluatorDetailsSchema
+>;
+
+function scoreColorClass(
+  score: number | undefined,
+  passThreshold: number,
+): string {
+  if (score == null) return "text-muted-foreground";
+  return score >= passThreshold ? "text-green-600" : "text-red-500";
```
frankkluijtmans: Use semantic tokens here instead of raw palette colors: text-success / text-destructive rather than text-green-600 / text-red-500, to match the status Badges elsewhere in this PR.

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/ixp-extraction-result.tsx
```diff
@@ -0,0 +1,49 @@
...
+        semanticallySame={semantically_same_field_count}
+        identical={identical_field_count}
+      />
+      <div className="flex flex-col gap-2">
+        {documents.length === 0 ? (
+          <p className="text-sm text-muted-foreground">
+            {t("ixp_no_documents")}
+          </p>
+        ) : (
+          documents.map((doc) => <DocumentRow key={doc.document} doc={doc} />)
```
frankkluijtmans: key={doc.document} can collide if two documents share a name. Let's use an index-composite key.

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/field-group.tsx
```diff
@@ -0,0 +1,108 @@
...
+                  {t("ixp_new")}
+                </TableHead>
+                <TableHead className="h-8 px-2 text-right text-xs">
+                  {t("ixp_verdict")}
+                </TableHead>
+              </TableRow>
+            </TableHeader>
+            <TableBody>
+              {visible.map((f) => (
+                <TableRow key={f.field} className="align-top">
```
frankkluijtmans: Same here: key={f.field} can collide across line items. Let's composite it with the index.

frankkluijtmans (changes requested):
A couple of things to sort before this goes in: the passed/failed detail view can now render nothing when there are no valid evaluator entries (we lost the output-panel fallback), and the showInputs config + a few locale keys are now orphaned. Plus two smaller cleanups (semantic color tokens, colliding React keys). Details inline.

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/schema.ts
RobinMennens-UiPath: Verified against the backend: `_resolve_ixp_provenance` always sets `project_name` (resolved from the loan's bank/loan-type config, env-default fallback), so whenever a provenance block exists the key is a present string. When a run has no provenance at all, `_coerce_provenance` returns None and the whole object is null. So the contract is: provenance = null | { project_name: string, resolved_project_version/tag/extractor: nullish } — which is what the schema now encodes.

apps/apollo-vertex/registry/solution-tests/evaluator-results-view.tsx
RobinMennens-UiPath: Done in f5b5a8ff — the attachment is validated at the fetch boundary via an exported `parseEvaluatorResults()`, and `EvaluatorResultsView` now takes a typed `EvaluatorResults` prop instead of `unknown`. The only `unknown` left is that parse function's input, which is the actual trust boundary. https://github.com/UiPath/apollo-ui/pull/870/commits/f5b5a8ff802fd07ed1fa268d1bebe8c3017a65ca

apps/apollo-vertex/registry/solution-tests/result-expanded-content.tsx
RobinMennens-UiPath: Good catch — fixed in f5b5a8ff. A passed/failed row with no parseable evaluator results now falls back to the raw expected/actual output panels, so it never renders empty. https://github.com/UiPath/apollo-ui/pull/870/commits/f5b5a8ff802fd07ed1fa268d1bebe8c3017a65ca

apps/apollo-vertex/registry/solution-tests/evaluators/generic-evaluator-result.tsx
RobinMennens-UiPath: Intended — the details view shows outputs only now. Dropped `showInputs` from config.ts + resolveConfig, the orphaned expected_input/actual_input/output_evaluator_results keys, and the dead ExpectedInput/ActualInput fetch + ExpandedRowData fields. https://github.com/UiPath/apollo-ui/pull/870/commits/f5b5a8ff802fd07ed1fa268d1bebe8c3017a65ca

apps/apollo-vertex/registry/solution-tests/evaluators/generic-evaluator-result.tsx
RobinMennens-UiPath: Done — text-success / text-destructive. https://github.com/UiPath/apollo-ui/pull/870/commits/f5b5a8ff802fd07ed1fa268d1bebe8c3017a65ca

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/ixp-extraction-result.tsx
RobinMennens-UiPath: Done — composite key. Note: our oxlint config bans any array index in keys and the backend gives no stable per-document id; these are static, render-only comparison tables that never reorder, so I used the index-composite with a scoped disable + justification. https://github.com/UiPath/apollo-ui/pull/870/commits/f5b5a8ff802fd07ed1fa268d1bebe8c3017a65ca

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/field-group.tsx
RobinMennens-UiPath: Same fix as the document rows — index-composite key with a scoped oxlint disable (no stable per-row id, static table). https://github.com/UiPath/apollo-ui/pull/870/commits/f5b5a8ff802fd07ed1fa268d1bebe8c3017a65ca

apps/apollo-vertex/registry/solution-tests/evaluators/generic-evaluator-result.tsx
RobinMennens-UiPath: Correction on my earlier reply — the inputs view was actually a useful dev-only debug aid, so we're keeping it, just reworked. Instead of the generic evaluator rendering input panels, `showInputs` now renders the raw Expected/Actual input panels once, alongside the evaluator results, decoupled from any single evaluator. Restored the flag + en.json keys accordingly. https://github.com/UiPath/apollo-ui/pull/870/commits/de8b903030ce8d825cdb4fdf381878ec3f40d57c

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/ixp-extraction-result.tsx
frankkluijtmans: If you use a composite key it's good. I think it's good to realize here that there's a very good reason for oxlint to complain about array index usage here. If there are server side mutations to the list it won't know how to properly re-render based on array index alone.

frankkluijtmans (approved):

frankkluijtmans (approved):

PR #877, opened 2026-07-02
refactor(apollo-vertex): extract solution-tests action errors into errors.ts

frankkluijtmans:
Extracts the app-level failure detection from `create-actions.ts` into a dedicated `errors.ts` module for the Solution Tests registry template. The new module reconciles the backend's translatable `UserMessage` (`{ key, details, message }`) across both the automation-functions RPC contract and the agent contract into a structured `ActionFailure`, thrown as a `SolutionTestActionError` that preserves the i18n `key` and interpolation `details` instead of flattening them to a string. `create-actions.ts` now delegates to `extractFailure`/`SolutionTestActionError`, the new file is registered in `registry.json`, and `SolutionTestActionError`/`ActionFailure` are re-exported from the barrel. This keeps the template in sync with the loan-setup reference app.

👨 Generated with Kluijt Code

### DIFF
--- a/apps/apollo-vertex/registry/solution-tests/errors.ts
+++ b/apps/apollo-vertex/registry/solution-tests/errors.ts
@@ -1,0 +1,95 @@
+"use client";
+
+/**
+ * Structured failure handling for the Solution Tests write actions. The backend
+ * returns a translatable `UserMessage` ({ key, details, message }) on error, as
+ * the top-level `body` on the RPC contract or `output_data.error_message` on the
+ * agent contract. `extractFailure` reconciles both; the actions throw the result
+ * as a `SolutionTestActionError` that keeps `key` + `details` so consumers can
+ * localize with interpolation rather than losing them to a flattened string.
+ */
+
+import { AGENT_SUCCESS_STATUSES } from "./constants";
+
+export interface ActionFailure {
+  /** Stable failure code; consumers may map it to a localized message. */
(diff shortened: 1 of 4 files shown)


apps/apollo-vertex/registry/solution-tests/errors.ts
```diff
@@ -0,0 +1,104 @@
...
+ * `SolutionTestActionError` carrying the `key` + `details` so consumers can
+ * localize with interpolation (e.g. `{{required_roles}}`) themselves, rather
+ * than losing the details to a flattened string.
+ */
+
+import { AGENT_SUCCESS_STATUSES } from "./constants";
+
+export interface ActionFailure {
+  /** Stable i18n key for the failure. */
+  key?: string;
```
ruudandriessen, line 16: not really an i18n key

ruudandriessen (approved):

apps/apollo-vertex/registry/solution-tests/errors.ts
frankkluijtmans: Good catch. `key` also falls back to `output.error`/`output.status`/`"request_failed"`, so it's a stable failure code a consumer *may* map to a translation, not a translation key itself. Reworded the doc comment to say "Stable failure code; consumers may map it to a localized message."

ruudandriessen (approved):

PR #892, opened 2026-07-07
feat(apollo-react): add NodeIOView + JsonTree schema-aware JSON components

BenGSchulz:
## Summary

Adds schema-aware JSON building blocks to `@uipath/apollo-react/canvas`, split into two layers plus a control:

- **`JsonTree`** (`src/canvas/components/JsonTree/`) — the low-level primitives: `buildJsonTree` (schema + value merge, flatten/filter, path helpers), the interactive `JsonTree`, `JsonTreeToolbar`, `JsonTypeBadge`, `JsonValueEditor`, and a `clipboard` helper.
- **`NodeIOView`** (`src/canvas/components/NodeIOView/`) — the composed panel *body* (Schema / JSON tabs + toolbar + tree), plus `PanelTitleBar`. It is domain-agnostic: any schema/value pair works, so it can serve as a general "JSON schema + value" surface, not just node I/O.
- **`NodeOutputModeSelect`** (`src/canvas/controls/NodeOutputModeSelect/`) — a standalone node-execution-mode dropdown.

Extracted from flow-workbench's node Input/Output properties panels, but deliberately de-node-ified.

### `JsonTree` + `NodeIOView`

- **Schema + value merge.** Schema-declared keys render even when the value omits them (as an editable "unset" row); value keys the schema does not declare render too; without a schema the tree is value-driven. The schema contributes descriptions, `enum` choices, `required` markers, and types — there is no schema-conformance/validation layer (mismatch/extra/type-checking was intentionally left out; no current use case).
- **Inline editing:** strings / numbers / null via an inline input; **booleans toggle on a single click**; enums via a dropdown; objects/arrays via a JSON editor (plain textarea, or a consumer-supplied `renderCodeEditor`, e.g. Monaco). Editable value cells get a hover background to signal they are clickable.
- **Navigation:** search, consumer-defined `filters`, expand/collapse (`defaultCollapsedDepth`), and copy path/value. `pathForCopy` lets consumers wrap copied paths (e.g. `$vars.path`); `onCopy` fires only after a *successful* clipboard write.
- **Decoration IoC** via `decorateNode`: `label` (display text replacing the raw key — the key still drives paths/edits), `sublabel`, `badge { icon, className }`, and `chip { label?, icon?, tooltip?, tone? }` (icon-only chips render bare with `role="img"` + tooltip). Search matches decoration labels/sublabels.
- **Extension points:** `renderValue` custom cells (+ `RenderValueContext.commit`/`clear`, clear routes through `removeValueAtPath`), `rowWrapper` (must be a stable component reference), `extraTabs` / `defaultTab` / `onTabChange`, `jsonView` slot, `beforeContent`, `basePath`, and `title*` props.
- **Truncation-safe layout:** custom cells are clipped so they can't paint over the row actions; the tree sublabel and the header badge shrink + truncate so a long id never pushes the actions/controls out of reach; smart tooltips on the header title/badge reveal the full text on hover *only when it is actually truncated*.
- **Neutral `JsonTypeBadge`:** the badge shows just the type icon; the value text carries the per-type color.
- Localized via `useSafeLingui` (`canvas.json_value_panel.*`; catalog keys added to `locales/en.json`, since `lingui extract` does not pick up `_()` descriptors).

### `NodeOutputModeSelect`

- Live / Static mock / Simulated / Skip node dropdown built on `DropdownMenuRadioGroup`.
- `DEFAULT_NODE_OUTPUT_MODES` (English const) + localized `useDefaultNodeOutputModes()` hook (single source of truth so the two can't drift); `options` prop for subsetting/overrides.
- When `value` matches no option the trigger falls back to the first option, and the dropdown's checked/highlight state tracks that same resolved value (no "trigger shows one, list checks none" mismatch).

### `useSafeLingui` hardening

- Renders design-system components correctly with no `I18nProvider` mounted: it pre-compiles the English `message` baked into each descriptor (so plurals/interpolation don't leak as raw `{count, plural, …}` in production builds), and routes ids missing from the host's active catalog through the same compiling fallback.

## Testing

- **96 tests** across `buildJsonTree`, `JsonTree`, `NodeIOView`, `NodeOutputModeSelect`, and `useSafeLingui` (tree build/merge, unset rows, inline edits, boolean toggle, enum dropdown, search/filter subtree behavior, copy, tabs, `basePath`, decorations, truncation).
- `tsc --noEmit` clean, Biome clean.
- Storybook: **`Components/Panels/Node IO` → "Document Extraction (Input / Properties / Output)"** — a three-panel, resizable split-pane (drag the handles to resize neighbors in tandem; fixed height) with controls for read-only mode, input-data variant (schema + value / schema-only / value-only), and the custom Extraction tab. It exercises decorations, a "Referenced in this node" filter, custom file value cells, and a Monaco JSON editor/viewer.

## Notes for reviewers

- First consumer is flow-workbench's flagged node IO panels; consumed via a locally packed tarball until this ships in a release.
- `JsonTreeNode` refers to tree nodes, not flow nodes.


packages/apollo-react/src/canvas/components/JsonTree/buildJsonTree.ts
```diff
@@ -0,0 +1,386 @@
...
+        required: required.has(key),
+      })
+    );
+  }
+
+  // Then value keys the schema does not declare. When `additionalProperties`
+  // is itself a schema it describes those keys (descriptions, enum editing).
+  if (value !== undefined) {
+    for (const [key, childValue] of Object.entries(value)) {
+      if (key in properties) continue;
```
BenGSchulz: **P1: Preserve arbitrary JSON keys safely**

`key in properties` treats inherited names such as `constructor`, `toString`, and `__proto__` as schema declarations, so those valid JSON keys disappear. A direct reproduction returned zero rows for all three keys. The write path also uses `object[head] = ...`, so editing `__proto__` creates prototype-backed data that serializes as `{}`.

Please use own-property checks throughout the exported helpers, including `buildObjectChildren`, `getValueAtPath`, and `removeValueAtPath`, and use an own-data-property write for arbitrary string keys:

```ts
const hasValue = value !== undefined && Object.hasOwn(value, key);
if (Object.hasOwn(properties, key)) continue;

const object: JsonObject = isJsonObject(root) ? { ...root } : {};
const previous = Object.hasOwn(object, head) ? object[head] : undefined;
Object.defineProperty(object, head, {
  value: setValueAtPath(previous, rest, value),
  enumerable: true,
  configurable: true,
  writable: true,
});
```

packages/apollo-react/src/canvas/components/JsonTree/buildJsonTree.ts
```diff
@@ -0,0 +1,386 @@
...
+ * Merges a JSON Schema with a JSON value into a tree.
+ *
+ * Returns the children of the root: schema-declared entries first (rendered
+ * even when the value omits them, as "unset"), then value entries the schema
+ * does not declare. Without a schema the tree is value-driven.
+ */
+export function buildJsonTree(options: BuildJsonTreeOptions): JsonTreeNode[] {
+  const { schema, value, basePath = '' } = options;
+
+  if (Array.isArray(value) || (schema && schemaDisplayType(schema) === 'array')) {
```
BenGSchulz: **P1: Handle scalar and mismatched JSON roots**

`BuildJsonTreeOptions.value` and `NodeIOViewProps.value` accept every `JsonValue`, but scalar values and scalar schemas return zero rows. The schema-driven array condition also wins over an existing non-array value, so a stale array schema can hide an object root completely.

Compute the root type from the runtime value whenever one exists, use the schema only when the value is absent, and emit a synthetic root node for scalar types. Please add regressions for `buildJsonTree({ value: 42 })`, a scalar root schema, and an object value paired with an array schema.

packages/apollo-react/src/canvas/components/JsonTree/JsonValueEditor.tsx
```diff
@@ -0,0 +1,382 @@
...
+  if (type === 'string') return 'string';
+  if (type === 'number') return 'number';
+  return 'auto';
+}
+
+function parseLeafValue(raw: string, editType: LeafEditType): JsonValue | undefined {
+  if (editType === 'string') return raw;
+  if (editType === 'number') {
+    if (raw.trim() === '') return undefined;
+    const parsed = Number(raw);
```
BenGSchulz: **P2: Enforce JSON-safe numbers and integer schemas**

`Number(raw)` accepts `Infinity` and `-Infinity`, which `JSON.stringify` silently converts to `null`. It also accepts fractional values for an `integer` schema, so the schema tree and raw JSON view can disagree after a successful edit.

Keep `integer` as a distinct edit type and validate before committing:

```ts
if (!Number.isFinite(parsed)) return undefined;
if (editType === 'integer' && !Number.isInteger(parsed)) return undefined;
```

BenGSchulz (commented):
Deep review: posting the three selected findings (F1–F3).

packages/apollo-react/src/canvas/components/JsonTree/JsonTree.tsx
```diff
@@ -0,0 +1,947 @@
...
+          {rows.length === 0 && (
+            <p className="py-4 text-center text-xs text-foreground-subtle">
+              {resolvedEmptyMessage}
+            </p>
+          )}
+        </div>
+      </CanvasTooltipProviderMarker>
+    </TooltipProvider>
+  );
+}
```
CalinaCristian, line 236: pretty large file with multiple components, maybe worth splitting?

packages/apollo-react/src/canvas/components/JsonTree/JsonValueEditor.tsx
```diff
@@ -0,0 +1,396 @@
...
+    return parsed;
+  }
+  const trimmed = raw.trim();
+  if (trimmed === '') return null;
+  try {
+    return JSON.parse(trimmed) as JsonValue;
+  } catch {
+    return raw;
+  }
+}
```
CalinaCristian: maybe utils for this?

packages/apollo-react/src/canvas/components/JsonTree/JsonValueEditor.tsx
```diff
@@ -0,0 +1,396 @@
...
+            message: 'Edit JSON of {key}',
+            values: { key: node.key },
+          })}
+        />
+      )}
+      {error && <p className="text-xs leading-4 text-error">{error}</p>}
+      <EditorActions onApply={apply} onCancel={onCancel} />
+    </div>
+  );
+}
```
CalinaCristian: same, multi-components, worth extracting?

packages/apollo-react/src/canvas/components/JsonTree/JsonTree.tsx
BenGSchulz: Done

CalinaCristian (approved):

PR #896, opened 2026-07-09
ci: replace Vercel previews with UiPath Coded App previews

fikewa-olatunji:
## Summary

Deploys apollo-docs, apollo-landing, apollo-design, and apollo-vertex as UiPath Coded Apps via [`@uipath/uip-go`](https://github.com/UiPath/uip-go), replacing the Vercel deployments — both per-PR previews and the stable production sites for merged pages.

All deployment knowledge is owned by this repo, so deploy changes never require a uip-go release:

- **`.uip-go.json`** defines the four deployment recipes (build commands, staged Next.js static exports, pagefind indexing, post-processing, AI Chat platform context).
- **Env-gated app sources**: `APOLLO_CODED_APP=1` switches the docs/vertex next configs to `output: "export"` with a basePath from `APOLLO_CODED_APP_PATH`, and the root/components index pages render the overview inline (static exports cannot redirect).
- **AI Chat runs live in previews and production**: the demo reads `NEXT_PUBLIC_APOLLO_VERTEX_AICHAT_*` (mapped in `next.config.ts` from the `UIP_GO_AICHAT_*` values uip-go resolves at deploy time) to call the platform directly — direct AgentHub/Data Fabric URLs, pinned org/tenant, identity logout origin, and the canonical OAuth redirect path. Dev keeps the same-origin proxy behavior via the existing fallbacks. Without a client id the chat renders a clear disabled state.
- **Preview workflow** (`preview-deploy.yml`): builds and deploys each app per PR, posts a status comment with preview links, auto-registers the AI Chat redirect URI, and cleans up apps + redirect URIs on PR close (checks out the PR head so the recipes are present; app deletion is best-effort due to a platform auth limitation).
- **Production workflow** (`production-deploy.yml`): on every push to `main`, updates the stable Coded Apps of the merged pages (`apollo-docs`, `apollo-landing`, `apollo-design`, `apollo-vertex`) — same recipes and deploy pattern, stable names/paths. The README demo links point at these.
- Turbo hashes `APOLLO_CODED_APP`, `APOLLO_CODED_APP_PATH`, and `UIP_GO_AICHAT_*` so coded-app and regular builds never share cache entries; `pagefind` is a proper devDependency of each app that needs it.

## Verification

- Deploy pipeline verified end-to-end: all four previews deploy Ready and return 200, including the vertex AI Chat page.
- AI Chat verified on a deployed Coded App preview: sign-in with the canonical redirect URI, AgentHub and Conversational Agent modes, and Data Fabric chart tools all work against staging.
- Regular (non-preview) builds are unaffected: coded-app behavior is gated behind env vars that are unset in dev/prod.
- The production workflow reuses the exact deploy pattern of the preview workflow (same recipes, secrets, and uip-go invocation) with stable app names; it runs on merge to `main`.


.github/workflows/preview-deploy.yml
```diff
@@ -0,0 +1,428 @@
...
+            }
+
+  deploy:
+    name: Deploy Apollo Coded App Previews
+    needs: initialize-comment
+    if: ${{ !cancelled() && github.event.action != 'closed' && needs.initialize-comment.result == 'success' && github.event.pull_request.head.repo.fork == false }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: read
```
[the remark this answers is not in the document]
fikewa-olatunji: Agreed this is worth hardening, but it needs repo-settings work that is out of scope for this PR: a GitHub Environment only helps once the UIPATH_* secrets are re-scoped into it by a repo admin, and requiring manual approval would gate every PR preview, which defeats the auto-preview UX. Interim mitigations already in place: both jobs are fork-gated (`head.repo.fork == false`), the workflow is deny-all `permissions: {}` with per-job minimal grants, and the deploy identity is a dedicated External App limited to Apps/Folders/OAuthApp scopes. Tracking the environment-scoping as a follow-up.

.github/workflows/preview-deploy.yml
```diff
@@ -0,0 +1,428 @@
...
+              });
+            }
+
+  cleanup-ai-chat-redirects:
+    name: Cleanup Apollo Coded App Previews
+    if: github.event.action == 'closed' && github.event.pull_request.head.repo.fork == false
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: read
```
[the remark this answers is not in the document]
fikewa-olatunji: Agreed this is worth hardening, but it needs repo-settings work that is out of scope for this PR: a GitHub Environment only helps once the UIPATH_* secrets are re-scoped into it by a repo admin, and requiring manual approval would gate every PR preview, which defeats the auto-preview UX. Interim mitigations already in place: both jobs are fork-gated (`head.repo.fork == false`), the workflow is deny-all `permissions: {}` with per-job minimal grants, and the deploy identity is a dedicated External App limited to Apps/Folders/OAuthApp scopes. Tracking the environment-scoping as a follow-up.

CalinaCristian (commented):
this PR removes the functionality of having previews of merged to develop (production) pages.

We need to maintain this

fikewa-olatunji:
@CalinaCristian good catch, restored in the latest push. Merged pages now stay deployed too: a new `production-deploy.yml` runs on every push to `main` and updates stable Coded Apps of the same four sites (apollo-docs / apollo-landing / apollo-design / apollo-vertex, e.g. https://engdogfood.staging.uipath.host/apollo-vertex/), with AI Chat enabled on vertex. So production pages move from Vercel to Coded Apps along with the PR previews, and the README demo links now point at the stable apps.

apps/apollo-vertex/templates/AiChatTemplate.tsx
```diff
@@ -79,14 +82,32 @@ function AiChatWithConnection({
 }
 
 export function AiChatTemplate() {
+  // Coded App previews without an AI Chat External App configured cannot sign
+  // in, so the demo is disabled instead of failing at the login step.
+  if (AICHAT_IS_CODED_APP && (!AICHAT_DIRECT_BASE_URL || !AICHAT_CLIENT_ID)) {
```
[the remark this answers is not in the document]
fikewa-olatunji: This one is actually correct as written, and the subtlety is now documented in a code comment: coded-app builds define the NEXT_PUBLIC vars unconditionally via next.config's env block (as an empty string when uip-go injected nothing), and `"" ?? fallback` keeps the empty string. So `!AICHAT_CLIENT_ID` does fire the disabled state in un-configured coded-app builds. The non-empty fallback is only reachable in dev, where the env is undefined and `AICHAT_IS_CODED_APP` is false anyway. Verified in the staged-build bundle: with no injection the baked client id is empty and the disabled panel renders.

README.md
```diff
@@ -138,10 +138,12 @@ Framework-agnostic AI chat interface web component for vanilla JS, Vue, Angular,
 
 ## 🎨 Live Demos
 
-Explore our components in interactive Storybook environments:
+Explore our components in interactive Storybook environments. The sites are hosted as UiPath Coded Apps on the staging environment (internal access):
```
CalinaCristian: not internal access, they are public

CalinaCristian (approved):

README.md
fikewa-olatunji: Fixed in [docs(repo): drop inaccurate (internal access) note on demo sites](https://github.com/UiPath/apollo-ui/pull/896/commits/037e5b38c38689122e4ebb930eeb317881531792)

apps/apollo-vertex/templates/ai-chat/ai-chat-example-utils.ts
```diff
@@ -26,7 +77,7 @@ export function createUiPathSdk(
     orgName: orgTenant.orgName,
     tenantName: orgTenant.tenantName,
     clientId: AICHAT_CLIENT_ID,
-    redirectUri: `${origin}/auth_callback`,
+    redirectUri: `${origin}${AICHAT_REDIRECT_PATH}`,
```
ruudandriessen, line 108: seems weird we have to change this? Why make this an env variable?

apps/apollo-vertex/templates/ai-chat/AiChatLoginGate.tsx
```diff
@@ -179,16 +192,19 @@ export function AiChatLoginGate({ children }: AiChatLoginGateProps) {
     );
     logout();
 
+    // In Coded App builds the identity endpoints live on the platform host,
+    // not on the origin serving the static export.
+    const identityOrigin = AICHAT_DIRECT_BASE_URL || window.location.origin;
```
ruudandriessen: No need for this fallback?

apps/apollo-vertex/templates/ai-chat/ai-chat-example-utils.ts
```diff
@@ -9,10 +9,61 @@ export const AICHAT_STORAGE_KEYS = {
...
+// values via next.config.ts. In dev and regular builds they are undefined and
+// the defaults below apply, with platform calls going through the same-origin
+// Next.js proxy routes.
+//
+// The ?? here is load-bearing: coded-app builds define the env vars
+// unconditionally (as "" when uip-go injected nothing), so an empty string
+// survives and the disabled-state check in AiChatTemplate can detect a
+// missing client id. The dev fallbacks only apply when the vars are truly
+// undefined, which never happens in a coded-app build.
+export const AICHAT_CLIENT_ID =
```
ruudandriessen: Why do we have all of these fallbacks here?

apps/apollo-vertex/templates/AiChatTemplate.tsx
```diff
@@ -79,14 +82,32 @@ function AiChatWithConnection({
...
+        </p>
+        <p className="mt-2 max-w-xl text-sm text-muted-foreground">
+          The demo needs an AI Chat External App client id and platform context
+          at build time to call UiPath services directly. This deployment was
+          built without them, so the demo is disabled. The rest of Apollo Vertex
+          is fully available.
+        </p>
+      </div>
+    );
+  }
```
ruudandriessen: We can't just disable this when deploying

ruudandriessen (changes requested):

pieman1313 (changes requested):
logging in to the ai chat doesn't work on previews https://engdogfood.staging.uipath.host/apollo-vertex-pr-896/patterns/ai-chat/ but if I remember correctly they didn't for Vercel previews either due to us not being able to whitelist preview urls in [first party service metadata](https://github.com/UiPath/first-party-service-metadata/blob/8cacfb0a320ffbdd84a1efa0c260189b50de30a5/UiPath.Identity.ClientDefinitions/UiPath.Identity.ClientDefinitions/Definitions/Portal/EnvironmentSpecificConfigs/clienturisconfig-alp.json#L284). perhaps that is doable now, but until then, maybe we should disable it / give a proper message?

fikewa-olatunji:
> logging in to the ai chat doesn't work on previews https://engdogfood.staging.uipath.host/apollo-vertex-pr-896/patterns/ai-chat/ but if I remember correctly they didn't for Vercel previews either due to us not being able to whitelist preview urls in [first party service metadata](https://github.com/UiPath/first-party-service-metadata/blob/8cacfb0a320ffbdd84a1efa0c260189b50de30a5/UiPath.Identity.ClientDefinitions/UiPath.Identity.ClientDefinitions/Definitions/Portal/EnvironmentSpecificConfigs/clienturisconfig-alp.json#L284). perhaps that is doable now, but until then, maybe we should disable it / give a proper message?

It currently works for me. Do you have access to engdogfood org in staging?
Screenshot 2026-07-13 at 11 35 19 PM

pieman1313 (approved):

apps/apollo-vertex/templates/ai-chat/ai-chat-example-utils.ts
fikewa-olatunji: These four (platform base URL, org, tenant, redirect path) are per-deployment values: they differ per environment and per PR and aren't known until deploy time. A Coded App is a static export with no server, so the only way to get a build-time value into the browser bundle is a NEXT_PUBLIC_* env var; there's no runtime config channel. uip-go injects them at build. The two values that are fixed, the client id and scope, are now hardcoded constants (first-party Uber.Client), so those are no longer env vars.

apps/apollo-vertex/templates/ai-chat/AiChatLoginGate.tsx
fikewa-olatunji: Reworked in 2ad1a35a. It wasn't really a fallback; it conflated two modes. Now explicit: `AICHAT_IS_CODED_APP ? AICHAT_DIRECT_BASE_URL : window.location.origin`. A Coded App hits the baked platform host; dev goes through the same-origin proxy, which genuinely needs window.location.origin.

apps/apollo-vertex/templates/ai-chat/ai-chat-example-utils.ts
fikewa-olatunji: Removed in 2ad1a35a. No more `?? ""` defaults. Coded App builds now zod-validate the injected platform context and throw at build time if anything is missing; dev uses explicit proxy values. The client-id/scope defaults became hardcoded constants.

apps/apollo-vertex/templates/AiChatTemplate.tsx
fikewa-olatunji: Agreed, disabled-state removed in 2ad1a35a. A misconfigured Coded App build now fails the build (the zod throw) instead of shipping a disabled chat. And with the first-party Uber.Client plus the redirect-URI whitelist now live in first-party-service-metadata (#4231/#4249), previews actually sign in, so there is nothing to disable. (Also closes @pieman1313's related note.)

fikewa-olatunji:
Thanks @ruudandriessen, addressed the review in 2ad1a35a. Short version: the `??`/`||` fallbacks and the disabled-state are gone; Coded App builds now validate the injected platform context with zod and fail the build if it's missing, and dev vs coded-app is an explicit branch rather than defensive defaults. The one value that stays an env var is the per-deployment platform context (base URL / org / tenant / redirect); explanation is on that thread. Re-requesting your review.

apps/apollo-vertex/next.config.ts
```diff
@@ -4,60 +4,95 @@ const withNextra = nextra({
...
 
+// Coded App preview builds (uip-go) produce a static export served from a
+// sub-path of the Coded App host. There is no server there, so rewrites and
+// headers are dropped and the AI Chat demo talks to the platform directly
+// using the UIP_GO_PLATFORM_AUTH_* values uip-go resolves at build time.
+const codedApp = process.env.APOLLO_CODED_APP === "1";
+const codedAppPath = process.env.APOLLO_CODED_APP_PATH?.replaceAll(
+  /^\/+|\/+$/g,
+  "",
+);
```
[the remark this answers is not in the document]
fikewa-olatunji: Done in 2de7b1c4: the required platform-auth check now runs in next.config (Node, during next build), so a misconfigured Coded App fails the build instead of crashing in the browser.

apps/apollo-vertex/templates/ai-chat/ai-chat-example-utils.ts
```diff
@@ -9,10 +10,90 @@ export const AICHAT_STORAGE_KEYS = {
...
+      tenantName: z.string().min(1),
+      tenantId: z.string().min(1).optional(),
+      redirectPath: z.string().min(1),
+    })
+    .safeParse({
+      baseUrl: process.env.NEXT_PUBLIC_APOLLO_VERTEX_PLATFORM_AUTH_BASE_URL,
+      orgName: process.env.NEXT_PUBLIC_APOLLO_VERTEX_PLATFORM_AUTH_ORG_NAME,
+      tenantName:
+        process.env.NEXT_PUBLIC_APOLLO_VERTEX_PLATFORM_AUTH_TENANT_NAME,
+      tenantId: process.env.NEXT_PUBLIC_APOLLO_VERTEX_PLATFORM_AUTH_TENANT_ID,
```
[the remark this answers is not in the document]
fikewa-olatunji: Done in 2de7b1c4: tenantId is now a required field. uip-go always injects it (its resolver falls back to the tenant name), so requiring it is correct and avoids the empty-string-vs-optional issue.

apps/apollo-vertex/next.config.ts
```diff
@@ -4,60 +4,116 @@ const withNextra = nextra({
...
+
+// A Coded App build calls the platform directly, so uip-go must have injected
+// the platform context. Validate it here — next.config runs in Node during
+// `next build`, so a misconfigured deployment fails the build immediately
+// rather than shipping a bundle that crashes in the browser. (tenantId is
+// optional; it falls back to the tenant name.)
```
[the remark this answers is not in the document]
fikewa-olatunji: Fixed in 048bfa59: corrected the comment; tenantId is required (uip-go always injects it).

.github/workflows/preview-deploy.yml
```diff
@@ -0,0 +1,441 @@
...
+            } else {
+              await github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.issue.number,
+                body,
+              });
+            }
+
+  cleanup-ai-chat-redirects:
```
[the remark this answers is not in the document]
fikewa-olatunji: Fixed in 048bfa59: renamed the job to cleanup-previews, since it only deletes preview apps now.

fikewa-olatunji:
Added commit `3143c7ef`: AI Chat now resolves the signed-in user's org and tenants **at runtime**, instead of baking in the deployment's org/tenant.

**Why:** the previous approach pinned a single org/tenant into the Coded App bundle, so the AI Chat only worked for the deployment's org. A user from any other org got "couldn't load organization info."

**How:** the browser calls the portal `tenantsAndOrganizationInfo` endpoint on the **api host** (which returns `Access-Control-Allow-Origin`, unlike the portal domain) for the caller's **own** org, taken from their token. Each user hits their own org, so it works across any org they belong to. The api host is derived from the platform base URL (`cloud.uipath.com` -> `api.uipath.com`; `<env>.uipath.com` -> `<env>.api.uipath.com`).

**Verified:** deployed to engdogfood staging and signed in from a different org in a real browser; it loads that org's tenants (200 on `staging.api.uipath.com/<org>/portal_/.../tenantsAndOrganizationInfo`).

**Reviewer note (deployment-contract change):** this **removes the baked org/tenant** (`AICHAT_STATIC_ORG` and the `..._ORG_NAME` / `..._TENANT_NAME` / `..._TENANT_ID` env vars). A coded-app build now only needs the base URL and redirect path. There is no pinned fallback, by design, so the org/tenant are always the caller's own.

ruudandriessen (approved):

PR #901, opened 2026-07-10
fix(apollo-react): stage task entry-condition icon sizing and icon-row spacing

BenNormann:
## Summary

Fixes [MST-12000](https://uipath.atlassian.net/browse/MST-12000) plus two follow-up spacing/sizing issues design flagged in review. In a stage node, the entry-condition diamond rendered 20px on tasks that hadn't run and 16px once they had, so diamonds in the same stage visibly differed. Per design (Kitty Li), the size should follow the card layout instead: **20px on single-line cards regardless of execution status, 16px only when execution details add a second line**. Separately, the play button read as both off-size and over-spaced next to the status icons; the root cause was subtle (see below), and the fix deletes more sizing code than it adds.

## Demo
image

## Root cause of the play-button issues

The apollo-wind `Button` base class applies `[&_svg]:size-4`, forcing **every descendant svg to render 16×16 regardless of its width/height attributes**. `TaskPlayButton`'s `iconSize`/`buttonSize` math (18/22px icons in 20/24px buttons) was therefore dead code — the icon always rendered 16px, and the oversized button box plus the icon's padded 24×**25** viewBox produced invisible horizontal padding and a ring ~1px smaller than the lucide status circles (whose stroked circles reach outer radius 11 on the same 24-unit grid).

## Changes

- `StageTaskEntryConditionIcon` (via `TaskContent`): first-row diamond no longer passes `small={!!hasExecutionStatus}` — single-line cards always get the 20px diamond; the second-row 16px call is unchanged.
- `TaskContent`: `gap={Spacing.SpacingXs}` on the first row so a truncated label (and its required `*`) never sits flush against the status icon.
- `TimelinePlayIcon`: normalized to a square 24-unit viewBox centered on the ring, with the glyph scaled 1.05 so its filled ring optically matches the lucide status circles' stroke band. (Long-term this should be a re-exported asset from design.)
- `TaskPlayButton`: now a plain 16×16 ghost Button with the exact classes of the status-icon button (`h-4 w-4`) — alignment and even gaps come from identical geometry instead of the removed negative-margin hacks. A 14px `small` variant (`[&_svg]:size-3.5` override, same pattern as `QuickActionsRow`) shadows the diamond's shrink on the execution second line.
- Renamed `showPlayButtonSmall` → `showReplayButton` — it gates the replay button for already-run tasks and no longer implies a size.

Note: `TimelinePlayer` (agent canvas) uses the same icon; its ring changes imperceptibly (~12.8 → 13.4px) and stays within a pixel of its `TimelinePause` sibling.

## Flow

```mermaid
flowchart TD
    TC[TaskContent] --> Q{execution details<br/>add a second line?}
    Q -->|no - single line| A[diamond 20px + play 16px<br/>matching 16px status button]
    Q -->|yes - two lines| B[second row: diamond 16px + play 14px]
```

## Testing

- [x] `pnpm lint` passes
- [x] `pnpm typecheck` passes
- [x] `pnpm test` passes (apollo-react: 107 files, 1915 tests; StageNode + AgentCanvas re-run after each change)
- [ ] `pnpm test:visual` — no local task; runs in CI
- [x] Manual testing / Storybook checked (`StageNode` → `Ad hoc Tasks` story, "Mixed with Parallel" stage with Completed/Failed/not-run entry-condition tasks; sizes and gaps verified against status icons with design)

[MST-12000]: https://uipath.atlassian.net/browse/MST-12000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/StageTaskEntryConditionIcon.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageTaskEntryConditionIcon.tsx
@@ -21,6 +21,8 @@ export const StageTaskEntryConditionIcon = ({
           justifyContent: 'center',
           color: 'var(--color-icon-default)',
           flexShrink: 0,
+          // Optically centre the diamond; the adjacent play button sits inset (more so when small).
+          transform: small ? 'translateX(1px)' : 'translateX(0.5px)',
         }}
       >
         <EntryConditionIcon w={small ? 16 : 20} h={small ? 16 : 20} />
--- a/packages/apollo-react/src/canvas/components/StageNode/StageTaskEntryConditionIcon.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageTaskEntryConditionIcon.tsx
@@ -22,6 +22,7 @@ export const StageTaskEntryConditionIcon = ({
           color: 'var(--color-icon-default)',
           flexShrink: 0,
           // Optically centre the diamond; the adjacent play button sits inset (more so when small).
+          // Value is in canvas-local px, so it renders ×canvas zoom — roughly 1–3px on screen in practice.
           transform: small ? 'translateX(1px)' : 'translateX(0.5px)',
         }}
       >
--- a/packages/apollo-react/src/canvas/components/StageNode/TaskContent.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/TaskContent.tsx
@@ -2,9 +2,9 @@ import { Padding, Spacing } from '@uipath/apollo-core';
 import { Column, Row } from '@uipath/apollo-react/canvas/layouts';
 import { Badge, Button, Spinner } from '@uipath/apollo-wind';
 import debounce from 'debounce';
+import { CirclePlay } from 'lucide-react';
 import { memo, useCallback, useMemo, useState } from 'react';
 import { useSafeLingui } from '../../../i18n';
-import { TimelinePlayIcon } from '../../icons';
 import { CanvasTooltip } from '../CanvasTooltip';
 import { ExecutionStatusIcon } from '../ExecutionStatusIcon';
 import { StageTaskIcon, StageTaskRetryDuration } from './StageNode.styles';
@@ -82,7 +82,7 @@ const TaskPlayButton = memo(
             padding: 0,
           }}
         >
-          {playLoading ? <Spinner size="sm" /> : <TimelinePlayIcon />}
+          {playLoading ? <Spinner size="sm" /> : <CirclePlay />}
         </Button>
       </CanvasTooltip>
     );
--- a/packages/apollo-react/src/canvas/components/StageNode/TaskContent.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/TaskContent.tsx
@@ -61,9 +61,6 @@ const TaskPlayButton = memo(
       [debouncedTaskPlay, taskId]
     );
 
-    const iconSize = small ? 18 : 22;
-    const buttonSize = small ? '20px' : Spacing.SpacingL;
-
     return (
       <CanvasTooltip content="Trigger task" placement="top">
         <Button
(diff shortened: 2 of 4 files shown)


packages/apollo-react/src/canvas/components/StageNode/StageTaskEntryConditionIcon.tsx
```diff
@@ -21,6 +21,8 @@ export const StageTaskEntryConditionIcon = ({
           justifyContent: 'center',
           color: 'var(--color-icon-default)',
           flexShrink: 0,
+          // Optically centre the diamond; the adjacent play button sits inset (more so when small).
+          transform: small ? 'translateX(1px)' : 'translateX(0.5px)',
```
kittyyueli, line 26: appreciate the detail but its a little overkill for less than 1px

packages/apollo-react/src/canvas/icons/TimelinePlayIcon.tsx
kittyyueli, line 1: Is there a lucide icon that can replace this?

packages/apollo-react/src/canvas/components/StageNode/StageTaskEntryConditionIcon.tsx
BenNormann: Fair, though it's ×canvas zoom though, so ~1–3px on screen in practice

packages/apollo-react/src/canvas/components/StageNode/TaskContent.tsx
```diff
@@ -127,7 +122,7 @@ export const TaskContent = memo(
     }, [taskExecution?.badge, taskExecution?.retryCount, taskExecution?.status, _]);
     const hasSecondRowContent =
       taskExecution && (taskExecution.duration || taskExecution.retryDuration || badgeText);
-    const showPlayButtonSmall = onTaskPlay && hasExecutionStatus;
+    const showReplayButton = onTaskPlay && hasExecutionStatus;
```
kittyyueli: let's rename to showTaskPlayButton

packages/apollo-react/src/canvas/icons/TimelinePlayIcon.tsx
BenNormann: lucide's `CirclePlay`. Only diff is it's stroke-based (hollow triangle) vs our solid-fill one, so a slightly lighter look. Should we go with that one?

packages/apollo-react/src/canvas/components/StageNode/StageTaskEntryConditionIcon.tsx
kittyyueli: I think its okay~ all our icons will have this issue

packages/apollo-react/src/canvas/icons/TimelinePlayIcon.tsx
kittyyueli: yep, lets do that and delete this file. You can use the CanvasIcon component for the lucide icons

packages/apollo-react/src/canvas/components/StageNode/TaskContent.tsx
```diff
@@ -73,21 +70,19 @@ const TaskPlayButton = memo(
           onClick={handlePlayClick}
           onMouseDown={(e: React.MouseEvent) => e.stopPropagation()}
           onKeyDown={(e: React.KeyboardEvent) => e.stopPropagation()}
-          className="task-menu-icon-button"
+          aria-label="Trigger task"
```
kittyyueli: use translated string

kittyyueli (approved):

alextran1 (approved):

kittyyueli (approved):

PR #925, opened 2026-07-16
feat(apollo-react): don't change node shape when collapsed [MST-12493]

Holychung:
## Problem
`resolveDisplay` squared a rectangle node whenever it was collapsed (`getCollapsedShape`). Agents were the only collapsible nodes, and design (UiPath/flow-workbench MST-12167) wants a collapsed agent to **keep its footprint** — only hide artifacts and show the stacked affordance, not shrink to a square.

## Change
- `resolveDisplay` no longer transforms shape on collapse — the shape is kept (resolved from instance override or manifest) whether expanded or collapsed. `isStacked` and artifact-handle hiding are unchanged.
- Removed the now-dead collapse-shape helpers: `getCollapsedShape`, `getCollapsedSize`, `getExpandedShape`.
- Renamed `collapse.ts` → `node-size.ts` — the file now only holds `getExpandedSize` (shape → default node size, used by add-node alignment + layout), which is **not** collapse-related.

## ⚠️ BREAKING CHANGE
`getCollapsedShape`, `getCollapsedSize`, and `getExpandedShape` are removed from the public `@uipath/apollo-react/canvas` API. They only supported the collapse↔shape tying being removed. The only in-tree consumer (flow-workbench) is updated in tandem to drop them. `getExpandedSize` + the size constants remain (re-exported from `node-size`). → should land as a **major** version bump.

## Tests / storybook
Resolver tests updated (collapsed shape unchanged; instance override honored). BaseNode → "Stacked Treatment" demos a collapsed rectangle staying a wide pill + stacked. Verified end-to-end via a local link into flow-workbench (collapsed agent = pill + stacked).

Driven by UiPath/flow-workbench MST-12167.


packages/apollo-react/src/canvas/utils/collapse.ts
```diff
@@ -18,14 +18,6 @@ export const COLLAPSED_NODE_SIZE = 96;
  */
 export const EXPANDED_RECTANGLE_WIDTH = 288;
 
-/**
```
akraminakib, line 21: I don't think we need this file at all anymore since we no longer have to deal with shape-related logic on collapse/expand?

Holychung: Only `getCollapsedShape` was shape-squaring. The rest is size helpers: `getExpandedSize` is used internally (AddNodeManager.helpers, useCanvasNodeLayout), and `getCollapsedSize` + `getExpandedShape` are public API that flow-workbench consumes (conversion / replace-node / swap-node). So the file needs to stay.

packages/apollo-react/src/canvas/utils/index.ts
```diff
@@ -4,7 +4,7 @@ export * from './CanvasEventBus';
 export * from './CssUtil';
 export * from './coded-agents/d3-layout';
 export * from './coded-agents/mermaid-parser';
-export * from './collapse';
+export * from './node-size';
 export * from './constraint-validator';
```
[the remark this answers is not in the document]
Holychung: Intentional — going with a breaking removal rather than re-adding the dead code behind a shim. Added a BREAKING CHANGE: footer + a note in the description so it lands as a major bump. flow-workbench (the only in-tree consumer) drops all three in tandem; getExpandedSize + constants stay (re-exported from node-size).

snuziale (approved):

PR #926, opened 2026-07-17
feat(apollo-wind): add autoGrow in textArea [MST-12041]

ayush578:
## Summary

Adds opt-in **auto-grow** to the apollo-wind `Textarea` via new `minRows` / `maxRows` props (mirroring how `ApTextArea` in apollo-react exposes them, but implemented natively — apollo-wind has no MUI dependency).

### Behavior
- **Backward compatible:** with neither prop set, the component is byte-for-byte the same plain `<textarea>` as before.
- **Auto-grow:** floors at `minRows`, grows with content up to `maxRows`, then scrolls.
- **Uncapped manual resize:** the ceiling is enforced in JS (`height`), never via CSS `max-height`, so the bottom-right resize handle can be dragged past `maxRows`. A `ResizeObserver` detects the drag and hands sizing control to the user.
- Row heights are derived from the element's *computed* line-height + padding/border, so sizing is correct across `text-base`/`text-sm` and all themes.

### demo

https://github.com/user-attachments/assets/0c3b1957-6827-42f1-a439-c78c9bf464e8

Note: jsdom has no layout engine, so the tests cover the structural guarantees; the pixel-level grow/scroll is best confirmed in the new Storybook `AutoGrow` story.

### DIFF
--- a/packages/apollo-wind/src/components/ui/textarea.stories.tsx
+++ b/packages/apollo-wind/src/components/ui/textarea.stories.tsx
@@ -49,6 +49,25 @@ export const WithRows: Story = {
   },
 };
 
+export const AutoGrow = {
+  render: () => (
+    <div className="grid w-full max-w-sm gap-1.5">
+      <Label htmlFor="autogrow">Auto-growing field</Label>
+      <Textarea
+        id="autogrow"
+        minRows={3}
+        maxRows={7}
+        placeholder="Starts at 3 rows and never shrinks below that. Grows with content up to 7 rows, then scrolls. Drag the bottom-right handle to make it taller than 7 rows."
+      />
+      <p className="text-sm text-muted-foreground">
+        <code>{'minRows={3}'}</code> is a floor: the field never renders or resizes shorter than 3
+        rows. <code>{'maxRows={7}'}</code> auto-grows with content up to 7 rows, then scrolls. The
+        resize handle is not capped, so you can drag past 7 rows to take manual control.
+      </p>
+    </div>
+  ),
+} satisfies Story;
+
 export const WithDescription = {
   render: () => (
     <div className="grid w-full max-w-sm gap-1.5">
--- a/packages/apollo-wind/src/components/ui/textarea.test.tsx
+++ b/packages/apollo-wind/src/components/ui/textarea.test.tsx
@@ -149,4 +149,28 @@ describe('Textarea', () => {
     const textarea = screen.getByRole('textbox');
     expect(textarea).toHaveValue('Default text');
   });
+
+  describe('auto-grow structure (minRows / maxRows)', () => {
+    it('stays a plain textarea when neither minRows nor maxRows is set', () => {
+      render(<Textarea aria-label="Message" />);
+      const textarea = screen.getByRole('textbox') as HTMLTextAreaElement;
+      expect(textarea).not.toHaveClass('resize-y');
+      expect(textarea.style.minHeight).toBe('');
+      expect(textarea.style.height).toBe('');
+    });
+
+    it('auto-grows and enables the resize handle once maxRows is set', () => {
+      render(<Textarea maxRows={5} aria-label="Message" />);
+      const textarea = screen.getByRole('textbox') as HTMLTextAreaElement;
+      expect(textarea).toHaveClass('resize-y');
+    });
+
+    it('still forwards onChange while auto-growing', async () => {
+      const user = userEvent.setup();
+      const handleChange = vi.fn();
+      render(<Textarea minRows={3} maxRows={5} onChange={handleChange} aria-label="Message" />);
+      await user.type(screen.getByRole('textbox'), 'hi');
+      expect(handleChange).toHaveBeenCalled();
+    });
+  });
 });
--- a/packages/apollo-wind/src/components/ui/textarea.tsx
+++ b/packages/apollo-wind/src/components/ui/textarea.tsx
@@ -2,20 +2,189 @@ import * as React from 'react';
 
 import { cn } from '@/lib/index';
 
-export type TextareaProps = React.ComponentProps<'textarea'>;
+export type TextareaProps = React.ComponentProps<'textarea'> & {
+  /**
+   * Minimum number of visible rows.
+   */
+  minRows?: number;
+  /**
+   * Enables auto-grow and sets the ceiling, in rows. The field grows with its content
+   * up to `maxRows`, then scrolls.
+   */
+  maxRows?: number;
+};
+
+/** Coerce a row count to a positive integer, or `undefined` when unusable. */
+function normalizeRows(value: number | undefined): number | undefined {
+  if (value === undefined || !Number.isFinite(value)) return undefined;
+  return Math.max(1, Math.floor(value));
+}
+
+interface RowMetrics {
+  lineHeight: number;
+  padding: number;
+  border: number;
+  borderBox: boolean;
+}
+
+/** Reads the per-row line height and vertical padding/border of a textarea. */
+function measureRowMetrics(el: HTMLTextAreaElement): RowMetrics {
+  const styles = window.getComputedStyle(el);
+  const lineHeight = parseFloat(styles.lineHeight) || parseFloat(styles.fontSize) * 1.2 || 20;
+  const padding = parseFloat(styles.paddingTop) + parseFloat(styles.paddingBottom) || 0;
+  const border = parseFloat(styles.borderTopWidth) + parseFloat(styles.borderBottomWidth) || 0;
+  return {
+    lineHeight,
+    padding,
+    border,
+    borderBox: styles.boxSizing === 'border-box',
+  };
+}
+
+/** Serialize a React style value to a CSS string (numbers become px), or '' when unset. */
+function cssLength(value: string | number | undefined | null): string {
+  if (value === undefined || value === null) return '';
+  return typeof value === 'number' ? `${value}px` : String(value);
+}
 
 const Textarea = React.forwardRef<HTMLTextAreaElement, TextareaProps>(
-  ({ className, ...props }, ref) => {
+  ({ className, minRows, maxRows, onChange, style, ...props }, ref) => {
+    // Normalize + guard inverted bounds: a floor taller than the ceiling would trip the
+    // manual-resize detector and permanently disable auto-sizing (min never exceeds max).
+    const { effMinRows, effMaxRows } = React.useMemo(() => {
+      const nMin = normalizeRows(minRows);
+      const nMax = normalizeRows(maxRows);
+      return {
+        effMinRows: nMin !== undefined && nMax !== undefined ? Math.min(nMin, nMax) : nMin,
+        effMaxRows: nMax,
+      };
+    }, [minRows, maxRows]);
+
+    const managed = effMinRows !== undefined || effMaxRows !== undefined;
+
+    const innerRef = React.useRef<HTMLTextAreaElement | null>(null);
+    const manuallyResized = React.useRef(false);
+    // Keep the latest consumer style available to imperative code without re-subscribing.
+    const styleRef = React.useRef(style);
+    styleRef.current = style;
+
+    const setRef = React.useCallback(
+      (node: HTMLTextAreaElement | null) => {
+        innerRef.current = node;
+        if (typeof ref === 'function') ref(node);
+        else if (ref) ref.current = node;
+      },
+      [ref]
+    );
+
+    // Restore the inline styles auto-grow owns to the consumer's declared values (or clear
+    // ours). React doesn't track these because we set them imperatively, so switching modes
+    // would otherwise leave stale height/overflow behind.
+    const restoreOwnedStyles = React.useCallback((el: HTMLTextAreaElement) => {
+      const consumer = styleRef.current;
+      el.style.height = cssLength(consumer?.height);
+      el.style.minHeight = cssLength(consumer?.minHeight);
+      el.style.overflowY = cssLength(consumer?.overflowY);
+    }, []);
+
+    // Apply the `minRows` floor and, when `maxRows` is set, fit the height to the content up
+    // to the ceiling (scrolling past it). `scrollHeight` excludes the border, so add it back
+    // in border-box mode; content-box keeps the content-only calculation.
+    const resize = React.useCallback(() => {
+      const el = innerRef.current;
+      if (!el || !managed) return;
+
+      const { lineHeight, padding, border, borderBox } = measureRowMetrics(el);
+      const extra = borderBox ? padding + border : 0;
+      const minHeight = effMinRows !== undefined ? effMinRows * lineHeight + extra : 0;
+
+      el.style.minHeight =
+        effMinRows !== undefined ? `${minHeight}px` : cssLength(styleRef.current?.minHeight);
+
+      if (effMaxRows === undefined || manuallyResized.current) return;
+
+      const maxHeight = effMaxRows * lineHeight + extra;
+      el.style.height = 'auto';
+      const contentNeeded = borderBox
+        ? el.scrollHeight + border
+        : Math.max(0, el.scrollHeight - padding);
+      const target = Math.min(Math.max(contentNeeded, minHeight), maxHeight);
+      el.style.height = `${target}px`;
+      el.style.overflowY = contentNeeded > maxHeight ? 'auto' : 'hidden';
+    }, [managed, effMinRows, effMaxRows]);
+
+    // Re-fit on mount and whenever the (possibly controlled) value changes.
+    // biome-ignore lint/correctness/useExhaustiveDependencies: re-fit when the content value changes
+    React.useLayoutEffect(() => {
+      resize();
+    }, [resize, props.value, props.defaultValue]);
+
+    // React to the sizing contract itself changing: reset the manual-drag latch and clear any
+    // stale inline styles (auto-grow -> floor-only, or managed -> unmanaged), then re-fit.
+    React.useLayoutEffect(() => {
+      manuallyResized.current = false;
+      const el = innerRef.current;
+      if (!el) return;
+      if (!managed) {
+        restoreOwnedStyles(el);
+        return;
+      }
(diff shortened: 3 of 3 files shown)


packages/apollo-wind/src/components/ui/textarea.tsx
```diff
@@ -2,20 +2,102 @@ import * as React from 'react';
...
+
     return (
       <textarea
         className={cn(
           // Base styles (all themes)
           'flex min-h-[80px] w-full rounded-md border border-input bg-transparent px-3 py-2 text-base transition-colors placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring disabled:cursor-not-allowed disabled:opacity-50 md:text-sm',
           // Future Dark / Future Light overrides
           'future:rounded-xl future:border-0 future:bg-surface-overlay future:text-sm future:placeholder:text-foreground-muted future:placeholder:font-normal future:focus-visible:ring-offset-2 future:focus-visible:ring-offset-background',
+          // Keep the resize handle available (and uncapped) when auto-growing.
+          autoGrow && 'resize-y',
```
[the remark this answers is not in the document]
ayush578: Fixed in 69bbc899. In auto-grow mode the base `min-h-[80px]` class is now dropped and the floor is owned by inline `min-height` (from `minRows`, or none). Since the rendered height is clamped to `maxHeight` in JS, `offsetHeight` can no longer exceed the ceiling and falsely trip `manuallyResized`. Auto-grow is also now gated on `maxRows` only.

packages/apollo-wind/src/components/ui/textarea.stories.tsx
```diff
@@ -49,6 +49,24 @@ export const WithRows: Story = {
...
 
+export const AutoGrow = {
+  render: () => (
+    <div className="grid w-full max-w-sm gap-1.5">
+      <Label htmlFor="autogrow">Auto-growing field</Label>
+      <Textarea
+        id="autogrow"
+        minRows={3}
+        maxRows={5}
+        placeholder="Type multiple lines — the field grows from 3 rows up to 5, then scrolls. Drag the bottom-right handle to make it taller than 5 rows."
```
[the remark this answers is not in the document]
ayush578: Fixed in 69bbc899 — em dashes removed from the story copy (placeholder now uses separate sentences).

packages/apollo-wind/src/components/ui/textarea.stories.tsx
```diff
@@ -49,6 +49,24 @@ export const WithRows: Story = {
...
+      />
+      <p className="text-sm text-muted-foreground">
+        <code>minRows={3}</code> floor, auto-grows to <code>maxRows={5}</code>, then scrolls. The
+        resize handle is not capped — drag past 5 rows to take manual control.
+      </p>
```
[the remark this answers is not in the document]
ayush578: Fixed in 69bbc899 — em dash removed from the description paragraph.

packages/apollo-wind/src/components/ui/textarea.tsx
```diff
@@ -2,20 +2,113 @@ import * as React from 'react';
...
+    // hand sizing control to the user (stop auto-sizing) so the field stays put.
+    React.useLayoutEffect(() => {
+      const el = innerRef.current;
+      if (!el || maxRows === undefined || typeof ResizeObserver === 'undefined') return;
+
+      const observer = new ResizeObserver(() => {
+        const { lineHeight, chrome } = measureRowMetrics(el);
+        const maxHeight = maxRows * lineHeight + chrome;
+        if (el.offsetHeight > maxHeight + 1) manuallyResized.current = true;
+      });
```
SreedharAvvari, line 164: ### 🟠 H1 — Width-driven reflow is not remeasured

The textarea is refit only when `value` or `defaultValue` changes. Its `ResizeObserver` fires when `w-full` or flex layouts change width, but this callback only checks height. Existing text can therefore wrap beyond the stale height while `overflow-y` remains hidden until the next keystroke; the `md:text-sm` metric change also leaves stale row calculations.

SreedharAvvari: Suggested change:

Track the observed width and rerun `resize()` when it changes, and remeasure row metrics when responsive or theme styles change, while keeping height-only user drags on the manual-resize path. Add a regression test that narrows the element after setting a wrapping value and asserts that height and overflow are recomputed.

packages/apollo-wind/src/components/ui/textarea.tsx
```diff
@@ -2,20 +2,113 @@ import * as React from 'react';
...
+      [ref]
+    );
+
+    // Apply the `minRows` floor (as an inline min-height, which also bounds how far the
+    // resize handle can shrink) and, when `maxRows` is set, fit the height to the content
+    // up to the ceiling (scrolling past it). Once the user drags past the ceiling we stop
+    // auto-sizing so their height sticks.
+    const resize = React.useCallback(() => {
+      const el = innerRef.current;
+      if (!el || !managed) return;
```
SreedharAvvari, line 95: ### 🟠 H2 — Disabling auto-grow leaves stale inline styles

When `minRows` and `maxRows` become unset, this early return leaves the previously managed `minHeight`, `height`, and `overflowY` on the DOM node. Switching from auto-grow to floor-only also retains the old fixed height and overflow state, so the textarea does not return to its declared mode.

SreedharAvvari: Suggested change:

Add transition cleanup for styles owned by auto-grow whenever the managed props change or become unset. Restore values supplied through the consumer `style` prop and reset `manuallyResized`; cover both `maxRows → undefined` and `maxRows → minRows-only` rerenders.

packages/apollo-wind/src/components/ui/textarea.tsx
```diff
@@ -2,20 +2,113 @@ import * as React from 'react';
...
+    // A rendered height above the ceiling can only come from a manual drag; once seen,
+    // hand sizing control to the user (stop auto-sizing) so the field stays put.
+    React.useLayoutEffect(() => {
+      const el = innerRef.current;
+      if (!el || maxRows === undefined || typeof ResizeObserver === 'undefined') return;
+
+      const observer = new ResizeObserver(() => {
+        const { lineHeight, chrome } = measureRowMetrics(el);
+        const maxHeight = maxRows * lineHeight + chrome;
+        if (el.offsetHeight > maxHeight + 1) manuallyResized.current = true;
```
SreedharAvvari: ### 🟠 H3 — Manual handoff can permanently suppress scrolling

This marks the field as manually resized but leaves the last auto-grow `overflowY` value in place, which is commonly `hidden`. Every later `resize()` then returns through the manual branch, so content that exceeds the manually selected height has no scrollbar.

SreedharAvvari: Suggested change:

When handing sizing control to the user, restore the consumer-provided overflow value or set `overflowY` to `auto`. Reset that manual state when the sizing props change, and add coverage for typing beyond a manually enlarged field.

packages/apollo-wind/src/components/ui/textarea.test.tsx
```diff
@@ -149,4 +149,58 @@ describe('Textarea', () => {
...
+      expect(textarea.style.minHeight).toMatch(/^\d+px$/);
+      // ...but there is no content-driven height (that only happens with maxRows).
+      expect(textarea.style.height).toBe('');
+    });
+
+    it('auto-grows and enables the resize handle once maxRows is set', () => {
+      render(<Textarea maxRows={5} aria-label="Message" />);
+      const textarea = screen.getByRole('textbox') as HTMLTextAreaElement;
+      expect(textarea).toHaveClass('resize-y');
+    });
```
SreedharAvvari, line 166: ### 🟠 H4 — The auto-grow test does not exercise auto-growth

Despite the test title, the only assertion checks the resize-handle class. None of the new tests verifies computed height, min/max clamping, overflow transitions, `ResizeObserver` handoff, prop removal, or width-driven reflow, so the core sizing contract remains untested.

SreedharAvvari: Suggested change:

Mock `scrollHeight`, `offsetHeight`, computed row metrics, and `ResizeObserver` to assert growth, clamping, overflow, manual handoff, and prop cleanup. Add a browser-level component test for width-driven wrapping, which JSDOM cannot model.

packages/apollo-wind/src/components/ui/textarea.tsx
```diff
@@ -2,20 +2,113 @@ import * as React from 'react';
...
+      el.style.minHeight = minRows !== undefined ? `${minHeight}px` : '';
+
+      // Floor-only (no `maxRows`): keep a fixed height, let content scroll. Nothing else to do.
+      if (maxRows === undefined || manuallyResized.current) return;
+
+      const maxHeight = maxRows * lineHeight + chrome;
+      el.style.height = 'auto';
+      const target = Math.min(Math.max(el.scrollHeight, minHeight), maxHeight);
+      el.style.height = `${target}px`;
+      el.style.overflowY = el.scrollHeight > maxHeight ? 'auto' : 'hidden';
```
SreedharAvvari: ### 🟡 M1 — Border-box sizing clips by the border thickness

`maxHeight` and `minHeight` include the vertical border through `chrome`, while `scrollHeight` excludes borders. Assigning raw `scrollHeight` as a border-box height makes the content box short by the border thickness and can keep overflow hidden at the boundary.

SreedharAvvari: Suggested change:

Expose the measured vertical border separately and use `scrollHeight + verticalBorder` as the target and overflow comparison in border-box mode. Keep a corresponding content-box calculation for consumer overrides, and add an exact-boundary test.

packages/apollo-wind/src/components/ui/textarea.tsx
```diff
@@ -2,20 +2,113 @@ import * as React from 'react';
...
+      const { lineHeight, chrome } = measureRowMetrics(el);
+      const minHeight = minRows !== undefined ? minRows * lineHeight + chrome : 0;
+
+      // Own the floor explicitly; the base `min-h` class is dropped in managed mode.
+      el.style.minHeight = minRows !== undefined ? `${minHeight}px` : '';
+
+      // Floor-only (no `maxRows`): keep a fixed height, let content scroll. Nothing else to do.
+      if (maxRows === undefined || manuallyResized.current) return;
+
+      const maxHeight = maxRows * lineHeight + chrome;
```
SreedharAvvari: ### 🟡 M2 — Conflicting row bounds disable auto-sizing

With `minRows > maxRows`, the inline `min-height` exceeds the JavaScript ceiling. CSS wins, then the observer interprets the oversized result as a manual drag and permanently disables auto-sizing.

SreedharAvvari: Suggested change:

Normalize row counts to positive finite integers and compute `effectiveMinRows = Math.min(minRows, maxRows)` when both are supplied, matching the existing `prompt-editor` contract. Add a regression test for conflicting bounds.

packages/apollo-wind/src/components/ui/textarea.stories.tsx
```diff
@@ -49,6 +49,25 @@ export const WithRows: Story = {
...
+      <Label htmlFor="autogrow">Auto-growing field</Label>
+      <Textarea
+        id="autogrow"
+        minRows={3}
+        maxRows={7}
+        placeholder="Starts at 3 rows and never shrinks below that. Grows with content up to 7 rows, then scrolls. Drag the bottom-right handle to make it taller than 7 rows."
+      />
+      <p className="text-sm text-muted-foreground">
+        <code>minRows={3}</code> is a floor: the field never renders or resizes shorter than 3 rows.
+        <code>maxRows={7}</code> auto-grows with content up to 7 rows, then scrolls. The resize
```
SreedharAvvari: ### 🔵 L1 — The story drops the literal JSX braces

Inside JSX text, `<code>minRows={3}</code>` and `<code>maxRows={7}</code>` render as `minRows=3` and `maxRows=7`, so the documented prop syntax is not what users see.

SreedharAvvari: Suggested change:

Render the examples as string expressions: `<code>{'minRows={3}'}</code>` and `<code>{'maxRows={7}'}</code>`.

packages/apollo-wind/src/components/ui/textarea.tsx
ayush578: Fixed in 0935eeb6. The ResizeObserver now tracks width and reruns `resize()` on width/breakpoint reflow (remeasuring row metrics), separate from the manual-drag path. Added a stubbed-layout test asserting a width change refits height + overflow.

packages/apollo-wind/src/components/ui/textarea.tsx
ayush578: Fixed in 0935eeb6. Added a cleanup effect keyed on the sizing contract: managed→unmanaged restores the consumer's `height`/`minHeight`/`overflowY` (or clears ours), and auto-grow→floor-only drops the fitted height/overflow. `manuallyResized` is reset too. Tests cover both transitions.

packages/apollo-wind/src/components/ui/textarea.tsx
ayush578: Fixed in 0935eeb6. On manual handoff we now set `overflowY` to the consumer's value or `auto` (never frozen `hidden`), and the latch resets when the sizing props change. Added a test that types beyond a manually enlarged field and asserts overflow stays scrollable.

packages/apollo-wind/src/components/ui/textarea.test.tsx
ayush578: Fixed in 0935eeb6. New 'sizing behavior (stubbed layout)' suite mocks getComputedStyle metrics, `scrollHeight`/`offsetHeight`/`clientWidth` and `ResizeObserver`, asserting growth, min/max clamping, overflow transitions, border-box correction, inverted bounds, manual handoff, width reflow, and prop cleanup. Pixel-level text wrapping remains browser-only, but the width-reflow refit path is now covered via the observer.

packages/apollo-wind/src/components/ui/textarea.tsx
ayush578: Fixed in 0935eeb6. `measureRowMetrics` exposes the border separately; in border-box mode the content target and overflow comparison use `scrollHeight + verticalBorder`, with a content-box (`scrollHeight - padding`) branch for consumer overrides. Added a border-box height test.

packages/apollo-wind/src/components/ui/textarea.tsx
ayush578: Fixed in 0935eeb6. Rows are normalized to positive integers and `effMinRows = Math.min(minRows, maxRows)` when both are supplied, so the floor never exceeds the ceiling and the observer can't misread it as a manual drag. Added an inverted-bounds regression test.

packages/apollo-wind/src/components/ui/textarea.stories.tsx
ayush578: Fixed in 0935eeb6 — rendered as `{'minRows={3}'}` / `{'maxRows={7}'}` so the literal braces display.

packages/apollo-wind/src/components/ui/textarea.tsx
```diff
@@ -2,20 +2,189 @@ import * as React from 'react';
...
+      if (!el || !managed || typeof ResizeObserver === 'undefined') return;
+
+      let lastWidth = el.clientWidth;
+      const observer = new ResizeObserver(() => {
+        const width = el.clientWidth;
+        if (width !== lastWidth) {
+          lastWidth = width;
+          resize();
+          return;
+        }
```
SreedharAvvari, line 154: ### 🟠 H1 follow-up — Typography metric changes remain unobserved

Width-driven reflow is now handled, but this callback reruns `resize()` only when `clientWidth` changes. A fixed-width field crossing `md:text-sm` or changing theme/font metrics keeps the old pixel height and `overflow-y` state, so `minRows`/`maxRows` can remain stale and content can be clipped until the value changes.

SreedharAvvari: Suggested change:

Trigger `resize()` on viewport breakpoint and relevant theme/style metric changes even when the element width is unchanged; compare the measured line height, padding, border, and box-sizing before deciding nothing changed. Add a regression test that keeps `clientWidth` constant while changing the computed line height and asserts that height and overflow are recomputed.

packages/apollo-wind/src/components/ui/textarea.test.tsx
```diff
@@ -149,4 +149,28 @@ describe('Textarea', () => {
...
+
+    it('still forwards onChange while auto-growing', async () => {
+      const user = userEvent.setup();
+      const handleChange = vi.fn();
+      render(<Textarea minRows={3} maxRows={5} onChange={handleChange} aria-label="Message" />);
+      await user.type(screen.getByRole('textbox'), 'hi');
+      expect(handleChange).toHaveBeenCalled();
+    });
+  });
```
SreedharAvvari, line 175: ### 🟠 H4 reopened — The sizing behavior suite was lost from the current head

This file still contains only structural class and `onChange` assertions. The stubbed-layout suite cited in the resolution existed in `0935eeb6` but is absent after the squash; the now-unused `act`, `beforeEach`, and `afterEach` imports are remnants of that lost block. Growth, clamping, overflow, manual handoff, reflow, bounds, and prop cleanup therefore remain untested.

SreedharAvvari: Suggested change:

Restore the stubbed-layout suite from `0935eeb6` or equivalent coverage on the current branch, including exact assertions for border-box sizing, min/max clamping, manual overflow handoff, width and metric reflow, inverted bounds, and both prop-cleanup transitions. Keep the imports only when the restored tests use them.

SreedharAvvari (approved):

PR #927, opened 2026-07-17
feat(apollo-react): configure sticky notes from canvas

uipath-utkarsh:
## Summary

- add canvas-wide `stickyNoteOptions` to `BaseCanvas`, so consumers can enable media embedding and read-only behavior without custom sticky-note renderers
- keep direct `StickyNoteNode` props as explicit overrides, including `false`, while preserving backward-compatible defaults
- own interactive-media drag/pan isolation and interrupted-resize settlement inside Apollo
- update the StickyNote Storybook playground to use the public one-option integration and demonstrate image, YouTube, direct-video, natural-width, and full-width behavior

## Why

Flow, Case, and other canvas consumers share Apollo sticky notes. The media experience should therefore be a plug-and-play Apollo capability instead of requiring each consumer to duplicate wrappers, contexts, event filtering, and lifecycle handling.

```tsx
<BaseCanvas
  stickyNoteOptions={{
    enableMediaEmbedding: true,
    readOnly: !allowPropertyEdits,
  }}
/>
```

Canvas option updates keep node types stable and do not remount playing media. Interactive links and video controls cannot drag or pan the canvas, while static images remain valid node drag surfaces. A transition to read-only also settles an active resize exactly once so consumer-owned undo transactions cannot be stranded.

## Validation

- Apollo React focused sticky-note tests: 33 passed
- Apollo React full suite: 114 files / 2,039 tests passed
- Apollo React production build and TypeScript declarations passed
- Storybook production build passed
- changed-file Biome format and lint passed

Jira: [MST-12137](https://uipath.atlassian.net/browse/MST-12137)

[MST-12137]: https://uipath.atlassian.net/browse/MST-12137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/apollo-react/src/canvas/components/StickyNoteNode/StickyNoteNode.tsx
```diff
@@ -164,8 +178,9 @@ const StickyNoteNodeComponent = ({
     if (readOnly) {
       setIsEditing(false);
       updateLocalContent(data.content || '');
+      endResizeLifecycle();
```
SreedharAvvari: ### 🟠 H1 — Read-only transition does not cancel the active resize

`endResizeLifecycle()` marks the session inactive and queues `onResizeEnd`, but removing `NodeResizeControl` does not cancel XYFlow’s window-level active drag listener. Pointer movement after this effect can still emit dimension changes while the note is read-only; when mouseup finally arrives, `handleResizeEnd` sees `resizeActiveRef.current === false` and suppresses the final `onResize`. This leaves the final size outside the consumer transaction or unpersisted.

SreedharAvvari: Suggested change:

Do not notify the consumer that resizing ended until the underlying XYResizer gesture has actually ended or been explicitly cancelled. Either defer applying read-only until the real `onResizeEnd`, or add a cancellation path that removes the active window listeners, captures the final dimensions, calls `onResize`, and then calls `onResizeEnd`. Add a non-mocked integration test that starts a real resize, enables read-only, moves the pointer, and verifies that no further dimension changes occur.

SreedharAvvari (approved):

PR #932, opened 2026-07-20
feat(apollo-wind): add container prop to Popover/Select content for scoped portaling

david-rios-uipath:
## Summary

`PopoverContent` and `SelectContent` bake a Radix `Portal` internally with no way to redirect it — content always portals to `document.body`. This adds an optional `container` prop to each, threaded straight into the internal Radix Portal, so consumers can portal dropdown content somewhere other than `document.body`.

This brings Popover/Select to parity with this repo's own **Tooltip**, which already exports `TooltipPortal` for exactly this reason. Fully additive — `container` defaults to `undefined`, preserving Radix's `document.body` fallback, so every existing consumer is unaffected.

## Motivation

Shadow-DOM / scoped-styles consumers (e.g. `@uipath/traceview` web components) inject their CSS into a style scope that cannot reach `document.body`. When a Popover/Select portals its content to the body, it renders **outside** that scope — unstyled in the light-DOM scoped case, and entirely outside the shadow tree (no tokens at all) in the web-component case. Passing a `container` inside the consumer's style scope fixes this without forking the component.

Verified: this unblocks https://github.com/UiPath/traceview-ui/pull/416, allowing traceview to correctly scope styling for the popover/select components it imports from apollo-wind.

## Follow-ups (out of scope, not in this PR)

We may want to add this option for other portaled components: `dialog`, `alert-dialog`, `sheet`, `drawer`, `dropdown-menu`, `context-menu`, and the prompt-editor autocomplete menu. `multi-select` composes `PopoverContent`, so it inherits the capability but doesn't yet forward its own `container`. These can follow the same trivial pattern in separate PRs.

### DIFF
--- a/packages/apollo-wind/src/components/ui/popover.test.tsx
+++ b/packages/apollo-wind/src/components/ui/popover.test.tsx
@@ -244,23 +244,25 @@ describe('Popover', () => {
     container.setAttribute('data-testid', 'portal-container');
     document.body.appendChild(container);
 
-    render(
-      <Popover>
-        <PopoverTrigger>Open</PopoverTrigger>
-        <PopoverContent container={container}>Contained content</PopoverContent>
-      </Popover>
-    );
+    try {
+      render(
+        <Popover>
+          <PopoverTrigger>Open</PopoverTrigger>
+          <PopoverContent container={container}>Contained content</PopoverContent>
+        </Popover>
+      );
 
-    const trigger = screen.getByRole('button', { name: 'Open' });
-    await user.click(trigger);
+      const trigger = screen.getByRole('button', { name: 'Open' });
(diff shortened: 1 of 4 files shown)


packages/apollo-wind/src/components/ui/popover.test.tsx
```diff
@@ -216,6 +216,55 @@ describe('Popover', () => {
     });
   });
 
+  it('defaults to portaling content into document.body', async () => {
```
david-rios-uipath, line 219: duplicate tests for both components, but figured it's worth verifying for both independently.

david-rios-uipath:
closing this PR in favor of https://github.com/UiPath/apollo-ui/pull/945

PR #935, opened 2026-07-20
feat(apollo-vertex): run selected tests + solution-tests terminology updates [AGVSOL-4876]

frankkluijtmans:
⚠️ NOTE: This should include all planned except the FINS specific loan type column addition and making the test run details into a separate view which I will do in a separate PR.

## Summary

Updates the Apollo Vertex solution-tests experience with a new bulk-run action and clearer terminology.

- Adds row-selection checkboxes to the test cases table and a "Run selected" button next to "Run all", disabled until at least one test is selected and routed through the existing run-confirmation dialog.
- Renames labels for clarity: "Baseline agent version" and "Tested agent version" in the run-details dialog, a dedicated "Solution version" column, and "expected result" wording replaced with "baseline" in toasts and tooltips.
- Makes the test cases tab label subject-aware (e.g. "Test loans") and refines KPI/label copy, with the corresponding English-only locale keys added.

👨 Generated with Kluijt Code


apps/apollo-vertex/locales/en.json
```diff
@@ -291,5 +296,6 @@
   "wrong_dimension_type_in_joined": "Field \"{{field}}\" is {{actual}}; this chart needs a {{expected}} field.",
   "wrong_metric_type_in_entity": "Field \"{{field}}\" on {{entity}} is {{actual}}; {{aggregation}} requires a numeric field.",
   "wrong_metric_type_in_joined": "Field \"{{field}}\" is {{actual}}; {{aggregation}} requires a numeric field.",
-  "you": "You"
+  "you": "You",
+  "cases": "cases"
```
RobinMennens-UiPath: `"cases"` is unused. The tab label falls back to a literal string (`...toLowerCase() ?? "cases"` in `solution-tests-view.tsx`), not `t("cases")`, so this key is never read. Suggest removing it.

apps/apollo-vertex/locales/en.json
```diff
@@ -5,14 +5,14 @@
   "actions": "Actions",
   "actual_input": "Actual input",
   "actual_output": "Actual output",
-  "actual_version": "Actual version",
+  "tested_agent_version": "Tested agent version",
```
RobinMennens-UiPath: AGENTS.md requires new keys in alphabetical order, but several are placed at their old key's position: `tested_agent_version` (here, in the `actual_*` block), `baseline_agent_version` (line 36, should precede `baseline_updated_successfully`), `solution_version` (line 287, after `version`), and `cases` (line 300, end of file). Biome doesn't reorder JSON keys, so these need to be moved manually.

RobinMennens-UiPath (approved):
Approving. Verified against AGVSOL-4876: 6/8 bullets addressed, and the two remaining (FINS loan-type column, run details as a separate page) are the follow-ups called out in the PR description. The `subject_passing` KPI wording change is backed by real data (`lastBatch.TestsPassed`), so "in last run" is accurate. Two small locale nits inline.

ruudandriessen (approved):

PR #937, opened 2026-07-20
feat(apollo-vertex): run details page for solution tests [AGVSOL-4876]

frankkluijtmans:
Screenshot 2026-07-20 at 17 22 17
Screenshot 2026-07-21 at 08 33 48

Replaces the Solution Tests run-details dialog with a dedicated full-page view: a left sidebar lists the agents in a run and the right pane shows the selected agent's result, reusing the existing evaluator renderers unchanged. The page is driven by a `?run=` search param, and the baseline adopt / update / remove actions move into the detail pane. Adds an interactive `RunDetailsTemplate` example to the solution-tests docs page and updates the demo template accordingly. Removes the old `run-details-dialog*` components.

👨 Generated with Kluijt Code


apps/apollo-vertex/locales/en.json
```diff
@@ -12,6 +12,7 @@
   "agent_name": "Agent name",
   "agent_produced_no_output": "Agent produced no output",
   "agent_removed_from_expected_results": "Agent removed from baseline",
+  "agents": "Agents",
   "agents_passed": "Agents passed",
```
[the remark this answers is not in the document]
frankkluijtmans: Intentional: per this app's AGENTS.md, PRs only edit `locales/en.json`. The other locale files are synced separately by the localization team, and PRs that modify non-English locales get sent back. New keys are therefore added to `en.json` only.

apps/apollo-vertex/registry/solution-tests/run-details-view.tsx
```diff
@@ -0,0 +1,288 @@
...
+          id: subjectId,
+        })
+      : subjectId
+  } — ${t("run_results")}`;
+
+  const selectedResult = results.find((r) => r.Id === selectedResultId);
+
+  return (
+    <div className="flex h-full min-h-[600px] flex-col">
+      <PageHeader>
```
RobinMennens-UiPath, line 101: The run-details page header only shows the title and the back button. The run-level Score, Agents passed, and Status (the columns from the Test runs table) aren't surfaced anywhere on this page, so that context is lost once you drill into a run. Now that this is a full page with more room, consider showing those run-level metrics in the header.

apps/apollo-vertex/registry/solution-tests/expanded-run-tests-view.tsx
```diff
@@ -86,18 +95,23 @@ export const ExpandedRunTestsView = ({
...
+                          <Button
+                            variant="ghost"
+                            size="icon-sm"
+                            aria-label={t("details")}
+                            onClick={(e) => {
+                              e.stopPropagation();
+                              onOpenDetails(run);
+                            }}
+                          >
+                            <Eye className="size-4" />
```
RobinMennens-UiPath: The details action went from a labelled `outline` button (`Details` + Info icon) to a bare ghost `Eye` icon with only a tooltip, which is a weaker affordance for the primary action on a completed run. Consider a labelled button (e.g. `View results`) so it's obvious. The whole row is clickable now too, but that isn't discoverable on its own.

apps/apollo-vertex/registry/solution-tests/run-details-view.tsx
```diff
@@ -0,0 +1,288 @@
...
+          </div>
+          <div className="flex flex-wrap items-center gap-x-6 gap-y-1 text-xs text-muted-foreground">
+            <span>
+              {`${t("baseline_agent_version")}: `}
+              {showBaselineVersion
+                ? (result.BaselineProcessVersion ?? "-")
+                : "—"}
+            </span>
+            <span className="inline-flex items-center gap-1">
+              {`${t("tested_agent_version")}: `}
```
RobinMennens-UiPath, line 253: Once you select an agent, the baseline and tested versions are shown as raw values and the only hint that the version changed is a small `▲` glyph next to the tested version. It's easy to miss. Consider a clear, labelled icon (e.g. a lucide warning icon) so a version change is obvious. Minor: the `▲` is currently a `<button>` with no `onClick`, so screen readers announce an actionable control that does nothing — it should be a non-interactive focusable element (or the previous `<span>`) instead.

apps/apollo-vertex/registry/solution-tests/run-details-view.tsx
```diff
@@ -0,0 +1,288 @@
...
+              {result.ProcessName}
+            </h3>
+            <Badge
+              variant="secondary"
+              status={runResultStatusBadgeMap[status] ?? "info"}
+              className={resultBadgeClassMap[status]}
+            >
+              {defaultRunResultStatusLabels[status] ?? "Unknown"}
+            </Badge>
+            {isBaselineSource && (
```
RobinMennens-UiPath: The old dialog had an *In baseline* column with four states: `current baseline` (blue), `✓ in baseline`, `removed` (red), and `—`. The new pane renders only the `current baseline` label. The `✓ in-baseline` state is now only implied by which action buttons appear, and the `removed` (red) state is dropped entirely — an agent whose baseline was removed now looks identical to one that was never in the baseline (both show only `Add to baseline`). That's a loss of signal; consider surfacing the in-baseline / removed state explicitly.

apps/apollo-vertex/registry/solution-tests/solution-tests.tsx
```diff
@@ -84,10 +85,36 @@ export const SolutionTests = () => {
...
   };
 
+  // The selected run drives the full-page run-details view. It's synced to the
+  // `?run=` search param so it survives reloads and is shareable.
+  const selectedRunId = useSearch({
+    strict: false,
+    select: (search: Record<string, unknown>) =>
+      typeof search.run === "string" ? search.run : "",
+  });
+  const selectedRun = allRuns.find((r) => r.Id === selectedRunId);
```
RobinMennens-UiPath: `selectedRun` is matched by id only, with no run-status guard, whereas the table exposes details only for completed runs. A `?run=<pending-run-id>` (a stale or shared link) will render the detail page for a not-done run with empty/partial results. It won't crash, but it's a confusing empty state. Consider guarding on run status (mirror `isRunDone`) and/or clearing an unmatched `run` param.

apps/apollo-vertex/registry/solution-tests/run-details-view.tsx
```diff
@@ -0,0 +1,288 @@
...
+              <span className="text-xs font-medium text-blue-600">
```
RobinMennens-UiPath: `text-blue-600` is a raw Tailwind color; the surrounding code uses semantic tokens (`text-warning-foreground`, `text-destructive`, `text-muted-foreground`) and AGENTS.md prefers semantic tokens. Consider a semantic token here. (Carried over from the old dialog.)

apps/apollo-vertex/registry/solution-tests/run-details-view.tsx
```diff
@@ -0,0 +1,288 @@
...
+              onAdopt={onAdopt}
+              onUpdateBaseline={onUpdateBaseline}
+              onRemoveBaseline={onRemoveBaseline}
+            />
+          ) : (
+            <Empty>
+              <EmptyDescription>
+                {results.length === 0
+                  ? t("no_results_available")
+                  : t("select_an_agent")}
```
RobinMennens-UiPath: The `select_an_agent` empty-state branch looks unreachable: this view computes `selectedResult` with no fallback, but every caller (the smart wrapper and both templates) always resolves `selectedResultId` to a real id via `?? results[0]`. So when `results` is non-empty a selection always matches, and this branch never renders. The newly-added `select_an_agent` key is therefore never shown. Either wire a code path that can actually hit it, or drop the branch (and the key).

RobinMennens-UiPath (changes requested):
Review of the run-details dialog → full-page refactor. Most of the refactor preserves behavior faithfully (baseline action guards, lazy attachment fetching, force-stop gating, and the shared evaluator renderers are all intact). A few UX items, one behavior regression, and some nits are noted inline. Two items can't be attached inline because they sit on unchanged lines:

**Mock warning copy — `templates/solution-tests/mock-db.ts:34`**: The demo warning "Baseline is older than the current solution version." is mock data (not produced by any real version/solution logic), so this isn't a code issue. But it reads as a *solution-version* warning surfaced at the *agent* level, which is confusing in the docs preview. Since it's just demo copy, consider rewording it (e.g. an agent-version-oriented message) to avoid that confusion.

**Silent catch — `registry/solution-tests/run-details.tsx:122`**: The outer `try { … } catch { setRowData(… loading:false) }` in `fetchRowAttachments` swallows genuine failures with no log or user feedback. The inner `.catch(() => null)` calls intentionally tolerate missing optional attachments (fine), but the outer empty catch also hides real errors. Consider at least logging in the outer catch so failures aren't invisible. (Carried over from the old dialog.)

apps/apollo-vertex/registry/solution-tests/expanded-run-tests-view.tsx
frankkluijtmans: I didn't like the button we had before tbh, it kinda looks off in a table. Maybe I should just make the loan id clickable like the test cases table?

apps/apollo-vertex/registry/solution-tests/run-details-view.tsx
frankkluijtmans: This is the same way we did this before, but I can look into an alternative

apps/apollo-vertex/registry/solution-tests/run-details-view.tsx
frankkluijtmans: Added run-level Score, Agents passed, and Status to the run-details page header (PageHeaderContent/Field), wired from the run.

apps/apollo-vertex/registry/solution-tests/expanded-run-tests-view.tsx
frankkluijtmans: Went with the loan-id approach: the test-name cell is now a clickable link (like the Test cases table) and the whole row stays clickable. Dropped the ghost Eye icon.

apps/apollo-vertex/registry/solution-tests/run-details-view.tsx
frankkluijtmans: Replaced the tiny ▲ with a labelled lucide TriangleAlert + tooltip, and made it a non-interactive icon (role=img + aria-label) so it's no longer a no-op button.

apps/apollo-vertex/registry/solution-tests/run-details-view.tsx
frankkluijtmans: Restored the explicit baseline state in the pane header: Current baseline / ✓ In baseline / Removed (destructive).

apps/apollo-vertex/registry/solution-tests/solution-tests.tsx
frankkluijtmans: Guarded: selectedRun now also requires isRunDone, so a stale/shared ?run=<pending-id> falls through to the table instead of an empty detail page.

apps/apollo-vertex/registry/solution-tests/run-details-view.tsx
frankkluijtmans: Switched text-blue-600 to the semantic text-info token.

apps/apollo-vertex/registry/solution-tests/run-details-view.tsx
frankkluijtmans: Dropped the unreachable select_an_agent branch and its en.json key; the empty pane shows no_results_available (only reachable when a run has zero results).

apps/apollo-vertex/registry/solution-tests/expanded-run-tests-view.tsx
RobinMennens-UiPath: In the feedback table we also have a button.
I considered the link as well but IMO it makes more sense if that linnks to the actual loan itself

frankkluijtmans: Ok, we'll put back the button as it was before.

frankkluijtmans: Follow-up: after trying the clickable loan-id link, we decided to keep the loan ID as plain text and bring back the original labelled `Details` button (outline + Info icon) instead. The whole row remains clickable too.

frankkluijtmans:
@RobinMennens-UiPath some screenshots from my latest changes

Screenshot 2026-07-21 at 16 00 18
Screenshot 2026-07-21 at 15 59 28

apps/apollo-vertex/templates/RunDetailsTemplate.tsx
```diff
@@ -0,0 +1,107 @@
...
+function RunDetailsTemplateContent() {
+  const [db] = useState(() => createMockDb());
+  const [selectedResultId, setSelectedResultId] = useState<string | null>(null);
+
+  // The mock db only fleshes out per-agent results for one run, so drive the
+  // example off that run — every agent then has evaluator output to show.
+  const run = db.runs.find((r) =>
+    db.results.some((res) => res.SolutionTestRunId === r.Id),
+  );
+  if (!run) return null;
```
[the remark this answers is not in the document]
frankkluijtmans: Fixed: the mock now generates per-agent results for every run, so I dropped the stale comment and the unnecessary `.find(...)` — the template just uses `db.runs[0]`.

RobinMennens-UiPath:
>
> Screenshot 2026-07-21 at 15 59 28

Why is there a warning icon next to the version number? Are there actual warnings on hover?
I would say we need some arrow up/down icon indicating visually there is a change in the agent version, like we have now:
image

frankkluijtmans:
> > Screenshot 2026-07-21 at 15 59 28
> 
> Why is there a warning icon next to the version number? Are there actual warnings on hover? I would say we need some arrow up/down icon indicating visually there is a change in the agent version, like we have now: image

This displays a tooltip warning the user of the version mismatch. The point was to replace the arrow, right? It was literally suggested even:

> Once you select an agent, the baseline and tested versions are shown as raw values and the only hint that the version changed is a small `▲` glyph next to the tested version. It's easy to miss. Consider a clear, labelled icon (e.g. a lucide warning icon) so a version change is obvious. Minor: the `▲` is currently a `<button>` with no `onClick`, so screen readers announce an actionable control that does nothing — it should be a non-interactive focusable element (or the previous `<span>`) instead.

RobinMennens-UiPath:
> > > Screenshot 2026-07-21 at 15 59 28
> > 
> > 
> > Why is there a warning icon next to the version number? Are there actual warnings on hover? I would say we need some arrow up/down icon indicating visually there is a change in the agent version, like we have now: image
> 
> This displays a tooltip warning the user of the version mismatch. The point was to replace the arrow, right? It was literally suggested even:
> 
> > Once you select an agent, the baseline and tested versions are shown as raw values and the only hint that the version changed is a small `▲` glyph next to the tested version. It's easy to miss. Consider a clear, labelled icon (e.g. a lucide warning icon) so a version change is obvious. Minor: the `▲` is currently a `<button>` with no `onClick`, so screen readers announce an actionable control that does nothing — it should be a non-interactive focusable element (or the previous `<span>`) instead.

Sorry my bad.. had to look better at what Claude wrote in my name. 
The version changing is not really a warning. Its not bad perse but we should highlight the fact that it changed. Because that can explain why regressions are there.
What I meant to say is that we _used to_ have the up/down `▲` glyph indicating a change but in the first screenshots you shared that glyph was not there anymore

frankkluijtmans:
> > > > Screenshot 2026-07-21 at 15 59 28
> > > 
> > > 
> > > Why is there a warning icon next to the version number? Are there actual warnings on hover? I would say we need some arrow up/down icon indicating visually there is a change in the agent version, like we have now: image
> > 
> > 
> > This displays a tooltip warning the user of the version mismatch. The point was to replace the arrow, right? It was literally suggested even:
> > > Once you select an agent, the baseline and tested versions are shown as raw values and the only hint that the version changed is a small `▲` glyph next to the tested version. It's easy to miss. Consider a clear, labelled icon (e.g. a lucide warning icon) so a version change is obvious. Minor: the `▲` is currently a `<button>` with no `onClick`, so screen readers announce an actionable control that does nothing — it should be a non-interactive focusable element (or the previous `<span>`) instead.
> 
> Sorry my bad.. had to look better at what Claude wrote in my name. The version changing is not really a warning. Its not bad perse but we should highlight the fact that it changed. Because that can explain why regressions are there. What I meant to say is that we _used to_ have the up/down `▲` glyph indicating a change but in the first screenshots you shared that glyph was not there anymore

That explains the confusion 👌

frankkluijtmans:
Screenshot 2026-07-22 at 14 49 48
@RobinMennens-UiPath

RobinMennens-UiPath (approved):

apps/apollo-vertex/registry/solution-tests/expanded-run-tests-view.tsx
```diff
@@ -60,8 +56,16 @@ export const ExpandedRunTestsView = ({
...
+              const done = isRunDone(run.Status);
+
               return (
-                <TableRow key={run.Id}>
+                <TableRow
+                  key={run.Id}
+                  className={done ? "cursor-pointer hover:bg-muted/50" : ""}
+                  onClick={() => {
+                    if (done) onOpenDetails(run);
+                  }}
```
alincadariu: should we show a toast that its not done? or disable it while its not done?

apps/apollo-vertex/registry/solution-tests/run-details.tsx
```diff
@@ -134,36 +129,38 @@ export const RunDetailsDialog = ({
...
+  useEffect(() => {
+    if (results.length === 0) return;
+    if (!results.some((r) => r.Id === selectedResultId)) {
+      setSelectedResultId(results[0].Id);
+    }
+  }, [results, selectedResultId]);
+
+  useEffect(() => {
+    if (!selectedResult) return;
+    if (!rowData[selectedResult.Id]) void fetchRowAttachments(selectedResult);
```
alincadariu: isn't this a useQuery with key "selectedResult"?

apps/apollo-vertex/registry/solution-tests/expanded-run-tests-view.tsx
frankkluijtmans: It should actually be disabled if the run is not done 🤔

alincadariu (approved):

PR #945, opened 2026-07-21
feat(apollo-wind): add support for passing containers for portal

scottcmg:
## Description
In apollo-wind components with popovers there's an issue with using the components in a larger component meant to be embedded in a variety of hosts. There's no way to pass a container to the Portal to control where the popover is placed in the DOM, so the popover is at the mercy of the host that is outside of the owning components DOM/tree.

This PR is meant to add a Portal container provider to allow caller to optionally pass a container that can be rendered inside the parent/owning component to avoid issues with popovers in a host (e.g. focus traps, shadow DOMs, etc.).

### DIFF
--- a/packages/apollo-wind/src/components/ui/popover.tsx
+++ b/packages/apollo-wind/src/components/ui/popover.tsx
@@ -1,6 +1,11 @@
+'use client';
+
 import * as PopoverPrimitive from '@radix-ui/react-popover';
 import * as React from 'react';
-import { usePortalContainer } from '@/components/ui/portal-container';
+import {
+  type PortalContainerOverride,
+  useResolvedPortalContainer,
+} from '@/components/ui/portal-container';
 import { cn } from '@/lib';
 
 const Popover = PopoverPrimitive.Root;
@@ -12,13 +17,12 @@ const PopoverAnchor = PopoverPrimitive.Anchor;
 const PopoverContent = React.forwardRef<
   React.ElementRef<typeof PopoverPrimitive.Content>,
   React.ComponentPropsWithoutRef<typeof PopoverPrimitive.Content> & {
-    container?: HTMLElement | null;
+    container?: PortalContainerOverride;
   }
 >(({ className, align = 'center', sideOffset = 4, container, ...props }, ref) => {
-  const portalContainer = usePortalContainer();
-  const resolvedContainer = container !== undefined ? container : portalContainer;
+  const resolvedContainer = useResolvedPortalContainer(container);
   return (
-    <PopoverPrimitive.Portal container={resolvedContainer ?? undefined}>
+    <PopoverPrimitive.Portal container={resolvedContainer}>
       <PopoverPrimitive.Content
         ref={ref}
         align={align}
--- a/packages/apollo-wind/src/components/ui/popover.tsx
+++ b/packages/apollo-wind/src/components/ui/popover.tsx
@@ -1,8 +1,7 @@
 import * as PopoverPrimitive from '@radix-ui/react-popover';
 import * as React from 'react';
-
-import { cn } from '@/lib';
 import { usePortalContainer } from '@/components/ui/portal-container';
+import { cn } from '@/lib';
 
 const Popover = PopoverPrimitive.Root;
 
@@ -17,8 +16,9 @@ const PopoverContent = React.forwardRef<
   }
 >(({ className, align = 'center', sideOffset = 4, container, ...props }, ref) => {
   const portalContainer = usePortalContainer();
+  const resolvedContainer = container !== undefined ? container : portalContainer;
   return (
-    <PopoverPrimitive.Portal container={container ?? portalContainer ?? undefined}>
+    <PopoverPrimitive.Portal container={resolvedContainer ?? undefined}>
       <PopoverPrimitive.Content
         ref={ref}
         align={align}
--- a/packages/apollo-wind/src/components/ui/popover.tsx
+++ b/packages/apollo-wind/src/components/ui/popover.tsx
@@ -2,6 +2,7 @@ import * as PopoverPrimitive from '@radix-ui/react-popover';
 import * as React from 'react';
 
 import { cn } from '@/lib';
+import { usePortalContainer } from '@/components/ui/portal-container';
 
 const Popover = PopoverPrimitive.Root;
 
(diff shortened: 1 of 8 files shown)


packages/apollo-wind/src/components/ui/select.tsx
```diff
@@ -60,34 +60,40 @@ SelectScrollDownButton.displayName = SelectPrimitive.ScrollDownButton.displayNam
...
-      {...props}
-    >
-      <SelectScrollUpButton />
-      <SelectPrimitive.Viewport
+  React.ComponentPropsWithoutRef<typeof SelectPrimitive.Content> & {
+    container?: HTMLElement | null;
+  }
+>(({ className, children, position = 'popper', container, ...props }, ref) => {
+  const portalContainer = usePortalContainer();
+  const resolvedContainer = container !== undefined ? container : portalContainer;
```
david-rios-uipath: imo it would be safer to extract a shared hook (e.g. `useResolvedPortalContainer()`) that includes this logic:
`const resolvedContainer = container !== undefined ? container : portalContainer;` so it's not duplicated across every consumer.

Testing `useResolvedPortalContainer` resolution once would also ensure it works as expected for all components as well.

packages/apollo-wind/src/components/ui/popover.tsx
```diff
@@ -11,21 +11,27 @@ const PopoverAnchor = PopoverPrimitive.Anchor;
...
-      className={cn(
-        'z-50 w-72 rounded-md border bg-popover p-4 text-popover-foreground shadow-md outline-none future:rounded-lg data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2 origin-[--radix-popover-content-transform-origin]',
-        className
-      )}
-      {...props}
-    />
-  </PopoverPrimitive.Portal>
-));
+  React.ComponentPropsWithoutRef<typeof PopoverPrimitive.Content> & {
+    container?: HTMLElement | null;
```
david-rios-uipath: **`null`-from-ref footgun in the `container` prop.** The natural call `<PopoverContent container={someRef.current}>` passes `null` on the first render (before the ref attaches), and the tri-state reads `null` as "force `document.body`", not "defer to provider". Tellingly, this PR's own test works around it at `portal-container.test.tsx:55` by guarding `{target && <PopoverContent container={target}>}`. When the tests must route around the API to avoid the wrong branch, the type isn't expressing the invariant. Consider a named sentinel (`container?: HTMLElement | 'body'`) so `null` from a ref can't silently select force-body, or at least document the hazard on the prop. (Applies equally to `select.tsx` and `dropdown-menu.tsx`.)

packages/apollo-wind/src/components/ui/portal-container.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const [boundary, setBoundary] = React.useState<HTMLElement | null>(null);
+  // `undefined` → auto boundary; explicit `null` → force `document.body`.
+  const value = container !== undefined ? container : boundary;
+  return (
+    <PortalContainerContext.Provider value={value}>
+      {children}
+      {container === undefined && (
+        // display:contents so the boundary never affects layout.
+        <div ref={setBoundary} style={{ display: 'contents' }} />
+      )}
```
david-rios-uipath, line 61: **Initially-open overlays flash into `document.body`, then remount into the boundary.** `boundary` starts `null` and the boundary `<div>` renders *after* `children`, so on the first commit the context value is `null` → any overlay **open at mount** resolves to `document.body` (outside the trigger root, the precise failure mode this change fixes), then remounts once the ref callback fires (re-running focus effects, replaying open animations). Either handle the initially-open case or document "open-after-mount is the supported path" as a known limitation.

david-rios-uipath (approved):
thanks for making this fix! was also looking to make a similar fix [here](https://github.com/UiPath/apollo-ui/pull/932), and I much prefer this approach.

Approved to unblock -- +1 to all the copilot comments and left some of my own.

scottcmg:
> thanks for making this fix! was also looking to make a similar fix [here](https://github.com/UiPath/apollo-ui/pull/932), and I much prefer this approach.
> 
> Approved to unblock -- +1 to all the copilot comments and left some of my own.

Thanks for reviewing @david-rios-uipath - I didn't see your original PR, sorry for the overlap! I addressed the comments from before but it looks like the original approval was dismissed on update, if you have a moment could take another look with the changes? Thanks!

david-rios-uipath (approved):

packages/apollo-wind/src/components/ui/portal-container.tsx
```diff
@@ -0,0 +1,65 @@
...
+ */
+export function PortalContainerProvider({ children, container }: PortalContainerProviderProps) {
+  const [boundary, setBoundary] = React.useState<HTMLElement | null>(null);
+  const value = container ?? boundary;
+  return (
+    <PortalContainerContext.Provider value={value}>
+      {children}
+      {container == null && (
+        // display:contents so the boundary never affects layout.
+        <div ref={setBoundary} style={{ display: 'contents' }} />
```
david-rios-uipath: nit: can we use the `contents` tailwind class here instead?

packages/apollo-wind/src/components/ui/portal-container.tsx
```diff
@@ -0,0 +1,65 @@
...
+import * as React from 'react';
+
+/**
+ * Node overlays portal into. `null` means fall back to `document.body` — either
+ * no provider is mounted, or one is but its boundary hasn't attached yet; it is
+ * not a reliable "is a provider mounted?" signal.
+ */
+const PortalContainerContext = React.createContext<HTMLElement | null>(null);
+
+export const usePortalContainer = (): HTMLElement | null =>
```
david-rios-uipath: is `usePortalContainer` still a hook we want to export? at first glance it looks like it's made redundant by `useResolvedPortalContainer` but not sure if there might be any uses for customizable behavior based on a given container

scottcmg: Makes sense - I had it as an export just in case but there's no use case for it now I can think of. The export can be added back if a need comes up in the future.

david-rios-uipath (approved):

david-rios-uipath (approved):

PR #955, opened 2026-07-22
fix(apollo-react): edge labels lose background and can render under crossing lines

dbacomputer:
## Summary

Fixes edge labels losing their background or rendering beneath crossing paths, and improves the interaction between a path and its associated label.

Screenshot 2026-07-30 at 10 28 51 AM

## What changed

- Render edge labels through xyflow's `EdgeLabelRenderer`, keeping labels above every edge SVG regardless of edge order or selection state.
- Preserve an opaque label surface with a `--color-background` fallback when `--canvas-background` is unavailable.
- Match the label's 1px border to the edge's resolved color across hover, selection, diff, validation, preview, and execution states.
- Elevate the hovered edge above intersecting default or selected edges so the complete path remains traceable at crossings.
- Truncate long labels at a 192px maximum width and show the full value in a tooltip only when the text is clipped.
- Treat the label as part of its edge interaction: hover highlights the owning edge and click selects it, including additive Ctrl/Cmd selection.
- Preserve read-only semantics: hover tracing and overflow tooltips remain available, while path and label selection stay disabled.
- Keep `StageEdge` label backgrounds on the same fallback behavior.

## Root cause

`EdgeLabel` previously rendered as a `foreignObject` inside its edge group. That left the label competing with other edge SVGs in DOM/z-index order, allowing a crossing edge to paint over it. The label background also referenced `--canvas-background` without a fallback, which could resolve to transparent in hosts without a canvas theme ancestor.

## Storybook coverage

The focused `Components/Edges/EdgeLabels` page now covers:

- Orientation
- Diff states
- Crossing labeled edges
- Long-label overflow and short-edge crowding
- Execution status
- Bent paths
- Read-only behavior

Theme-matrix, missing-theme diagnostic, and unsupported inline-editing stories were removed to keep the reference page focused on supported consumer behavior. Theme switching remains available from the Storybook toolbar.

## Validation

- [x] Apollo React test suite: 2,067 tests across 116 files
- [x] TypeScript: `tsc --noEmit -p tsconfig.json`
- [x] Focused Biome formatting check
- [x] `git diff --check`
- [x] Verified in Storybook: crossing-edge stacking, synchronized label colors, overflow detection, label selection, and read-only selection prevention

The localization compile step reports existing non-strict catalog warnings, but completes successfully and the full test command exits cleanly.

### DIFF
--- a/packages/apollo-react/src/canvas/components/Edges/CanvasEdge.tsx
+++ b/packages/apollo-react/src/canvas/components/Edges/CanvasEdge.tsx
@@ -1,5 +1,5 @@
-import { Position } from '@uipath/apollo-react/canvas/xyflow/react';
-import { memo, useCallback, useRef, useState } from 'react';
+import { Position, useReactFlow } from '@uipath/apollo-react/canvas/xyflow/react';
+import { memo, type MouseEvent, useCallback, useRef, useState } from 'react';
 import { isPreviewEdge } from '../../utils/createPreviewNode';
 import { useBaseCanvasMode } from '../BaseCanvas/BaseCanvasModeProvider';
 import { EdgeToolbar, useEdgeToolbarState } from '../Toolbar';
@@ -53,6 +53,30 @@ export const CanvasEdge = memo(function CanvasEdge({
   const onMouseEnter = useCallback(() => setIsHovered(true), []);
   const onMouseLeave = useCallback(() => setIsHovered(false), []);
   const pathRef = useRef<SVGPathElement | null>(null);
+  const { setEdges, setNodes } = useReactFlow();
+
+  const onLabelClick = useCallback(
+    (event: MouseEvent<HTMLDivElement>) => {
+      event.stopPropagation();
+      const additiveSelection = event.metaKey || event.ctrlKey;
+
+      setEdges((edges) =>
+        edges.map((edge) => {
+          if (edge.id === id) {
+            return { ...edge, selected: additiveSelection ? !edge.selected : true };
+          }
+          return additiveSelection ? edge : { ...edge, selected: false };
+        })
+      );
+
+      if (!additiveSelection) {
+        setNodes((nodes) =>
+          nodes.map((node) => (node.selected ? { ...node, selected: false } : node))
+        );
+      }
+    },
+    [id, setEdges, setNodes]
+  );
 
   const routing = data?.routing ?? 'waypoint';
   const storedWaypoints = data?.waypoints ?? EMPTY_WAYPOINTS;
@@ -150,6 +174,7 @@ export const CanvasEdge = memo(function CanvasEdge({
   return (
     <>
       <g
+        data-edge-hovered={isHovered || undefined}
         onMouseEnter={onMouseEnter}
         onMouseLeave={onMouseLeave}
         onMouseMove={toolbarEnabled ? toolbar.handleMouseMoveOnPath : undefined}
@@ -208,7 +233,10 @@ export const CanvasEdge = memo(function CanvasEdge({
             x={geometry.labelPoint.x}
             y={geometry.labelPoint.y}
             text={label}
-            selected={selected}
+            borderColor={color}
+            onClick={isReadOnly ? undefined : onLabelClick}
+            onMouseEnter={onMouseEnter}
+            onMouseLeave={onMouseLeave}
           />
         )}
       </g>
--- a/packages/apollo-react/src/canvas/components/Edges/CanvasEdge.tsx
+++ b/packages/apollo-react/src/canvas/components/Edges/CanvasEdge.tsx
@@ -1,5 +1,5 @@
-import { Position, useReactFlow } from '@uipath/apollo-react/canvas/xyflow/react';
-import { memo, type MouseEvent, useCallback, useRef, useState } from 'react';
+import { Position, useStoreApi } from '@uipath/apollo-react/canvas/xyflow/react';
+import { type MouseEvent, memo, useCallback, useRef, useState } from 'react';
 import { isPreviewEdge } from '../../utils/createPreviewNode';
 import { useBaseCanvasMode } from '../BaseCanvas/BaseCanvasModeProvider';
 import { EdgeToolbar, useEdgeToolbarState } from '../Toolbar';
@@ -53,29 +53,30 @@ export const CanvasEdge = memo(function CanvasEdge({
   const onMouseEnter = useCallback(() => setIsHovered(true), []);
   const onMouseLeave = useCallback(() => setIsHovered(false), []);
   const pathRef = useRef<SVGPathElement | null>(null);
-  const { setEdges, setNodes } = useReactFlow();
+  const store = useStoreApi();
 
   const onLabelClick = useCallback(
     (event: MouseEvent<HTMLDivElement>) => {
       event.stopPropagation();
-      const additiveSelection = event.metaKey || event.ctrlKey;
+      const {
+        addSelectedEdges,
+        edgeLookup,
+        elementsSelectable,
+        multiSelectionActive,
+        unselectNodesAndEdges,
+      } = store.getState();
+      const edge = edgeLookup.get(id);
 
-      setEdges((edges) =>
-        edges.map((edge) => {
-          if (edge.id === id) {
-            return { ...edge, selected: additiveSelection ? !edge.selected : true };
-          }
-          return additiveSelection ? edge : { ...edge, selected: false };
-        })
-      );
+      if (!edge || !(edge.selectable ?? elementsSelectable)) return;
 
-      if (!additiveSelection) {
-        setNodes((nodes) =>
-          nodes.map((node) => (node.selected ? { ...node, selected: false } : node))
-        );
+      store.setState({ nodesSelectionActive: false });
+      if (edge.selected && multiSelectionActive) {
+        unselectNodesAndEdges({ nodes: [], edges: [edge] });
+      } else {
+        addSelectedEdges([id]);
       }
     },
-    [id, setEdges, setNodes]
+    [id, store]
   );
 
   const routing = data?.routing ?? 'waypoint';
--- a/packages/apollo-react/src/canvas/components/Edges/shared/primitives/EdgeLabel.tsx
+++ b/packages/apollo-react/src/canvas/components/Edges/shared/primitives/EdgeLabel.tsx
@@ -1,3 +1,7 @@
+import { EdgeLabelRenderer } from '@uipath/apollo-react/canvas/xyflow/react';
+import { cn } from '@uipath/apollo-wind';
+import { useMemo } from 'react';
+
 export type EdgeLabelProps = {
   x: number;
   y: number;
@@ -5,35 +9,34 @@ export type EdgeLabelProps = {
   selected?: boolean;
 };
 
+// Falls back to --color-background when a host doesn't import canvas/styles/variables.css
+// (e.g. a shadow-DOM host), so the label never renders with a transparent background.
+const EDGE_LABEL_BASE_CLASS =
+  'react-flow__edge-label nodrag nopan absolute top-0 left-0 whitespace-nowrap pointer-events-none ' +
+  'px-2 py-1 rounded text-xs font-medium border shadow-[0_1px_3px_0_rgba(0,0,0,0.1)] ' +
+  'text-(--canvas-foreground) bg-[var(--canvas-background,var(--color-background))]';
+
+/**
+ * Portals into xyflow's `edgelabel-renderer` div, which is a DOM sibling that
+ * always paints after every edge's own `<svg>`. Rendering the label as a plain
+ * `foreignObject` inside the edge's own `<g>` (the old approach) left it
+ * competing in the same per-edge z-index/DOM-order stack as every other edge's
+ * stroke, so a crossing unselected edge could paint over the label.
+ */
 export function EdgeLabel({ x, y, text, selected }: EdgeLabelProps) {
+  const transform = useMemo(() => `translate(-50%, -50%) translate(${x}px, ${y}px)`, [x, y]);
+
   return (
-    <foreignObject
-      x={x}
-      y={y}
-      width={1}
-      height={1}
-      overflow="visible"
-      style={{ pointerEvents: 'none' }}
-    >
+    <EdgeLabelRenderer>
       <div
-        className="react-flow__edge-label nodrag nopan"
-        style={{
-          position: 'absolute',
-          transform: 'translate(-50%, -50%)',
-          whiteSpace: 'nowrap',
-          pointerEvents: 'none',
-          color: 'var(--canvas-foreground)',
-          background: 'var(--canvas-background)',
-          padding: '4px 8px',
-          borderRadius: '4px',
-          fontSize: '12px',
-          fontWeight: 500,
-          border: `1px solid ${selected ? 'var(--canvas-primary)' : 'var(--canvas-border)'}`,
-          boxShadow: '0 1px 3px 0 rgba(0, 0, 0, 0.1)',
-        }}
+        className={cn(
+          EDGE_LABEL_BASE_CLASS,
+          selected ? 'border-(--canvas-primary)' : 'border-(--canvas-border)'
+        )}
+        style={{ transform }}
       >
         {text}
       </div>
-    </foreignObject>
+    </EdgeLabelRenderer>
   );
 }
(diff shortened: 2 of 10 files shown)


packages/apollo-react/src/canvas/components/StageNode/StageEdge.tsx
```diff
@@ -20,7 +20,7 @@ export const StageEdgeLabel = styled.div`
   pointer-events: all;
 
   &:hover {
-    background: var(--canvas-background-hover);
+    background: var(--canvas-background-hover, var(--color-background-hover));
     border-color: var(--canvas-border-hover);
   }
```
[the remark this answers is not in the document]
dbacomputer: Addressed in 24d49cb8: StageEdgeLabel now falls back to --color-border-hover when --canvas-border-hover is unavailable. The same commit also completes the default border and foreground fallback chain.

packages/apollo-react/src/canvas/components/Edges/shared/primitives/EdgeLabel.tsx
```diff
@@ -1,39 +1,42 @@
...
+
 export type EdgeLabelProps = {
   x: number;
   y: number;
   text: string;
   selected?: boolean;
 };
 
+// Falls back to --color-background when a host doesn't import canvas/styles/variables.css
+// (e.g. a shadow-DOM host), so the label never renders with a transparent background.
```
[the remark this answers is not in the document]
dbacomputer: Addressed in 24d49cb8: reworded the comment to describe the specific --color-background fallback without claiming transparency is impossible when every token is absent.

packages/apollo-react/src/canvas/components/StageNode/StageEdge.tsx
```diff
@@ -10,7 +10,7 @@ import { memo, useMemo } from 'react';
 export const StageEdgeLabel = styled.div`
   position: absolute;
   color: var(--canvas-foreground);
-  background: var(--canvas-background);
+  background: var(--canvas-background, var(--color-background));
```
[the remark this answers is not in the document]
dbacomputer: Addressed in 24d49cb8: StageEdgeLabel now falls back from canvas tokens to --color-foreground, --color-border, and --color-border-hover, in addition to the existing background fallbacks.

BenGSchulz (commented):
A couple stacking issues I am seeing:

1. A selected edge shows above its label (specific to this PR)

Image

2. An edge is not raised above others on hover (this seems to be existing, but could be nice to have fixed too)

Image

packages/apollo-react/src/canvas/components/Edges/shared/primitives/EdgeLabel.tsx
```diff
@@ -1,39 +1,57 @@
...
-          fontSize: '12px',
-          fontWeight: 500,
-          border: `1px solid ${selected ? 'var(--canvas-primary)' : 'var(--canvas-border)'}`,
-          boxShadow: '0 1px 3px 0 rgba(0, 0, 0, 0.1)',
-        }}
-      >
-        {text}
-      </div>
-    </foreignObject>
+    <EdgeLabelRenderer>
```
BenGSchulz, line 46: **F1 · P1 · Pillar 12 (UX / design system) + Pillar 5 (spec-vs-impl drift)**

The `EdgeLabelRenderer` move fixes crossing *unselected* edges, but not selected ones, and it introduces a self-overlap the old `foreignObject` didn't have. The PR body claims labels stay above every edge SVG "regardless of edge order or **selection state**" — the selection half doesn't hold.

Chain, all verified against the installed 12.8.2:

- `.react-flow__viewport` is a stacking context (`transform` + `z-index: 2`, `base.css:85-89`).
- `.react-flow__edges` is `position: absolute` with `z-index: auto` (`base.css:110-112`), so it is *not* a stacking context and each edge `<svg>`'s inline z-index participates directly in the viewport's context.
- `.react-flow__edgelabel-renderer` is `position: absolute` with `z-index: auto` (`base.css:282-292`), and nothing in apollo elevates it.
- `getElevatedEdgeZIndex` returns **1000** for a selected edge when `elevateOnSelect` (`@xyflow/system` `index.js:1003`), and `BaseCanvas.tsx:206` sets `elevateEdgesOnSelect={isDesignMode}`.

Per CSS 2.1 Appendix E, `z-index: auto`/`0` positioned descendants paint in step 8 and positive z-index in step 9. So a selected edge (1000) paints *after* the entire label layer: selecting a labeled edge in design mode draws its own stroke straight through its own label, and over any label it crosses. The old `foreignObject` rendered after the path inside the same `<svg>`, so the self-overlap case used to be correct.

Fix, combined with F2 since they share the same layer:

```css
/* xyflow puts the stacking z-index on the <svg> wrapping .react-flow__edge, so the
   hover bump must target that svg. z-index is inert on an SVG <g>. */
.react-flow__edges > svg:has([data-edge-hovered='true']) {
  z-index: 1001 !important;
}

/* Labels portal into a sibling div that defaults to z-index:auto, so any elevated
   edge (selected: 1000, hovered: 1001) paints over them. Lift the whole label layer
   above the edge band. Hovered node chrome (1002) still wins on DOM order. */
.react-flow__edgelabel-renderer {
  z-index: 1002;
}
```

One trade-off worth confirming before you take this: at 1002 labels also paint above non-hovered nodes (z-index 0), where today nodes win on DOM order. That's arguably the intent given the goal, but it changes the short-edge crowding case in the `Overflow` story, so it's worth an eyeball.

packages/apollo-react/src/canvas/styles/reactflow-reset.css
```diff
@@ -49,6 +49,15 @@
   z-index: 1002 !important;
 }
 
+/**
+ * Keep the edge currently under the pointer visually continuous at crossings.
+ * React Flow elevates selected edges to 1000; hover takes temporary precedence
+ * at 1001 while remaining below hovered node controls at 1002.
+ */
+.react-flow__edge:has([data-edge-hovered="true"]) {
```
BenGSchulz: **F2 · P1 · Pillar 12 + "verify against the installed library"**

This rule is a no-op. `EdgeWrapper` renders:

```js
jsx("svg", { style: { zIndex }, children: jsxs("g", { className: cc(['react-flow__edge', ...
```

(`@xyflow/react` `dist/esm/index.js:2872`)

So `.react-flow__edge` is the `<g>`, and `z-index` does not apply to SVG child elements — the effective stacking z-index sits on the parent `<svg>`. The neighbouring node rule at line 49 works precisely *because* `.react-flow__node` is a `<div>`, which is easy to over-generalize from.

Net effect: "The hovered edge also rises above intersecting edges so its full path remains traceable" (PR body, and the `CrossingLabeledEdges` story description at `EdgeLabel.stories.tsx:367`) documents behavior that never happens. The story text should be re-verified once the rule actually applies.

The `data-edge-hovered` attribute itself is correct — React stringifies booleans on `data-*`, so it does emit `"true"`. Only the selector target is wrong. Fix is the first block in F1.

packages/apollo-react/src/canvas/components/Edges/CanvasEdge.tsx
```diff
@@ -53,6 +53,30 @@ export const CanvasEdge = memo(function CanvasEdge({
   const onMouseEnter = useCallback(() => setIsHovered(true), []);
   const onMouseLeave = useCallback(() => setIsHovered(false), []);
   const pathRef = useRef<SVGPathElement | null>(null);
+  const { setEdges, setNodes } = useReactFlow();
+
+  const onLabelClick = useCallback(
```
BenGSchulz, line 58: **F3 · P1 · Pillar 2 (reuse) + Pillar 16 (React canvas)**

This isn't the xyflow-standard way to select an edge. xyflow's own `EdgeWrapper` does it through the store (`dist/esm/index.js:2819-2828`):

```js
const { addSelectedEdges, unselectNodesAndEdges, multiSelectionActive } = store.getState();
if (isSelectable) {
  store.setState({ nodesSelectionActive: false });
  if (edge.selected && multiSelectionActive) { unselectNodesAndEdges({ nodes: [], edges: [edge] }); }
  else { addSelectedEdges([id]); }
}
```

`addSelectedEdges` already does everything this handler hand-rolls: it reads `multiSelectionActive`, selects the edge, and deselects all other edges *and* all nodes (`index.js:3361-3369`). `setEdges` is the documented API for bulk graph mutation, not for selection — selection has dedicated store actions.

Four things measurably diverge:

1. **`elementsSelectable` / `edge.selectable` ignored.** `EdgeWrapper` gates on `isSelectable` (`index.js:2778,2822`). `BaseCanvas` exposes `elementsSelectable` as a public prop (`BaseCanvas.tsx:88,183`), so `elementsSelectable={false}` on a design-mode canvas is silently bypassed by a label click. The `isReadOnly` guard only covers `mode === 'readonly'`.
2. **`multiSelectionKeyCode` ignored.** Default is `isMacOs() ? 'Meta' : 'Control'` (`index.js:3533`). Hardcoding `metaKey || ctrlKey` makes macOS Ctrl+click additive (conventionally a context-menu gesture), and a consumer-configured `multiSelectionKeyCode` has no effect at all.
3. **`setNodes(...)` is redundant** — `addSelectedEdges` already clears node selection via `triggerNodeChanges(getSelectionChanges(nodeLookup, new Set(), true))` (`index.js:3368`).
4. **Change churn.** `edges.map` allocates a new object for *every* edge, so `getElementsDiffChanges` emits a `replace` change per edge (`index.js:797-817`) — N changes and N new identities per label click. `getSelectionChanges` emits `select` changes only for edges whose state actually changed (`index.js:778-796`).

```tsx
// After
import { useStoreApi } from '@uipath/apollo-react/canvas/xyflow/react';

const store = useStoreApi();

const onLabelClick = useCallback(
  (event: MouseEvent<HTMLDivElement>) => {
    event.stopPropagation();
    const {
      addSelectedEdges, unselectNodesAndEdges, multiSelectionActive, elementsSelectable, edgeLookup,
    } = store.getState();

    const edge = edgeLookup.get(id);
    if (!edge || !(edge.selectable ?? elementsSelectable)) return;

    store.setState({ nodesSelectionActive: false });
    if (edge.selected && multiSelectionActive) unselectNodesAndEdges({ nodes: [], edges: [edge] });
    else addSelectedEdges([id]);
  },
  [id, store]
);
```

This also drops `setNodes` and the `useReactFlow()` call entirely.

packages/apollo-react/src/canvas/components/Edges/shared/primitives/EdgeLabel.tsx
```diff
@@ -1,39 +1,57 @@
...
 };
 
-export function EdgeLabel({ x, y, text, selected }: EdgeLabelProps) {
+// Falls back to --color-background when the canvas-specific background token is unavailable,
+// such as in a host that doesn't import canvas/styles/variables.css.
+const EDGE_LABEL_BASE_CLASS =
+  'react-flow__edge-label nodrag nopan absolute top-0 left-0 max-w-48 overflow-hidden text-ellipsis ' +
+  'whitespace-nowrap pointer-events-auto cursor-default ' +
+  'px-2 py-1 rounded text-xs font-medium border shadow-[0_1px_3px_0_rgba(0,0,0,0.1)] ' +
+  'text-(--canvas-foreground) bg-[var(--canvas-background,var(--color-background))]';
```
BenGSchulz: **F4 · P2 · Pillar 12 (design tokens)**

The fallback is only half-applied. `StageEdge` got fallbacks on all five tokens (`StageEdge.tsx:12-24`), but here only the background did:

```tsx
// Before
'text-(--canvas-foreground) bg-[var(--canvas-background,var(--color-background))]';
borderColor = 'var(--canvas-border)',   // line 35

// After
'text-[var(--canvas-foreground,var(--color-foreground))] bg-[var(--canvas-background,var(--color-background))]';
borderColor = 'var(--canvas-border, var(--color-border))',
```

In the exact host the PR names (one that doesn't import `canvas/styles/variables.css`), `border-color: var(--canvas-border)` is invalid at computed-value time and falls back to `currentColor` — a visibly wrong border in the very scenario this PR set out to fix.

Worth noting the default-param fallback only covers the no-prop case: `EDGE_COLORS` (`shared/constants.ts:58-64`) are all bare `var(--canvas-*)`, so `borderColor={color}` stays unresolved in that host for every hover / selection / diff / validation / execution state. If the goal is genuine host-independence, the fallbacks belong in `EDGE_COLORS`.

packages/apollo-react/src/canvas/components/Edges/CanvasEdge.tsx
```diff
@@ -208,7 +233,10 @@ export const CanvasEdge = memo(function CanvasEdge({
             x={geometry.labelPoint.x}
             y={geometry.labelPoint.y}
             text={label}
-            selected={selected}
+            borderColor={color}
+            onClick={isReadOnly ? undefined : onLabelClick}
```
BenGSchulz, line 238: **F5 · P2 · Pillar 10 (testing discipline)**

`EdgeLabel.test.tsx` asserts the handlers are *forwarded*, but nothing covers the behavior this line gates: exclusive vs additive selection, node deselection, or the read-only `onClick: undefined` path. Given the F3 rewrite changes how selection is written, a test here is what keeps that refactor honest.

`SequenceEdge.test.tsx` already renders a real canvas, so this can seed off it rather than adding mocks:

```tsx
it('selects only the clicked edge and deselects nodes', () => {
  const onSelectionChange = vi.fn();
  renderCanvas({ mode: 'design', onSelectionChange }); // two labeled edges, one node selected

  fireEvent.click(screen.getByText('Alpha'));

  expect(onSelectionChange).toHaveBeenLastCalledWith({
    nodes: [],
    edges: [expect.objectContaining({ id: 'e1' })],
  });
});

it('does not select when read-only', () => {
  const onSelectionChange = vi.fn();
  renderCanvas({ mode: 'readonly', onSelectionChange });

  fireEvent.click(screen.getByText('Alpha'));

  expect(onSelectionChange).not.toHaveBeenCalled();
});
```

A third case worth locking once F3 lands: with `elementsSelectable={false}`, a label click must not select.

BenGSchulz (changes requested):
## /deep-review — architecture + runtime pass

**Verdict: request changes.** The `EdgeLabelRenderer` port is the right call and does fix the unselected-crossing case, but two of the three stacking behaviors described in the PR body don't hold in practice: a **selected** edge still paints over labels, and the new hover-elevation rule targets an element where `z-index` is inert.

Everything below was verified against the installed `@xyflow/react@12.8.2` / `@xyflow/system@0.0.66` rather than from memory, with `file:line` citations into `dist/esm/index.js` and `dist/base.css`.

**Slop meter:** clean — slop 1/100, tunnel-vision 0/100. The only hit is the deliberate `vi.mock` of the xyflow module in `EdgeLabel.test.tsx`, which is load-bearing (the portal target doesn't exist without a mounted `<ReactFlow>`) and correctly kept.

### Findings

| # | Sev | Location | Headline |
|---|-----|----------|----------|
| F1 | P1 | `EdgeLabel.tsx:43` | Selected edge (z-index 1000) still paints over every label in design mode |
| F2 | P1 | `reactflow-reset.css:57` | Hover z-index rule is inert: `.react-flow__edge` is a `<g>`, z-index lives on the wrapping `<svg>` |
| F3 | P1 | `CanvasEdge.tsx:58` | Hand-rolled `setEdges`/`setNodes` selection bypasses `elementsSelectable` and duplicates `addSelectedEdges` |
| F4 | P2 | `EdgeLabel.tsx:22` | Token fallback half-applied: background has one, text/border don't (StageEdge got all five) |
| F5 | P2 | `CanvasEdge.tsx:237` | New selection behavior has no test |

F1 and F2 share a single CSS fix, given inline on F1.

### Verified clean

Noting these so they don't get re-checked: `smartTooltip` genuinely gates on truncation (`CanvasTooltip.tsx:106,160`) and its `cloneElement` forwards the child's `onMouseEnter` (`CanvasTooltip.tsx:123-139`), so the hover-highlights-edge wiring survives Radix; `pointer-events-auto` is the correct counterpart to the label container's `pointer-events: none`; `EdgeLabel` has exactly one caller so dropping the `selected` prop is safe; every `--color-*` fallback token exists; `data-edge-hovered={isHovered || undefined}` does emit `"true"` (React stringifies booleans on `data-*`); and there are no em dashes in the new Storybook copy.

Three things I checked and am explicitly **not** filing, since they'd be reasonable guesses that turn out to be wrong: `useReactFlow().setEdges` *does* route through `onEdgesChange` for controlled flows (`BatchProvider`, `index.js:971-985`), so controlled canvases aren't broken; `onSelectionChange` still fires either way; and the un-cleared `nodesSelectionActive` is benign because `NodesSelection` returns `null` on empty bounds (`index.js:1927`).

BenGSchulz (approved):

PR #960, opened 2026-07-23
Add LockableValueField component and HITL panel

dbacomputer:
## Overview

Promotes `LockableValueField` from a one-off prototype into a documented, tested Apollo Wind component and updates the Apollo React Form HITL panel to consume it.

The Form HITL story now uses its standard typed control in Fixed value mode and a syntax-aware inline JavaScript editor in Expression mode. The Monaco suggestions overlay is configured to escape the compact field boundary instead of being clipped.

## Consumer integration

The API is now aligned with Flow Workbench’s adapter pattern.

Apollo Wind owns the field shell, locking, field-type selection, fixed/expression mode, and common value lifecycle. Consumers can supply their production editor through `renderExpressionEditor`, which receives the current value, `onValueChange`, `onBlur`, `readOnly`, placeholder, field ID, and field type.

This keeps Flow Workbench’s application-specific concerns in Flow Workbench—its `ExpressionValue` storage model, scoped `$vars` / `$metadata` / `$self` declarations, diagnostics, theming, variable insertion, and Monaco configuration—without coupling Apollo Wind to those services. The built-in value controls also forward the same blur contract for consistent touched-state, validation, and deferred-commit behavior.

## Apollo Wind — what to review

- `LockableValueField`: lockable read-only/editable states, typed fixed-value controls, and fixed/expression switching.
- Public API: `renderExpressionEditor` supports consumer-owned rich editors; `onValueBlur` and `fieldType` complete the adapter contract.
- Storybook organization: Lockable Value Field and Prompt Editor now live under `Components → UiPath`, with Lockable Value Field and Prompt Editor ordered above Canvas.
- Tests: coverage includes built-in value behavior, expression-editor context, blur forwarding, accessibility, invalid dates, options, and field actions.
- Date and effective-mode safeguards prevent invalid dates from throwing and prevent expression labels from appearing for types that cannot use expressions.

Screenshot 2026-07-27 at 2 34 22 PM

## Apollo React — what to review

- Form HITL consumes `LockableValueField` from `@uipath/apollo-wind` instead of a local prototype.
- Expression mode injects the inline Monaco editor while Fixed value mode retains the standard typed field.
- The suggestions widget uses fixed overflow positioning so variables and completions are not masked by the compact field row.
- The demo panel links to the canonical Lockable Value Field documentation under its new UiPath location.
- JSON schema edits are validated before applying them, preventing malformed hand edits from breaking the field list.

Form HITL

## Also fixed

Two pre-existing CI issues reproduced on `main` are included:

- Apollo Core, Apollo React, and Apollo Wind Biome configs now declare `"root": false` for Biome 2.x.
- PR checks set `TURBO_SCM_BASE` explicitly so affected-package detection resolves the PR base reliably.

## Test plan

- [x] Apollo Wind Vitest suite: 62 files, 1,062 tests passed
- [x] Filtered Apollo Wind / Apollo React type and build checks passed
- [x] Biome formatting passed on the updated files
- [x] Verified Form HITL Fixed value and Expression modes in Storybook
- [x] Verified Monaco suggestions render outside the compact field boundary
- [x] Verified Lockable Value Field and Prompt Editor navigation under `Components → UiPath`

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/packages/apollo-react/src/canvas/components/NodePropertyPanel/NodePropertyPanel.stories.tsx
+++ b/packages/apollo-react/src/canvas/components/NodePropertyPanel/NodePropertyPanel.stories.tsx
@@ -1,16 +1,48 @@
+import {
+  closestCenter,
+  DndContext,
+  type DragEndEvent,
+  type DragOverEvent,
+  DragOverlay,
+  type DragStartEvent,
+  KeyboardSensor,
+  PointerSensor,
+  useSensor,
+  useSensors,
+} from '@dnd-kit/core';
+import {
+  arrayMove,
+  SortableContext,
+  sortableKeyboardCoordinates,
+  useSortable,
+  verticalListSortingStrategy,
+} from '@dnd-kit/sortable';
+import { CSS } from '@dnd-kit/utilities';
 import type { EditorProps } from '@monaco-editor/react';
 import type { Meta, StoryObj } from '@storybook/react-vite';
-import type { FormSchema } from '@uipath/apollo-wind';
+import type { FormSchema, LockableFieldType, LockableValueFieldMode } from '@uipath/apollo-wind';
 import {
   Badge,
   Button,
+  Card,
+  CardContent,
   cn,
   DropdownMenu,
   DropdownMenuContent,
   DropdownMenuItem,
   DropdownMenuSeparator,
   DropdownMenuTrigger,
+  FIELD_TYPE_META,
+  HoverCard,
+  HoverCardContent,
+  HoverCardTrigger,
+  Input,
+  Label,
+  LockableValueField,
   MetadataForm,
+  Popover,
+  PopoverContent,
+  PopoverTrigger,
   ResizableHandle,
   ResizablePanel,
   ResizablePanelGroup,
@@ -20,6 +52,13 @@ import {
   TabsContent,
   TabsList,
   TabsTrigger,
+  Textarea,
+  ToggleGroup,
+  ToggleGroupItem,
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
 } from '@uipath/apollo-wind';
 import {
   apolloCoreDarkHCMonaco,
@@ -31,27 +70,37 @@ import {
 } from '@uipath/apollo-wind/editor-themes';
 import {
   ChevronDown,
+  ChevronsDownUp,
+  ChevronsUpDown,
   CircleAlert,
   CircleCheck,
+  CircleDot,
+  CircleOff,
   Code2,
+  Copy,
   Eye,
   File,
+  FileBracesCorner,
   GitFork,
   Globe,
   GripVertical,
   HardDrive,
+  Pencil,
   Play,
   Plus,
   ScanText,
+  Search,
   Sparkles,
   Trash2,
   Type,
   Upload,
+  UserRoundCheck,
   X,
   Zap,
 } from 'lucide-react';
 import type { CSSProperties, ReactNode } from 'react';
-import { lazy, Suspense, useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { lazy, Suspense, useCallback, useEffect, useId, useMemo, useRef, useState } from 'react';
+import { createPortal } from 'react-dom';
 import { NodeOutputModeSelect } from '../../controls';
 import type {
   DeriveTypeIcon,
@@ -639,6 +688,34 @@ const COMPACT_EDITOR_OPTIONS = {
   scrollbar: { vertical: 'auto', horizontal: 'hidden', alwaysConsumeMouseWheel: false },
   automaticLayout: true,
 } as const;
+
+const JSON_VIEWER_OPTIONS = {
+  readOnly: true,
+  fontSize: 12,
+  lineHeight: 18,
+  minimap: { enabled: false },
+  scrollBeyondLastLine: false,
+  wordWrap: 'off',
+  fontFamily:
+    'ui-monospace, SFMono-Regular, "SF Mono", Consolas, "Liberation Mono", Menlo, monospace',
+  padding: { top: 8, bottom: 8 },
+  lineNumbers: 'off' as const,
+  lineDecorationsWidth: 0,
+  glyphMargin: false,
+  folding: true,
+  renderLineHighlight: 'none' as const,
+  hideCursorInOverviewRuler: true,
+  overviewRulerBorder: false,
+  overviewRulerLanes: 0,
+  scrollbar: {
+    vertical: 'auto' as const,
+    horizontal: 'auto' as const,
+    alwaysConsumeMouseWheel: false,
+  },
+  automaticLayout: true,
+} as const;
+
+const JSON_EDITOR_OPTIONS = { ...JSON_VIEWER_OPTIONS, readOnly: false } as const;
 
 const INLINE_EDITOR_OPTIONS = {
   fontSize: 13,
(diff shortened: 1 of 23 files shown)


packages/apollo-wind/src/components/ui/lockable-value-field.tsx
BenGSchulz, line 1: There is quite a lot stacked into this file. Not sure the standards for apollo-wind, but I might imagine that this could be broken up into types, utils, and separate component files within a parent folder.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
```diff
@@ -0,0 +1,736 @@
...
+  SelectTrigger,
+  SelectValue,
+} from '@/components/ui/select';
+import { Switch } from '@/components/ui/switch';
+import { Textarea } from '@/components/ui/textarea';
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from '@/components/ui/tooltip';
+import { cn } from '@/lib';
+
+/**
+ * Raw Material Symbols Outlined glyphs (FILL 0, wght 400, GRAD 0, opsz 24),
```
BenGSchulz: Can we use lucide `lock-keyhole` / `lock-keyhole-open` icons instead of introducing these custom material-based ones?

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
```diff
@@ -0,0 +1,736 @@
...
+  id,
+  className,
+}: LockableValueFieldProps) {
+  const promptId = useId();
+  const generatedId = useId();
+  const fieldId = id ?? generatedId;
+  const [aiPrompt, setAiPrompt] = useState('');
+  const typeMeta = FIELD_TYPE_META[fieldType];
+  const effectiveMode = typeMeta.supportsExpression ? mode : 'fixed';
+  const collapsedTextClass = cn('@max-[259px]:hidden', compact && '!hidden');
```
BenGSchulz: This `259px` value shows up a lot and doesn't have much explanation. Maybe we can extract to a const / standard class value?

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
```diff
@@ -0,0 +1,736 @@
...
+                    <span
+                      className={cn(
+                        'text-xs font-medium',
+                        effectiveMode === 'expression' ? 'text-brand' : 'text-foreground'
+                      )}
+                    >
+                      Expression
+                    </span>
+                  </div>
+                  <span className="pl-[21px] text-[11px] leading-4 text-foreground-subtle">
```
BenGSchulz: These `pl-[21px]` could be replaced with `pl-6`.

I am starting to wonder about pulling out a more reusable dropdown styling since we seem to always be needing to duplicate how we style the items.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
```diff
@@ -0,0 +1,736 @@
...
+        ? value
+          ? formatDateValue(value)
+          : ''
+        : fieldType === 'single-select'
+          ? (options.find((option) => option.value === value)?.label ?? '')
+          : fieldType === 'multi-select'
+            ? parseListValue(value)
+                .map((v) => options.find((option) => option.value === v)?.label ?? v)
+                .join(', ')
+            : value;
```
CalinaCristian: this is hard to read / follow, should be refactored maybe into a method.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
```diff
@@ -0,0 +1,736 @@
...
+                className="flex-1"
+                onFilesChange={(files) => onValueChange?.(files.map((f) => f.name).join(', '))}
+              />
+            )
+          )}
+        </div>
+      )}
+    </div>
+  );
+}
```
CalinaCristian: there's a lot returned here, maybe it can be split up?

packages/apollo-wind/src/index.ts
```diff
@@ -58,6 +58,17 @@ export type {
   InputGroupTextareaProps,
 } from './components/ui/input-group';
 
+export {
+  FIELD_TYPE_META,
+  FIELD_TYPE_ORDER,
+  LockableValueField,
```
CalinaCristian, line 64: Stepping back before this becomes public API: a good chunk of the component's visible surface is stubbed today. Per the doc comment at `lockable-value-field.tsx:317`, expression mode has no real evaluation or highlighting, AI-assist is a no-op, Insert is dead (F2), uploads aren't persisted, and select options are demo data. That's fine in a prototype file; it's a harder sell exported from the design system, where the affordances read as capabilities.

The thing I can't see from here is the intended consumer. The obvious candidate — the HITL / property-panel surface — already has purpose-built equivalents (`CodeEditorField` with a real literal↔expression editor + cursor insertion, `ValueSourceField`, `VariablePicker`, field types from `@uipath/hitl-schema-types`), and nothing consumes `LockableValueField` today. apollo-wind also already ships `PromptEditor` for real expression/variable work, so this risks being a second value-field abstraction with a fake expression mode next to a real one.

Not asking to block on this — it's your + the team's call and there may be a roadmap reason to land the shell now. But two honest questions: (1) who's the first real consumer, and (2) should the stubbed affordances (expression mode, AI, Insert) be gated behind their callbacks so the component never renders a capability it can't deliver, rather than shipping them always-on? If the answer is "it's a presentational shell," scoping it to that explicitly would make it much easier to commit to as stable API.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
```diff
@@ -0,0 +1,736 @@
...
+  showFieldActions?: boolean;
+  /** Options for 'single-select' / 'multi-select' field types. Defaults to a small set of demo options. */
+  options?: { label: string; value: string }[];
+  /** Called with the entered prompt when the user clicks Generate in the AI-assist popover. */
+  onGenerateWithAi?: (prompt: string) => void;
+  id?: string;
+  className?: string;
+}
+
+const FIELD_LABEL: Record<LockableValueFieldMode, string> = {
```
CalinaCristian: These two maps still say "String" for every field type. They're keyed on `mode` but applied regardless of `fieldType`, so an `integer` field with no `label` renders a `type="number"` input placeholdered "String value", and a `boolean`/`date` in expression mode gets "Write a string expression". `typeMeta.label` already has the right word:

```tsx
const fieldLabel =
  effectiveMode === 'expression'
    ? `Write a ${typeMeta.label.toLowerCase()} expression`
    : `${typeMeta.label} value`;
```

A few tests assert `getByPlaceholderText('String value')` (`:10`, `:111`, `:153`, `:177`) — they'd move with it.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
```diff
@@ -0,0 +1,736 @@
...
+                        size="sm"
+                        className="w-full"
+                        disabled={!onGenerateWithAi}
+                        onClick={() => onGenerateWithAi?.(aiPrompt)}
+                      >
+                        Generate
+                      </Button>
+                    </PopoverContent>
+                  </Popover>
+                  <Tooltip>
```
CalinaCristian: This button has no `onClick` and no callback prop, so it's a dead control, and the `⌄` chevron implies a menu that never opens. It's really a symptom of the scope question in `index.ts` (D1): a real variable picker is context-aware and can't live in apollo-wind, so the component can't own this button meaningfully on its own.

If the component stays, the cleanest wiring is to let the consumer own the trigger+picker while the field keeps caret access — expose an imperative `insertAtCursor(text)` handle (matching how apollo-wind's own `PromptEditor` already works) and let consumers drop their button into the existing `headerActions` slot. Short of that, I'd remove the button rather than ship it inert.

CalinaCristian (commented):
🤖 **Claude deep-review** — automated two-lens architecture + runtime pass, posted for a human to weigh. Not a merge gate on its own.

No correctness blockers, and the tests/stories are solid. My main hesitation isn't a bug — it's a scope question about shipping this as public `apollo-wind` API in its current state, which I've left inline on the export. Two concrete items below (`F1` a real label bug, `F2` the inert Insert button, which is really a symptom of the scope question).

| # | File:Line | What |
|---|---|---|
| D1 | `packages/apollo-wind/src/index.ts:64` | Scope: should the stubbed affordances ship as public API before there's a consumer? |
| F1 | `packages/apollo-wind/src/components/ui/lockable-value-field.tsx:307` | Default label/placeholder hardcode "String" for every field type |
| F2 | `packages/apollo-wind/src/components/ui/lockable-value-field.tsx:538` | "Insert variable" button is a dead control |

packages/apollo-wind/src/index.ts
dbacomputer: LockableValueField is intended as a presentational shell: HITL is the first consumer, and a few other teams are lined up to adopt it directly for the same kind of locked/typed field UI rather than each building their own. Per your point, every stubbed affordance is now gated behind its own callback so it never renders a capability it can't deliver: AI-assist's Generate button is disabled without `onGenerateWithAi`, the mode dropdown trigger is disabled without `onModeChange`, and Insert-variable is now disabled unless a consumer supplies `variables` (see the sibling comment below — Insert now actually inserts). Select/multiselect options are also fully consumer-suppliable via `options` rather than defaulting to demo data in a real usage.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
dbacomputer: Wired this up: added a `variables` prop (`{ label, value }[]`, matching the shape of `options`). The Insert-variable button now opens a real dropdown listing `variables` and appends the clicked one to the current value via `onValueChange`, mirroring the InputEditor story's insertSnippet behavior you linked. It's disabled when `variables` is empty (the default) instead of rendering as a dead control.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
dbacomputer: Done — split into packages/apollo-wind/src/components/ui/lockable-value-field/ with types.ts, utils.ts, and a components/ subfolder (LockToggleButton, FieldHeader, ModeMenuItem), mirroring the prompt-editor package's folder convention. Main component file is now ~230 lines.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
dbacomputer: Addressed as part of the folder split -- the header row (label, type dropdown, required switch, AI-assist, insert-variable) is now its own FieldHeader component, and the value-control switch stays in the main component alongside the InputGroup/plain-Input branch.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
dbacomputer: Done — swapped the custom Material SVGs for lucide's LockKeyhole / LockKeyholeOpen.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
dbacomputer: Added a comment explaining where 259px comes from, and consolidated it into 3 named constants (collapsedTextClass, collapsedPaddingClass, compactOnlyClass) reused across all 4 call sites instead of the literal being duplicated. Couldn't extract it further into a JS-level const used inside the class strings themselves -- Tailwind's scanner needs the full class token to appear literally in source, so a template-interpolated `${BREAKPOINT}:hidden` silently wouldn't generate any CSS.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
dbacomputer: Swapped pl-[21px] for pl-6, and pulled the Fixed/Expression items into a shared ModeMenuItem component (packages/apollo-wind/src/components/ui/lockable-value-field/components/mode-menu-item.tsx) so the icon+label+description layout only exists in one place.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
dbacomputer: Extracted into getLockedDisplayValue(fieldType, value, options) in utils.ts, using a switch statement instead of the nested ternary chain.

packages/apollo-wind/src/components/ui/lockable-value-field.tsx
dbacomputer: Applied your suggested diff (computed fieldLabel from typeMeta.label instead of the hardcoded FIELD_LABEL/FIELD_PLACEHOLDER records), and updated the 4 affected tests that asserted 'String value' for non-string fieldTypes.

packages/apollo-wind/src/components/ui/lockable-value-field/types.ts
```diff
@@ -0,0 +1,141 @@
...
+  'multi-select': {
+    label: 'Multiselect',
+    icon: ListChecks,
+    supportsExpression: false,
+    fixedLabel: 'Fixed value',
+    fixedDescription: 'Choose one or more options',
+  },
+  file: {
+    label: 'File',
+    icon: Paperclip,
```
BenGSchulz: We have been using `File` as the icon for file-type fields in flow-workbench. If this is an explicit decision to change to `Paperclip` we should align all other usage too.

BenGSchulz (commented):
IMO we should still consider renaming the component / folder because there is a lot more functionality here than the lockable aspect of the field. This seems like a full inline field builder/designer (field-builder-inline?)

packages/apollo-wind/src/index.ts
CalinaCristian: Thanks — the "presentational shell" framing and the callback-gating both land well, and the gating change is exactly right.

One thing I want to surface concretely, since HITL is named as the anchor consumer: I looked at what "HITL adopts it directly" would actually mean, and I don't think the HITL schema surface can consume this as-is. That surface is `HitlSchemaEditor` (`@uipath/hitl-schema-editor`, wrapped in `hitl/HITLSchemaEditor.tsx`), and it's a different altitude of component — it edits a whole `HitlSchema` (a list of fields **plus** outcomes), with per-field `direction` (input/output/inOut) that feeds `node.data.outputs`, variable `binding` strings (`=vars.x`), conditional field `rules`, a JSON tab, preview, and persistence. `LockableValueField` is one field's value control, so it can't replace that editor; the only real overlap is the per-row fixed/expression value input.

And that per-row slot is already an apollo-wind consumer: the schema editor has a `useInputGroup` option that renders exactly that binding control via apollo-wind's `InputGroup`, plus `renderExpressionEditor` / `renderVariablePicker` render-props that inject the real Monaco editor and variable picker host-side. So for HITL specifically, adopting `LockableValueField` would mean an adapter from its flat `value: string` to the field's `binding`/`bindingMode` contract, mapping 7 field types onto the schema's 11 (`number`/`float`/`double`/`datetime`/`object`/`array` have no home here), and swapping a richer editor for a plainer one — i.e. net less than what's already wired.

None of this blocks the PR — it's a shell, and shipping the shell is your call. But it does soften the "HITL will adopt it directly" premise, so I'd want the first-consumer story to be someone whose field surface is genuinely flat-string-shaped, or to reframe this as a sibling to the schema editor's per-row control rather than something HITL replaces its editor with. If the longer-term goal is one shared primitive, the realistic path is the inverse: teach this component the `binding`/direction contract and the 11-type vocab, then let the schema editor consume it behind its existing `useInputGroup` seam.

packages/apollo-wind/src/components/ui/lockable-value-field/types.ts
dbacomputer: Good catch, thank you. This was an unreviewed choice on my part, not an intentional divergence from flow-workbench's `File` icon. Leaving this open rather than resolving it. I want to revisit icon consistency with the HITL team after this PR merges, since they're the ones driving these decisions going forward for their surface.

packages/apollo-wind/src/index.ts
dbacomputer: Agreed, the altitude mismatch with `HitlSchemaEditor` is a fair point and I don't want to overstate this as a drop-in replacement.

To be clear about the actual goal here: the HITL team asked for frontend engineering assistance building out this locked/typed field UI, which is what this PR delivers. Once it's in Storybook (targeting Monday), the HITL team will own further iteration and can adapt the integration however makes sense for their schema editor, including the adapter/binding-contract approach you outlined. We can also help them set it up differently if that turns out to be the better path. Appreciate you pressure-testing the framing.

packages/apollo-wind/src/components/ui/lockable-value-field/types.ts
dbacomputer: Updated the file field metadata to use the Lucide File icon, matching the Flow Workbench convention.

packages/apollo-wind/src/components/custom/panel-flow.tsx
```diff
@@ -219,11 +219,13 @@ function ToolCard({ card }: { card: FlowPanelToolCard }) {
...
-              {card.items.map((item, i) => {
+              {card.items.map((item) => {
                 const Icon = toolIconMap[item.icon];
                 return (
-                  // biome-ignore lint/suspicious/noArrayIndexKey: static display list
-                  <div key={i} className="flex items-start gap-2.5 pl-4 pr-9">
+                  <div
+                    key={`${item.icon}-${item.title}-${item.description}`}
+                    className="flex items-start gap-2.5 pl-4 pr-9"
+                  >
```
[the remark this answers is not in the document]
dbacomputer: Added an optional stable id to FlowPanelToolItem and a collision-safe indexed fallback for existing consumers.

packages/apollo-wind/src/components/ui/lockable-value-field/lockable-value-field.tsx
```diff
@@ -0,0 +1,274 @@
...
+              placeholder="Select options..."
+              disabled={!onValueChange}
+            />
+          ) : (
+            fieldType === 'file' && (
+              <FileUpload
+                id={fieldId}
+                ariaLabel={fieldLabel}
+                className="flex-1"
+                onFilesChange={(files) => onValueChange?.(files.map((f) => f.name).join(', '))}
```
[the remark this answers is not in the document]
dbacomputer: Added fileUploadAriaLabel, with automatic fallback to string labels and then the computed field label, plus coverage for custom ReactNode labels.

packages/apollo-wind/src/components/ui/file-upload.tsx
```diff
@@ -215,10 +235,13 @@ export function FileUpload({
       {/** biome-ignore lint/a11y/useSemanticElements: Intended */}
       <div
         role="group"
```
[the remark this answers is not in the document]
dbacomputer: Changed the keyboard-activated dropzone to role=button and moved the hidden file input beside it to avoid nested interactive controls.

packages/apollo-wind/src/components/ui/file-upload.test.tsx
```diff
@@ -35,10 +35,32 @@ describe('FileUpload', () => {
...
+      render(<FileUpload />);
+      expect(screen.getByRole('group')).toHaveAttribute('tabIndex', '0');
+    });
+
+    it('removes the dropzone from tab order and marks it aria-disabled when disabled', () => {
+      render(<FileUpload disabled />);
+      const dropzone = screen.getByRole('group');
+      expect(dropzone).toHaveAttribute('tabIndex', '-1');
+      expect(dropzone).toHaveAttribute('aria-disabled', 'true');
+    });
```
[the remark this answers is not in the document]
dbacomputer: Updated FileUpload tests to assert button semantics and accessible naming.

packages/apollo-wind/src/components/ui/lockable-value-field/lockable-value-field.test.tsx
```diff
@@ -0,0 +1,407 @@
...
+        id="tags-field"
+      />
+    );
+    expect(screen.getByLabelText('Tags')).toBeInTheDocument();
+  });
+
+  it("uses the field's computed label as the file upload area's accessible name", () => {
+    render(<LockableValueField locked={false} fieldType="file" />);
+    expect(screen.getByRole('group', { name: 'File value' })).toBeInTheDocument();
+  });
```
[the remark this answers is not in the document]
dbacomputer: Updated LockableValueField file-upload coverage to assert the dropzone button role.

packages/apollo-wind/src/components/ui/lockable-value-field/lockable-value-field.tsx
```diff
@@ -0,0 +1,285 @@
...
+            <Popover>
+              <PopoverTrigger asChild>
+                <button
+                  type="button"
+                  id={fieldId}
+                  data-slot="input-group-control"
+                  disabled={!onValueChange}
+                  onBlur={onValueBlur}
+                  className="flex h-full flex-1 items-center text-left text-sm text-foreground outline-none disabled:cursor-not-allowed disabled:opacity-50"
+                >
```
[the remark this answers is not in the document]
dbacomputer: Made the date popover controlled and now invoke onValueBlur when the calendar closes, rather than when focus moves into the portal. Added regression coverage.

packages/apollo-wind/src/components/ui/lockable-value-field/lockable-value-field.tsx
```diff
@@ -0,0 +1,292 @@
...
+            />
+          ) : fieldType === 'single-select' ? (
+            <Select
+              value={value || undefined}
+              onValueChange={onValueChange}
+              disabled={!onValueChange}
+            >
+              <SelectTrigger id={fieldId} className="flex-1" onBlur={onValueBlur}>
+                <SelectValue placeholder="Select an option" />
+              </SelectTrigger>
```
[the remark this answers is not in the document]
dbacomputer: Updated the single-select to report onValueBlur only when its controlled Radix menu closes, with regression coverage confirming opening the menu does not mark the field blurred.

packages/apollo-wind/src/components/ui/multi-select.tsx
```diff
@@ -69,15 +74,27 @@ const MultiSelect = React.forwardRef<HTMLDivElement, MultiSelectProps>(
...
+                    ? `${selected.length} ${selected.length === 1 ? 'item' : 'items'} selected`
+                    : placeholder
+              }
               className={cn(
                 'w-full justify-between future:rounded-xl future:border-0 future:bg-surface-overlay future:hover:bg-surface-hover future:font-normal future:text-muted-foreground',
                 selected.length > 0 ? 'h-auto min-h-10' : 'h-10'
               )}
               disabled={disabled}
+              onBlur={onBlur}
             >
```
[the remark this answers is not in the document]
dbacomputer: Updated MultiSelect to report onBlur when its popover closes after being opened, rather than when focus moves from the trigger into the portal. Added matching regression coverage.

packages/apollo-wind/src/components/ui/lockable-value-field/lockable-value-field.tsx
```diff
@@ -0,0 +1,298 @@
...
+  const generatedId = useId();
+  const [datePopoverOpen, setDatePopoverOpen] = useState(false);
+  const [selectOpen, setSelectOpen] = useState(false);
+  const fieldId = id ?? generatedId;
+  const typeMeta = FIELD_TYPE_META[fieldType];
+  const effectiveMode = typeMeta.supportsExpression ? mode : 'fixed';
+  const fieldLabel =
+    effectiveMode === 'expression'
+      ? `Write a ${typeMeta.label.toLowerCase()} expression`
+      : `${typeMeta.label} value`;
```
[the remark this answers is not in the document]
dbacomputer: Updated the expression label to choose a/an from the normalized field-type label, so integer now renders as Write an integer expression. Added coverage for both the consumer renderer payload and built-in input placeholder.

packages/apollo-wind/src/components/ui/file-upload.tsx
```diff
@@ -231,6 +268,7 @@ export function FileUpload({
         onDragOver={handleDragOver}
         onDragLeave={handleDragLeave}
         onDrop={handleDrop}
+        onBlur={onBlur}
         onClick={handleClick}
```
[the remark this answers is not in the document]
dbacomputer: Moved blur handling to the FileUpload widget boundary and now ignore focus transitions whose related target remains inside the widget. Added regression coverage for dropzone to Remove-button focus and for the eventual transition outside.

packages/apollo-wind/src/components/ui/lockable-value-field/components/field-header.tsx
```diff
@@ -0,0 +1,259 @@
...
+                    </TooltipTrigger>
+                    <TooltipContent>Insert variable</TooltipContent>
+                  </Tooltip>
+                  <DropdownMenuContent align="end" className="w-48">
+                    {variables.map((variable) => (
+                      <DropdownMenuItem
+                        key={variable.value}
+                        onClick={() =>
+                          onValueChange?.(value ? `${value} ${variable.value}` : variable.value)
+                        }
```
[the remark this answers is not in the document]
dbacomputer: The locked state now removes the header value-change capability before passing it to FieldHeader, which disables Insert Variable and provides a defensive no-op path while locked. Added regression coverage for variables plus an onValueChange callback on a locked field.

packages/apollo-wind/src/components/ui/lockable-value-field/lockable-value-field.tsx
```diff
@@ -0,0 +1,300 @@
...
+          <InputGroupAddon align="inline-start">
+            <LockToggleButton locked={locked} onLockedChange={onLockedChange} />
+          </InputGroupAddon>
+
+          {effectiveMode === 'expression' ? (
+            renderExpressionEditor ? (
+              renderExpressionEditor({
+                id: fieldId,
+                value,
+                onValueChange,
```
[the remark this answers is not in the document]
dbacomputer: Introduced a single editableOnValueChange capability that is undefined while locked, and now use it for the consumer expression renderer, built-in expression input, and header actions. Added regression coverage for both expression-editor paths.

packages/apollo-react/src/canvas/components/NodePropertyPanel/NodePropertyPanel.stories.tsx
```diff
@@ -1672,6 +1749,1768 @@ export const InputEditor: Story = {
...
+      if (!Array.isArray(parsed)) {
+        setJsonError('Expected a JSON array of fields.');
+        return;
+      }
+      if (!parsed.every(isValidLockableCase)) {
+        setJsonError('Each field needs id, title, required, value, locked, mode, and fieldType.');
+        return;
+      }
+      setCases(parsed);
+      setJsonError(null);
```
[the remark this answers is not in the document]
dbacomputer: Accepted Quick Form JSON now advances nextIdRef above the highest field ID. Validation also requires positive safe-integer IDs and rejects duplicates so drag-and-drop IDs remain unique before Add field is used.

CalinaCristian (approved):

PR #962, opened 2026-07-23
feat(apollo-vertex): Roadmap Status page with live Jira data

hfrancis31:
## Summary
- Adds a `/design-system-status` page surfaced as **"Roadmap status"** in the sidebar, positioned directly under Introduction
- Reads live from the VS Horizontal UX Jira board (`label = "horizontal-ux"`) and displays three sections: Recently Delivered, Coming Soon, and Backlog
- Coming Soon cards are sorted with In Review first, In Progress after
- Delivered section is curated: only tickets with the `ds-delivered` Jira label appear
- Epic name shown on cards when the ticket has a parent Epic
- Link resolution for delivered tickets: explicit `Vertex: <url>` in description (supports both plain text and Jira smartlinks) → convention slug lookup → Jira URL fallback

## Environment variables required
Add to `.env.local` locally and to the Vercel project for production:
```
JIRA_BASE_URL=https://uipath.atlassian.net
JIRA_EMAIL=<your-email>
JIRA_API_TOKEN=<your-api-token>
```

## How to add tickets to the Delivered section
Add the `ds-delivered` label to the Jira ticket. No code change needed.

## How to set the Vertex link on a delivered card
Add `Vertex: https://...` as a line in the Jira ticket description.

## Test plan
- [ ] Visit `/design-system-status` and confirm three sections render with live data
- [ ] Confirm "In Review" cards appear at the top of Coming Soon
- [ ] Confirm the Delivered section shows only the `ds-delivered`-labelled ticket(s)
- [ ] Confirm epic names appear on cards that have a parent Epic
- [ ] Confirm clicking a delivered card with a `Vertex:` URL navigates to the Vertex page (not Jira)
- [ ] Confirm page loads gracefully without env vars (shows setup error state)

### DIFF
--- a/apps/apollo-vertex/app/api/jira/route.ts
+++ b/apps/apollo-vertex/app/api/jira/route.ts
@@ -1,0 +1,22 @@
+import { type NextRequest, NextResponse } from "next/server";
+import { fetchJiraIssues } from "@/lib/jira";
+import { processIssues } from "@/lib/jira-resolve";
+
+// Intentionally public — this endpoint surfaces cross-team design system status.
+// No auth guard is by design; confirmed with @ruudandriessen.
+export async function GET(_req: NextRequest) {
+  try {
+    const issues = await fetchJiraIssues();
+    const jiraBaseUrl =
+      process.env.JIRA_BASE_URL ?? "https://uipath.atlassian.net";
+    const data = processIssues(issues, jiraBaseUrl);
+    return NextResponse.json(data, {
+      headers: {
+        "Cache-Control": "public, s-maxage=300, stale-while-revalidate=60",
+      },
+    });
+  } catch (e) {
+    const message = e instanceof Error ? e.message : String(e);
+    return NextResponse.json({ error: message }, { status: 500 });
+  }
+}
--- a/apps/apollo-vertex/lib/jira-resolve.ts
+++ b/apps/apollo-vertex/lib/jira-resolve.ts
@@ -1,0 +1,259 @@
+import { extractAdfText, type JiraIssue } from "./jira";
+
+// All known routable slugs per section, derived from the site's _meta.ts files
+const SLUGS_BY_SECTION: Record<string, string[]> = {
+  components: [
+    "accordion",
+    "alert",
+    "alert-dialog",
+    "aspect-ratio",
+    "avatar",
+    "badge",
+    "breadcrumb",
+    "button",
+    "button-group",
+    "calendar",
+    "card",
+    "carousel",
+    "chart",
+    "line-chart",
+    "multi-line-chart",
+    "bar-chart",
+    "distribution-chart",
+    "kpi-chart",
+    "table-chart",
+    "checkbox",
+    "collapsible",
+    "combobox",
+    "command",
+    "context-menu",
+    "data-table",
+    "date-picker",
+    "dialog",
+    "drawer",
+    "dropdown-menu",
+    "empty",
+    "field",
+    "feature-flags",
+    "filter-dropdown",
+    "form",
+    "form-wizard",
+    "hover-card",
+    "input",
+    "input-group",
+    "input-otp",
+    "item",
+    "kbd",
+    "label",
+    "menubar",
+    "navigation-menu",
+    "pagination",
+    "popover",
+    "progress",
+    "radio-group",
+    "resizable",
+    "scroll-area",
+    "select",
+    "separator",
+    "sheet",
+    "sidebar",
+    "skeleton",
+    "slider",
+    "sonner",
+    "spinner",
+    "switch",
+    "table",
+    "tabs",
+    "textarea",
+    "toggle",
+    "toggle-group",
+    "tooltip",
+  ],
+  patterns: [
+    "ai-chat",
+    "feedback-vote-widget",
+    "metric-card",
+    "page-header",
+    "shell",
+  ],
+  templates: ["list-page", "settings", "solution-tests"],
+  guidelines: ["ai-toolkit", "notifications"],
+  foundation: ["colors", "spacing", "grid", "typography", "icons", "logos"],
+};
+
+// Flat map: slug → absolute path
+const SLUG_PATH_MAP = new Map<string, string>();
+for (const [section, slugs] of Object.entries(SLUGS_BY_SECTION)) {
+  for (const slug of slugs) {
+    SLUG_PATH_MAP.set(slug, `/${section}/${slug}`);
+  }
+}
+
+export type LinkSource = "explicit" | "convention" | "jira";
+export type Section = "delivered" | "coming-soon" | "backlog";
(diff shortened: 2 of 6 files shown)


apps/apollo-vertex/app/api/jira/route.ts
0xr3ngar, line 1: should this even be in the apollo-vertex? shouldn't we just mock jira tickets in the UI and explicitly tell the consumers to connect it to the actual JIRA API. I don't think this needs to be here

apps/apollo-vertex/lib/jira.ts
0xr3ngar, line 1: instead of hand rolling our own jira processing can't we just use a sdk npm package ? like

https://github.com/jira-node/node-jira-client

apps/apollo-vertex/lib/jira-resolve.ts
0xr3ngar, line 1: lets split this file up in smaller functions that each handle 1 thing on their own.

apps/apollo-vertex/lib/jira-resolve.ts
```diff
@@ -0,0 +1,225 @@
...
+  linkStats: { explicit: number; convention: number; jiraFallback: number };
+}
+
+function toSection(statusName: string): Section {
+  const s = statusName.toLowerCase();
+  if (s === "closed" || s === "done") return "delivered";
+  if (s === "in progress" || s === "in review" || s === "review")
+    return "coming-soon";
+  return "backlog";
+}
```
0xr3ngar, line 130: I would use a record

```ts
const SECTION_BY_STATUS = {
  closed: "delivered",
  done: "delivered",
  "in progress": "coming-soon",
  "in review": "coming-soon",
  review: "coming-soon",
} satisfies Record<string, Section>;

function toSection(statusName: string): Section {
  return SECTION_BY_STATUS[statusName.trim().toLowerCase()] ?? "backlog";
}
```

apps/apollo-vertex/lib/jira-resolve.ts
```diff
@@ -0,0 +1,225 @@
...
+
+function toBadge(labels: string[]): BadgeLabel {
+  if (labels.includes("ai-legal-required")) return "required";
+  if (labels.includes("ai-legal-best-practice")) return "best-practice";
+  return null;
+}
```
0xr3ngar: this would be cleaner if it's 

```ts
const BADGE_BY_LABEL = {

  "ai-legal-required": "required",

  "ai-legal-best-practice": "best-practice",

} as const satisfies Record<string, Exclude<BadgeLabel, null>>;



const BADGE_LABEL_PRIORITY = [

  "ai-legal-required",

  "ai-legal-best-practice",

] as const;



function toBadge(labels: readonly string[]): BadgeLabel {

  const label = BADGE_LABEL_PRIORITY.find((candidate) =>

    labels.includes(candidate),

  );



  return label ? BADGE_BY_LABEL[label] : null;

}
```

apps/apollo-vertex/lib/jira-resolve.ts
```diff
@@ -0,0 +1,225 @@
...
+        .replace(/[^a-z0-9]+/g, "-")
+        .replace(/(^-|-$)/g, "");
+      const path = SLUG_PATH_MAP.get(slug);
+      if (path) return { url: path, source: "convention" };
+    }
+  }
+
+  // Priority 3: fall back to Jira ticket
+  return { url: jiraUrl, source: "jira" };
+}
```
0xr3ngar, line 179: all of this regex processing starts making me question if we should really process the description at all

0xr3ngar: The best version, is to avoid parsing the description entirely. Jira should ideally expose these as structured custom fields

```ts
issue.fields.vertexUrl
issue.fields.componentName
```

Then the function shrinks to
```ts
function resolveDeliveredLink(
  issue: JiraIssue,
  jiraUrl: string,
): { url: string; source: LinkSource } {
  if (issue.fields.vertexUrl) {
    return {
      url: issue.fields.vertexUrl,
      source: "explicit",
    };
  }

  const path = issue.fields.componentName
    ? PATH_BY_COMPONENT[issue.fields.componentName]
    : undefined;

  return path
    ? { url: path, source: "convention" }
    : { url: jiraUrl, source: "jira" };
}
```

apps/apollo-vertex/lib/jira-resolve.ts
```diff
@@ -0,0 +1,225 @@
...
+      const sl = statusName.toLowerCase();
+      if (sl === "in review" || sl === "review") comingSoon.unshift(card);
+      else comingSoon.push(card);
+    } else {
+      backlog.push(card);
+    }
+  }
+
+  return { delivered, comingSoon, backlog, linkStats };
+}
```
0xr3ngar, line 259: this does way to many things in one place, this let link mutability is a smell. Tell your claude to extract this into seperate small functions so `processIssues` becomes an orchestrator

0xr3ngar (approved):
left some comments, address them with Claude, if you need my help lmk.

Also CI checks are failing, please fix before re-requesting review.

0xr3ngar (changes requested):
I accidentally approved oops

apps/apollo-vertex/app/api/jira/route.ts
hfrancis31: I discussed with @ruudandriessen that this is needed so that we can provide visibility to all eng teams on what is coming and the status. We are wanting more visibility across teams.

apps/apollo-vertex/lib/jira.ts
hfrancis31: Thanks for the suggestion. I looked into it. The main reasons I'm keeping the raw fetch approach here:

- Single endpoint: we only call POST /rest/api/3/search/jql. Adding a full SDK to wrap one request adds dependency weight without meaningful simplification.
- v3 cursor pagination: we use nextPageToken, which is specific to the REST v3 API. Most Jira JS SDKs still target v2 (startAt/maxResults) and would require workarounds.
- Intentionally narrow types: JiraIssue only types the fields we actually fetch. SDK-generated types cover the full response shape, which adds noise for no gain here.

Happy to revisit if the integration scope grows, but for a single read-only JQL query this felt like the right trade-off.

apps/apollo-vertex/lib/jira-resolve.ts
hfrancis31: For now, keeping the description-based Vertex: <url> convention as a pragmatic workaround to get the board live and visible. Since we don't have this available right now, I'm working with an admin to create one. Once it's available, I'll update lib/jira.ts to fetch the field directly and simplify the parsing in jira-resolve.ts.

0xr3ngar (approved):

pieman1313 (approved):

petervachon:
Opening back up to add in API key

PR #981, opened 2026-07-27
ci(apollo-vertex): serve the component registry from the Coded App, drop Vercel

fikewa-olatunji:
## Why

#978 restored a **Vercel-only** deploy of Apollo Vertex purely to keep the component registry updated after production moved to Coded Apps (#896). That extra pipeline isn't needed: the Coded App build already runs `registry:build`, so the registry is published at `engdogfood.staging.uipath.host/apollo-vertex/r/*.json` on every deploy and stays in sync automatically (verified live, returns valid JSON).

## What changed

- `registry.json` `homepage` → Coded App URL
- Docs `@uipath` registry base + component links (`ai-chat`, `feedback-vote-widget`, `mcp`, `localization`) → Coded App URL
- Removed the Vercel-only registry deploy workflow added in #978

## Notes

- Registry item JSON is **host-agnostic** (no baked URL); external consumers set the `@uipath` base in their own `components.json`, so this is a canonical-URL change, not a per-item rewrite.
- `npx shadcn add` is a Node fetch (no CORS), and the Coded App host serves the registry anonymously.
- The `engdogfood` Coded App URL is permanent, so it's safe as the canonical registry endpoint.
- No separate pipeline after this — the registry republishes with each Coded App deploy.


frankkluijtmans (approved):
Ah, we did actually deploy the registry itself in a coded app 😅 Looks like just the doc updates we're missing 😁

CalinaCristian (approved):

PR #986, opened 2026-07-29
feat(apollo-react): ability to add stages rules to stage node, collapsible sections

joelasher-ap:
This adds the ability to add the entry/exit rules of the Stage to the Stage node so that they can be displayed in the node. This also includes collapsible section headers to help control the node height. This is all for a spike for Mounish, and only shows when the new optional properties are provided.

This also cleans up the folder structure of the Stage Node components. Previously it was all under one folder which was messy, now it's spread out across a few sub-folders to keep it a bit cleaner

Screenshot 2026-07-29 at 3 20 54 PM
Screenshot 2026-07-29 at 3 21 02 PM


apps/apollo-vertex/next-env.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/dev/types/routes.d.ts";
+import "./.next/types/routes.d.ts";
```
Miyuki-L: is there a reason this changes? Its in vertex

joelasher-ap: 👀 good catch, yea I don't think this should have changed

packages/apollo-react/src/canvas/components/StageNode/shared/StageCompletionConditionIcon.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const StageCompletionConditionIcon = ({
+  dataTestId,
+  small,
+}: {
+  dataTestId?: string;
+  small?: boolean;
+}) => {
+  return (
+    <CanvasTooltip content="Completion condition" placement="top">
```
Miyuki-L, line 12: think we have localization in apollo ui now do we want to localize the tooltip contents? For this and the other icon files too

packages/apollo-react/src/canvas/components/StageNode/shared/CollapsibleStageHeader.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  &:hover {
+    background-color: var(--color-background-hover);
+  }
+
+  &:focus-visible {
+    outline: 2px solid var(--color-primary);
+    outline-offset: 2px;
+  }
+`;
```
Miyuki-L, line 22: Remember Kitty mentioning before that we don'g want to use Styled Component it has some negative impact on the perf

packages/apollo-react/src/canvas/components/StageNode/rules/StageRuleContent.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    expect(screen.getByTestId('exit-condition-icon')).toBeInTheDocument();
+    expect(screen.queryByTestId('entry-condition-icon')).not.toBeInTheDocument();
+  });
+
+  it('renders the completion-condition icon for a completion rule', () => {
+    render(<StageRuleContent rule={rule} ruleType="completion" />);
+
+    expect(screen.getByTestId('completion-condition-icon')).toBeInTheDocument();
+    expect(screen.queryByTestId('entry-condition-icon')).not.toBeInTheDocument();
+  });
```
Miyuki-L, line 59: nit: do we want to check that Complete for the first one and exit for the second one aren't there? Since we did that for the entry condition icon not for these two

packages/apollo-react/src/canvas/components/StageNode/rules/ExitRulesContainer.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  let rulesContainer = (
+    <StageRulesContainer rules={exitRules} ruleType="exit" isReadOnly={isReadOnly} />
+  );
+
+  if (sectionStates?.exitRules) {
+    rulesContainer = (
+      <CollapsibleStageHeader
+        isOpen={!sectionStates.exitRules.isCollapsed}
+        label="Exit rules"
```
Miyuki-L, line 28: same here? should we use localization

packages/apollo-react/src/canvas/components/StageNode/rules/ExitRulesContainer.tsx
```diff
@@ -0,0 +1,45 @@
...
+        testId={`exit-rules-header-${id}`}
+        onToggle={sectionStates.exitRules.onCollapsedToggle}
+      >
+        {rulesContainer}
+      </CollapsibleStageHeader>
+    );
+  } else {
+    rulesContainer = (
+      <StageItemsSection>
+        <StageItemsHeaderTitle title="Exit rules" testId={`exit-rules-header-${id}`} />
```
Miyuki-L, line 38: loc?

packages/apollo-react/src/canvas/components/StageNode/rules/EntryRulesContainer.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  let rulesContainer = (
+    <StageRulesContainer rules={entryRules} ruleType="entry" isReadOnly={isReadOnly} />
+  );
+
+  if (sectionStates?.entryRules) {
+    rulesContainer = (
+      <CollapsibleStageHeader
+        isOpen={!sectionStates.entryRules.isCollapsed}
+        label="Entry rules"
```
Miyuki-L, line 28: same here loc?

packages/apollo-react/src/canvas/components/StageNode/rules/EntryRulesContainer.tsx
```diff
@@ -0,0 +1,45 @@
...
+        testId={`entry-rules-header-${id}`}
+        onToggle={sectionStates.entryRules.onCollapsedToggle}
+      >
+        {rulesContainer}
+      </CollapsibleStageHeader>
+    );
+  } else {
+    rulesContainer = (
+      <StageItemsSection>
+        <StageItemsHeaderTitle title="Entry rules" testId={`entry-rules-header-${id}`} />
```
Miyuki-L, line 38: same here loc?

packages/apollo-react/src/canvas/components/StageNode/StageNode.stories.tsx
Miyuki-L, line 1: I know we control this through what we pass to the stages but for the stories do we still want to show the header chips for the rules when we show the rules section? They're duplicate info and also the Figma Mounish shared didn't have the chips.

packages/apollo-react/src/canvas/components/StageNode/shared/StageCompletionConditionIcon.tsx
joelasher-ap: Because this is just for a Spike Prototype and not something we're committing I don't want to localize yet.

packages/apollo-react/src/canvas/components/StageNode/rules/ExitRulesContainer.tsx
joelasher-ap: Because this is just for a Spike Prototype and not something we're committing I don't want to localize yet.

packages/apollo-react/src/canvas/components/StageNode/rules/ExitRulesContainer.tsx
joelasher-ap: Because this is just for a Spike Prototype and not something we're committing I don't want to localize yet.

packages/apollo-react/src/canvas/components/StageNode/rules/EntryRulesContainer.tsx
joelasher-ap: Because this is just for a Spike Prototype and not something we're committing I don't want to localize yet.

packages/apollo-react/src/canvas/components/StageNode/rules/EntryRulesContainer.tsx
joelasher-ap: Because this is just for a Spike Prototype and not something we're committing I don't want to localize yet.

packages/apollo-react/src/canvas/components/StageNode/StageNode.stories.tsx
joelasher-ap: I think this is something we could want to do, but because this is just a prototype spike I just want to get the rules to appear so we can show them to the PMs

packages/apollo-react/src/canvas/components/StageNode/shared/CollapsibleStageHeader.tsx
joelasher-ap: I saw that Apollo was still using a lot of `styled` 🤔 @kittyyueli

packages/apollo-react/src/canvas/components/StageNode/StageNode.stories.tsx
joelasher-ap: I also think that sometimes the rules count in the header chips doesn't always match the number of rules in the stage? I think theres some extra calculation that goes into that number shown in the chips. It's something we can return to if we're actually gonna show the rules in the stage past this spike

packages/apollo-react/src/canvas/components/StageNode/shared/CollapsibleStageHeader.tsx
Miyuki-L: Think more so to stop adding to it and eventually move off.

kittyyueli: yep, we should move off of styled so for now let's stop adding to it

Miyuki-L (approved):
Since its in the Collapsible Header part we can come back to it once the spike is actually adopted since its not rendered unless adopted.

Miyuki-L (approved):

PR #965, opened 2026-07-24
feat(apollo-react): collapse stage-node task execution to one line with runs/rework icons

BenNormann:
## Summary

Collapses each stage-node task's execution state onto a **single line** so stage nodes stop
growing tall during execution (Maestro Case Management). Per Carl Schultze's Figma
("Debug on single line", node 19687-68917): duration moves inline, and runs/rework become
compact amber icon-chips with detail in tooltips instead of full-text badges wrapping to a
second row.

This is the apollo-ui side of **MST-12724**. PO.Frontend consumes it via a later version bump
(no PO.Frontend change required — it already supplies the data props).

## Demo

Screenshot 2026-07-24 at 10 00 35 AM
image
image
image

## Changes

- `TaskContent.tsx`: rewrote the two-row `Column` into a single `Row` (removed the
  `hasSecondRowContent` second-row branch).
- Runs → compact amber `Badge` with a `Redo2` (↱) icon + total-run count (`retryCount + 1`);
  tooltip "Ran N times" / "Running again" (localised).
- Rework → compact amber `Badge` with a `RotateCcw` (↺) icon, shown when there is rework time;
  tooltip "Reworked (+{duration})" — this is where the old inline `(+25m)` moved.
- Removed the never-used plain consumer-`badge` text chip (e.g. "Action needed").
- Ad-hoc tasks no longer render the required `*`; entry-condition diamond kept.
- Play button unified to a single far-right render with the status icon to its left.
- New i18n message `stage-node.task-badge.rework-time`.
- Existing `data-testid`s (`stage-task-*`, `stage-task-play-*`) unchanged.

## Flow

```mermaid
flowchart LR
    D[StageTaskExecution] --> R[Single Row]
    D -->|duration| Dur[duration text]
    D -->|retryCount > 0| Runs["↱ runs chip + count<br/>tooltip: Ran N times"]
    D -->|retryDuration| Rew["↺ rework chip<br/>tooltip: Reworked +time"]
    D -->|status| St[status icon]
    Dur --> R
    Runs --> R
    Rew --> R
    St --> R
    R --> P[play button · far right]
```

## E2E Impact

N/A for this repo (the E2E-impact check is a PO.Frontend tool). Change is a visual refactor of an
existing component; all `data-testid`s are preserved, so consumer/E2E selectors are unaffected.

## Testing

- [x] `vitest` — StageNode suite green (192 tests, incl. rewritten TaskContent runs/rework/
      required-marker/badge-suppression suites)
[...]

PR #989, opened 2026-07-30
feat(apollo-react): tighten stage task spacing and put debug breakpoints on the gutter dot

BenNormann:
## Summary

Follow-up to #965 (**MST-12724**). #965 collapsed stage-node task execution onto one line; this
tightens the spacing that shook out from it, then picks up the debug-view items agreed in the
[design thread](https://uipath-product.slack.com/archives/C090NTF6LRL/p1784216031237709): drop the
task `⋮` in debug mode, and put breakpoints on the gutter dot itself.

Three commits, reviewable independently.

## Demo

image
image
image

### `fix: constrain stage task metadata spacing`

- Entry-condition diamond moves to the leading icon group; `EntryConditionIcon` gains `tight`,
  cropping the viewBox to the painted diamond so its dead margin stops padding the row. Replaces
  the old `small` prop and its `translateX` nudge.
- `StageTaskIcon` collapses its wrapper for plain SVGs (`&:has(> svg)`) — 16px, not a transparent
  24px box.
- New `trailingContent` slot on `TaskContent`; `DraggableTask` passes `TaskMenu` through it so the
  `⋮` sits in the actions row and inherits its gap.
- `TaskMenu` trigger sized to the glyph, via a new optional `triggerClassName` on
  `CanvasDropdownMenu` (default unchanged).
- New `formatDuration` caps a formatted duration at its 3 largest units, preserving prefixes and
  separators. Applied to the task duration.
- New testids `stage-task-identity-{id}` / `stage-task-actions-{id}`.

### `fix: cap stage duration at its three largest units`

Same helper applied to the stage header — per Carl's "Yes" on the thread.

### `feat: put debug-view breakpoints on the task gutter dot`

- **`⋮` dropped in the Debug view.** Add/remove breakpoint is the only action left there, so it
  doesn't earn a permanent affordance; right-click still opens the menu. `TaskMenu` keeps an
  invisible anchor sized to the box the button occupied — see the note below on why.
- **The gutter dot is now the breakpoint control.** New `onTaskBreakpointToggle` prop: a set
  breakpoint clicks to remove, and hovering a task without one ghosts a dot in to click. Follows
  Flow's `ExecutionBreakpoint` for opacity (0.5), transition (`0.2s ease-in-out`) and tooltip
  wording. Wired only when the stage is read-only, so design time keeps a quiet hover state.
- Existing testid `stage-task-breakpoint-{id}` still means "set"; the affordance is
  `stage-task-add-breakpoint-{id}`.

## Anchor placement

Worth a look in review. Radix opens the popup against its trigger, so hiding the `⋮` still needs an
anchor. A zero-height one at the row's top opens the menu **under the cursor**, and the mouse-up
that ends a right-click then selects an item instantly — caught in Storybook, invisible to jsdom.
The anchor now reproduces the button's box (16px tall, centred, right edge on the row's content
edge). Measured in-place: `deltaRight 0, deltaTop 0` against where the button would be, and the
menu lands at an identical offset to design mode.

## Not included

Alfredo is taking the remaining thread items: removing the stop icon from the "add breakpoint"
label, and the pulsating effect on an active breakpoint.

No loading state on the toggle — Flow's version awaits a server call and shows a spinner; this one
fires synchronously and relies on the consumer re-rendering `execution.taskStatus[id].breakpoint`.
Worth adding if the PO.Frontend toggle goes to the network.

## E2E Impact

N/A for this repo. No visual-regression or snapshot suite in `apollo-react`. All testids from
`main` are preserved; four are added.

## Testing

- [x] `vitest --run src/canvas` — 96 files, 1834 tests passing
- [x] Manual in Storybook: verified the menu placement matches design mode, right-click no longer
      fires an item, and the gutter dot adds/removes in both directions
- [x] `biome lint` + `format` clean on changed paths; `tsc --noEmit` clean
- [ ] Preview of the hidden-`⋮` case on ad-hoc / event-driven tasks — no read-only story covers
      those two; the anchor resolves against `StageTask` in all three, so it holds by construction


packages/apollo-react/src/canvas/components/StageNode/formatDuration.ts
```diff
@@ -0,0 +1,23 @@
+const DURATION_UNIT_PATTERN = /\d+(?:[.,]\d+)?\s*(?:ms|mo|w|d|h|m|s)\b/gi;
+
+/**
+ * Limits an already formatted, largest-to-smallest duration to its largest units.
+ * Prefixes and separators are preserved (for example, "Duration: 6m, 2w, 2d").
+ */
+export function formatDuration(duration: string, maxUnits = 3): string {
```
kittyyueli: Is the duration passed into the node like 6m, 2w, 2d, etc. etc.? If so then we should be passing in like num in MS and doing the string calculations in here instead

packages/apollo-react/src/canvas/components/StageNode/formatDuration.test.ts
```diff
@@ -0,0 +1,59 @@
...
+const MINUTE = 60 * SECOND;
+const HOUR = 60 * MINUTE;
+const DAY = 24 * HOUR;
+
+describe('formatDurationMs', () => {
+  it('keeps only the three largest units', () => {
+    // 2 days, 3 hr, 4 min, 5 sec -> the seconds are dropped.
+    const ms = 2 * DAY + 3 * HOUR + 4 * MINUTE + 5 * SECOND;
+
+    expect(formatDurationMs(ms)).toBe('2 days, 3 hr, 4 min');
```
alfredo-mruiz, line 14: We could benefit on having a test checking for some locales, like Japanese or German.

packages/apollo-react/src/canvas/components/StageNode/StageNode.styles.ts
```diff
@@ -209,6 +222,13 @@ export const StageTaskIcon = styled.div`
   justify-content: center;
   flex-shrink: 0;
 
+  /* Plain SVG task icons paint at 16px, so collapse their otherwise transparent
+     24px wrapper. TaskIcon components with a visible 24px container keep it. */
+  &:has(> svg) {
```
alfredo-mruiz: This rule affects every task whose icon is a plain SVG. can we do a quick visuall pass over existing tasks?

alfredo-mruiz (approved):

CalinaCristian (approved):

PR #991, opened 2026-07-31
ci(ci): storybook visual diff check + extract inline workflow scripts [PLT-108380]

CalinaCristian:
## What

Two related CI changes:

1. **Storybook visual diff check** (informational, never merge-blocking): every PR sync compares the PR's `apollo-design` preview Storybook against the deployed main Storybook using [`@uipath/storybook-visual-diff`](https://github.com/UiPath/storybook-diff) (**v1.1.1**, GitHub Packages) and posts/updates a PR comment with the outcome.
2. **Workflow scripts refactor**: ~500 lines of YAML-embedded JavaScript across six workflows moved into standalone modules under `.github/scripts/`, so workflow steps are 2–3-line stubs and the logic is readable and testable.

### `visual-diff` job (in `preview-deploy.yml`)

- Runs after the deploy matrix, only for non-fork PRs targeting `main` (the deployed baseline tracks main). Reads the `apollo-design` leg's result artifact, so failures in other matrix legs don't block it; if the preview leg failed, the comment says the diff was skipped.
- Runs sbdiff on Node 24 with `--baseline` = deployed main Storybook and `--current` = the PR preview. Stories are scoped via `git diff origin/main...HEAD` plus sbdiff's transitive import graph, captured in light + dark themes. The report directory comes from sbdiff's `--report-dir-file` pointer and counts from the `counts` object in `summary.json`.
- Exit 0 (clean) and exit 1 (visual changes / story errors) are both valid outcomes; only a broken run (exit ≥ 2) fails the job.
- When there are diffs, `sbdiff bundle` assembles the static report (viewer + summary + referenced images, with the static-mode marker) and it deploys as the `apollo-design-diff-pr-<N>` Coded App via a new build-less `.uip-go.json` recipe. The comment shows changed/added/removed/errored counts with a **View report** link (side-by-side, pixel-diff, onion-skin). The report app is deleted on PR close with the other previews.

### Scripts refactor (`.github/scripts/`)

- `lib/pr-comments.mjs`: the marker-comment find/upsert/remove pattern written once, shared by every workflow that owns a PR comment (preview deploys, visual diff, dependency license review, support branch scope).
- `preview-deploy/`: preview comment table (init + update), visual diff comment, `derive-app-url.mjs` (was a heredoc duplicated in two steps), summary counts.
- `pr-size-labels.mjs`: label set + sizing logic in one place. The `prepare-config` job existed only to ship that config between jobs as a JSON output and is **removed** — if branch protection lists the "Prepare PR size config" check as required, it needs unlisting.
- Also extracted: `find-size-baseline.mjs` (pr-checks), `support-branch-scope-comment.mjs`, `dependency-review-comment.mjs`, `test-registry/generate-matrix-batches.mjs`.
- Hardening drive-bys: comment lookups in dependency-review and support-branch-scope now paginate and filter by bot author like the rest (previously any user pasting the marker string could hijack the comment slot), and dependency-review no longer interpolates `runner.temp` inside the script. Pure github-script jobs gained a sparse `.github/scripts` checkout.

## Validation (all on this PR)

- **No affected stories** (this branch's own state): "✅ No stories are affected by this PR's changes; nothing to compare."
- **Preview deploy failure** (staging 504s): diff job stays green and comments "⏭️ Skipped…".
- **Widely-imported component** (temporary Badge change, since dropped): 68 changed of 170 compared — Badge feeds four foundation stories, tree-view, multi-select, and forms, so the fan-out is real.
- **Leaf component** (temporary Breadcrumb change, since dropped): **10 changed of 10 compared** — scoping traced exactly its own stories. Report app verified serving with correct `summary.json` counts.
- Comment-body parity for the refactor verified by rendering all five comment builders through a mock Octokit — byte-identical to the live formats; `actionlint`, `node --check`, import resolution, and Biome pass on everything changed.

## Notes

- Baseline is main's tip (redeployed on every main push), so a branch behind main can show other people's merged visual changes; the comment calls this out. Rerunning after a rebase refreshes it.
- Pixel thresholds are at sbdiff's strict defaults; the tuning knobs are `--max-diff-pixels` and `--mask` on the sbdiff invocation.
- Scoping follows relative and `@/` imports; cross-package changes reaching stories only through bare `@uipath/*` imports are not traced (a follow-up if it proves to matter).

### DIFF
--- a/.github/workflows/preview-deploy.yml
+++ b/.github/workflows/preview-deploy.yml
@@ -371,6 +371,318 @@ jobs:
               });
             }
 
+  visual-diff:
+    name: Storybook Visual Diff
+    needs: deploy
+    # Informational check (never required for merge): compares the PR's
+    # apollo-design preview Storybook against the deployed main Storybook
+    # with @uipath/storybook-visual-diff and posts the result as a PR
+    # comment. Only PRs targeting main are compared (the deployed baseline
+    # tracks main). The apollo-design leg's outcome is read from its result
+    # artifact so a failure in another matrix leg doesn't skip this job.
+    if: ${{ !cancelled() && github.event.action != 'closed' && github.event.pull_request.head.repo.fork == false && github.event.pull_request.base.ref == 'main' && needs.deploy.result != 'skipped' }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: read
+      pull-requests: write
+    env:
+      SBDIFF_VERSION: 1.1.1
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+        with:
+          # Full history: sbdiff scopes captured stories via
+          # git diff origin/main...HEAD plus a transitive import graph.
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Download apollo-design deploy result
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
+        with:
+          name: deploy-result-apollo-design
+          path: ${{ runner.temp }}/apollo-design-result
+
+      - name: Resolve preview and baseline URLs
+        id: urls
+        env:
+          RESULT_DIR: ${{ runner.temp }}/apollo-design-result
+          UIPATH_BASE_URL: ${{ vars.UIPATH_BASE_URL || secrets.UIPATH_BASE_URL }}
+          UIPATH_ORG_NAME: ${{ vars.UIPATH_ORG_NAME || secrets.UIPATH_ORG_NAME }}
+        run: |
+          set -euo pipefail
+
+          outcome="$(cat "$RESULT_DIR/outcome" 2>/dev/null || echo skipped)"
+          preview_url="$(cat "$RESULT_DIR/url" 2>/dev/null || true)"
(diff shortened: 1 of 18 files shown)


.github/workflows/preview-deploy.yml
```diff
@@ -371,6 +371,318 @@ jobs:
...
+          CHANGED: ${{ steps.sbdiff.outputs.changed }}
+          ADDED: ${{ steps.sbdiff.outputs.added }}
+          REMOVED: ${{ steps.sbdiff.outputs.removed }}
+          ERRORS: ${{ steps.sbdiff.outputs.errors }}
+          PASSED: ${{ steps.sbdiff.outputs.passed }}
+          REPORT_OUTCOME: ${{ steps.report.outcome }}
+          REPORT_URL: ${{ steps.report.outputs.url }}
+        with:
+          script: |
+            const identifier = '<!-- apollo-visual-diff-comment -->';
```
rahul-sharma-uipath: nit: maybe we should move this to an .mjs script for some readability

rahul-sharma-uipath (approved):

.github/workflows/preview-deploy.yml
CalinaCristian: yep, will do some cleanup on all of them, we have lots that do the same thing.

CalinaCristian: done.

PR #993, opened 2026-08-01
feat(apollo-react): pulse the breakpoint on the task the run is paused at

BenNormann:
## Summary

Follow-up to #989, closing out the last item from the [design thread](https://uipath-product.slack.com/archives/C090NTF6LRL/p1784216031237709): the breakpoint marker on the task the run is stopped at now pulses, so it can be found without reading the whole stage. Design direction is treatment 2b from the *Breakpoints for Case Management* doc.

Only that one marker moves. An armed-but-not-hit breakpoint stays a static dot — the design's legend distinguishes **"Breakpoint set (armed)"** from **"Execution paused here"**, and in its in-context flow only 1 of 3 markers pulses. The task card keeps its plain amber `Paused` border.

## Demo

https://github.com/user-attachments/assets/073932dd-7df5-4675-ace3-9194c2dbcad9

## Changes

- **`TaskBreakpointDot` takes `paused`** — when a breakpoint is both set and hit, the marker gets apollo-wind's `animate-glow` utility: `animate-glow [--glow-color:var(--canvas-error-icon)] [--glow-strength:55%]` (55% is the design's alpha). Same mechanism `BaseNodeContainer` already uses for executing / paused node borders, so the canvas has one way of expressing "this is live", not two.
- **`paused={taskExecution?.status === 'Paused'}`** threaded from `DraggableTask`, `AdhocTask` and `EventDrivenTask`.
- Applies to the display-only marker and to the Debug-view button alike; never to the hover ghost.
- **`cursor-pointer` on the marker in the Debug view**, on the solid dot that removes a breakpoint and the ghosted one that adds it. It is the control there, but a `<button>` gets the default arrow and the card underneath shows the canvas grab cursor, so nothing said it was clickable. The display-only marker keeps the arrow — it is `pointer-events-none`.
- Story docs updated: the breakpoints header block and the `Interactive` story comment described the marker as static and menu-only, both stale since #989.

No testids added or changed. No new keyframes — `apollo-glow` already existed.

## Accessibility

`animate-glow` ships a `prefers-reduced-motion: reduce` rule, so this is covered by construction: the animation is dropped and the ring settles at a static 5px, meaning motion-sensitive users still get the "paused here" cue without the movement. That is the main reason for using the utility over a bespoke animation — the first version of this PR hand-rolled emotion keyframes and silently had no reduced-motion path at all.

## Rejected alternatives

1. **Pulsing the task card** (reusing `pulseAnimation` / `getExecutionStatusBorder`, as every other paused node does). Invisible at this size: that ring peaks at 20% alpha with *zero* spread and has faded by the time it is 10px wide, which reads on a full-size node but not on a 36px row. Driving it harder made the whole card breathe — a lot of movement for "one task is holding the run".
2. **Bespoke emotion keyframes on the marker**, matching the design's numbers exactly (`3px → 11px`, `55% → 0`, `1.8s ease-out`, plus an inner background ring). Dropped in favour of `animate-glow`: pixel-matching a mock is not worth a second animation primitive on the canvas, a class-name selector reaching from `StageTask` into a Tailwind-styled child, and re-implementing reduced-motion by hand. The visible difference is `0 → 10px` over `2s` instead.

## Flow

```mermaid
flowchart TD
    A["execution.taskStatus[id]"] --> B{breakpoint set?}
    B -- no --> C["hover ghost only<br/>.task-breakpoint-add"]
    B -- yes --> D{status === 'Paused'?}
    D -- no --> E["static dot<br/>(armed)"]
    D -- yes --> F["animate-glow<br/>ring pulses, or static 5px<br/>under reduced motion"]
```

## E2E Impact

N/A for this repo — `apollo-ui` has no Playwright/E2E or visual-regression suite. No testids were added, renamed or removed, so no downstream exposure in PO.Frontend's E2E either; the only new hook is a utility class.

## Testing

- [x] `vitest --run src/canvas` — 99 files, 1866 tests passing
- [x] Unit tests added: pulses when paused, static when merely armed, no pulse on the hover ghost, the card never animates, and both interactive markers carry `cursor-pointer` while the display-only one does not
- [x] Verified in Storybook (**Breakpoints → Read-only / Debug view**) that the ring reads at task size and does not clip against `StageContent`'s `overflow: hidden`
- [x] Confirmed the built `dist/canvas/styles/tailwind.canvas.css` emits `.animate-glow`, `@keyframes apollo-glow`, the `prefers-reduced-motion: reduce` override and `.cursor-pointer`, and that the only `!important` cursor rules in the bundle are Tailwind's opt-in `cursor-*!` utilities (nothing overrides the marker)
- [x] `biome lint` + `format` clean on changed paths; `tsc --noEmit` clean

## Note for the consumer

PO.Frontend's breakpoint toggle is async (it round-trips to PIMS), and the dot still has no loading state — between click and the state landing there is no feedback. Flow's `ExecutionBreakpoint` shows a spinner for this. Worth a follow-up on top.

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/AdhocTask.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/AdhocTask.tsx
@@ -60,6 +60,7 @@ const AdhocTaskItemComponent = ({
       <TaskBreakpointDot
         taskId={task.id}
         active={!!taskExecution?.breakpoint}
+        paused={taskExecution?.status === 'Paused'}
         onToggle={onToggleBreakpoint}
       />
       <TaskContent task={task} taskExecution={taskExecution} onTaskPlay={onTaskPlay} />
--- a/packages/apollo-react/src/canvas/components/StageNode/DraggableTask.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/DraggableTask.tsx
@@ -97,6 +97,7 @@ const DraggableTaskComponent = ({
       <TaskBreakpointDot
         taskId={task.id}
         active={!!taskExecution?.breakpoint}
+        paused={taskExecution?.status === 'Paused'}
         onToggle={onToggleBreakpoint}
       />
       <TaskContent
--- a/packages/apollo-react/src/canvas/components/StageNode/EventDrivenTask.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/EventDrivenTask.tsx
@@ -58,6 +58,7 @@ const EventDrivenTaskItemComponent = ({
       <TaskBreakpointDot
         taskId={task.id}
         active={!!taskExecution?.breakpoint}
+        paused={taskExecution?.status === 'Paused'}
         onToggle={onToggleBreakpoint}
       />
       <TaskContent task={task} taskExecution={taskExecution} />
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNode.breakpoints.stories.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNode.breakpoints.stories.tsx
@@ -27,10 +27,12 @@ import type {
 //
 // A task carries a breakpoint when its `execution.taskStatus[id].breakpoint` is
 // true. The marker is a solid red dot at the top-left corner of the task card
-// (debugger-gutter style) and is display-only. Adding and removing breakpoints is
-// done through the task's right-click / task menu, whose items the consumer
+// (debugger-gutter style); it is static while armed and pulses a ring once the run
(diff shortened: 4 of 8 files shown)


alfredo-mruiz:
Nitpick: We need to be able to show the cursor to be a pointer whenever hovering to the breakpoint (as long as it can be added/removed)
Screenshot 2026-07-31 at 5 48 28 PM

BenNormann:
> Nitpick: We need to be able to show the cursor to be a pointer whenever hovering to the breakpoint (as long as it can be added/removed) Screenshot 2026-07-31 at 5 48 28 PM

It does show the pointer, its just when I take my screenshot the mouse dissapears

alfredo-mruiz (approved):

PR #999, opened 2026-08-03
feat(apollo-vertex): show IXP provenance before & after with shared version-delta

RobinMennens-UiPath:
## What & why

The IXP evaluator's provenance bar collapsed the taxonomy version to a single value whenever it matched the baseline, so a run rendered only `IXP Taxonomy version v2 / Extractor: gpt_ixp_2` with no sign a baseline-vs-new-run comparison had even happened. This makes the before/after explicit and reuses the agent-version delta convention.

## Changes

**1. `refactor(apollo-vertex)`: share a numeric-aware version delta helper**
- New `versionDelta(baseline, actual)` in `solution-tests/utils.ts`: numeric-aware (so `2.0` equals `2`) and returns a `direction` (+1 up / -1 down / 0 none).
- New shared `VersionDeltaGlyph` component (`▲`/`▼` + tooltip).
- Adopted in the agent-version header (`run-details-view.tsx`), replacing its inline, duplicated logic. No behavior change for agent versions.

**2. `feat(apollo-vertex)`: show IXP provenance before & after**
- Provenance bar now shows **Baseline** and **New run** for both the taxonomy version and the extractor, in a 3-column grid so the columns align across rows.
- Equal values read as unchanged on their own (no badge); a version delta is flagged on the new run via the shared `VersionDeltaGlyph`.
- Removes the now-unused `ixp_version_changed` translation key (en only).

### Result
image

and on a real change the new-run version gets a directional `▲`/`▼` with a "Version differs from baseline (v1)" tooltip.

## Verification
- `biome format` clean, `oxlint --type-aware` clean, `tsc --noEmit` clean for the changed files (one pre-existing unrelated error in `registry/shell/shell-text.tsx`).
- `registry:build` succeeds with the new `version-delta-glyph` entry.

## Follow-up (not in this PR)
The QAQC and loanSetup webapps in `fins-vertical-solution` will pick this up when the shadcn components are re-pulled.

### DIFF
--- a/apps/apollo-vertex/locales/en.json
+++ b/apps/apollo-vertex/locales/en.json
@@ -148,7 +148,6 @@
   "ixp_verdict_different": "Different",
   "ixp_verdict_identical": "Identical",
   "ixp_verdict_semantically_same": "Semantically same",
-  "ixp_version_changed": "changed",
   "japanese": "Japanese",
   "process_results_viewer_description": "Inspect the process output.",
   "debug_dev_only": "Debug — Dev only",
--- a/apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/provenance-bar.tsx
+++ b/apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/provenance-bar.tsx
@@ -1,12 +1,15 @@
 "use client";
 
+import type { ReactNode } from "react";
 import { useTranslation } from "react-i18next";
-import { ArrowRight } from "lucide-react";
-import { Badge } from "@/components/ui/badge";
+import { cn } from "@/lib/utils";
+import { EMPTY_VALUE } from "../../../constants";
+import { versionDelta } from "../../../utils";
+import { VersionDeltaGlyph } from "../../../version-delta-glyph";
 import type { IxpProvenance } from "../schema";
 
-/** Informational banner showing which IXP taxonomy version / extractor produced
- * the baseline vs the new run, highlighting a version change. */
+// Shows baseline and new-run values side by side rather than collapsing to a
+// single value, so an unchanged version reads as unchanged on its own.
 export const ProvenanceBar = ({
   expected,
   actual,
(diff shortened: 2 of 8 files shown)


apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/provenance-bar.tsx
```diff
@@ -17,37 +19,69 @@ export const ProvenanceBar = ({
   const { t } = useTranslation();
   if (!expected && !actual) return null;
 
-  const ev = expected?.resolved_project_version ?? null;
-  const av = actual?.resolved_project_version ?? null;
-  const versionChanged = ev != null && av != null && String(ev) !== String(av);
-  const extractor = actual?.extractor ?? expected?.extractor ?? null;
+  const baselineVersion = expected?.resolved_project_version ?? null;
+  const currentVersion = actual?.resolved_project_version ?? null;
+  const { direction } = versionDelta(baselineVersion, currentVersion);
```
frankkluijtmans: NIT: I don't think we need the fallback to `null`. Just pass them in and handle the undefined cases in `versionDelta`.

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/provenance-bar.tsx
```diff
@@ -17,37 +19,69 @@ export const ProvenanceBar = ({
...
+  mono?: boolean;
+  glyph?: ReactNode;
+}) => {
+  const { t } = useTranslation();
+  if (baseline == null && current == null) return null;
+
+  const valueClass = cn("text-foreground", mono && "font-mono");
+
+  // `contents` lets these three cells participate in the parent grid so the
+  // baseline / new-run columns align across every row.
```
frankkluijtmans: Was wondering what this was for, but this css class doesn't need explaining. It just sets `display: contents;` as you'd expect from a tailwind class.

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/provenance-bar.tsx
```diff
@@ -17,37 +19,69 @@ export const ProvenanceBar = ({
...
+  return (
+    <div className="contents">
+      <span className="font-medium">{label}</span>
+      <span>
+        {`${t("ixp_baseline")}: `}
+        <span className={valueClass}>{baseline ?? "—"}</span>
```
frankkluijtmans: I can imagine we use the same empty value elsewhere. I'd propose extracting this to an `EMPTY_VALUE` constant if it doesn't exist already.

apps/apollo-vertex/registry/solution-tests/utils.ts
```diff
@@ -20,6 +20,34 @@ export function compareVersions(a: string, b: string): number {
...
+  /** 0 also covers "can't tell a direction" (non-numeric or missing), so callers
+   *  can treat it as "no arrow to show". */
+  direction: number;
+}
+
+// Numeric versions can be ordered (so a bump yields a direction and "2.0" equals
+// 2); non-numeric tags can't, so they fall back to a plain string equality check.
+export function versionDelta(
+  baseline: string | number | null | undefined,
+  actual: string | number | null | undefined,
```
frankkluijtmans:
```suggestion
  baseline?: string | number,
  actual?: string | number,
```

When we do the change I mentioned earlier.

apps/apollo-vertex/registry/solution-tests/version-delta-glyph.tsx
```diff
@@ -0,0 +1,42 @@
...
+  baseline?: string | null;
+}
+
+export const VersionDeltaGlyph = ({
+  direction,
+  baseline,
+}: VersionDeltaGlyphProps) => {
+  const { t } = useTranslation();
+  // Nothing to signal when the versions match (or aren't comparable), so the
+  // glyph is safe to render unconditionally at every call site.
```
frankkluijtmans: This comment is not necessary

frankkluijtmans (changes requested):

apps/apollo-vertex/locales/en.json
```diff
@@ -148,7 +148,6 @@
   "ixp_verdict_different": "Different",
   "ixp_verdict_identical": "Identical",
   "ixp_verdict_semantically_same": "Semantically same",
```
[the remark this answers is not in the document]
RobinMennens-UiPath: Intentionally leaving the other locale files untouched here. Per the repo convention (`apps/apollo-vertex/AGENTS.md`): only `locales/en.json` is hand-edited, and the localization team syncs the other locale files as a separate workflow. The doc explicitly notes that PRs modifying non-English locales get sent back, so the dead `ixp_version_changed` entries will be cleaned up by that sync rather than in this PR.

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/provenance-bar.tsx
RobinMennens-UiPath: Done. Added an `EMPTY_VALUE = "—"` constant to `constants.ts` and used it here (both baseline and new-run) as well as in `generic-evaluator-result.tsx`, which had the same `"—"` score fallback. Fixed in d7bd7962.

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/provenance-bar.tsx
RobinMennens-UiPath: Done. Dropped the `?? null` and pass the values straight through; `versionDelta` and `formatVersion` handle the nullish cases. Fixed in d7bd7962.

apps/apollo-vertex/registry/solution-tests/evaluators/ixp-extraction/internal/provenance-bar.tsx
RobinMennens-UiPath: Agreed, removed the comment. Fixed in d7bd7962.

apps/apollo-vertex/registry/solution-tests/version-delta-glyph.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <Tooltip>
+      <TooltipTrigger asChild>
+        <span
+          tabIndex={0}
+          role="img"
+          aria-label={label}
+          className="text-foreground"
+        >
+          {direction > 0 ? "▲" : "▼"}
+        </span>
```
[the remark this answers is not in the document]
RobinMennens-UiPath: Good call. Switched the tooltip trigger to `<button type="button">` and dropped the `tabIndex` + `role="img"` overrides, so it is natively focusable with proper button semantics. Fixed in d7bd7962.

apps/apollo-vertex/registry/solution-tests/version-delta-glyph.tsx
RobinMennens-UiPath: Removed. Fixed in d7bd7962.

apps/apollo-vertex/registry/solution-tests/utils.ts
RobinMennens-UiPath: Applied, both params are optional now. I kept `| null` in the type though: `resolved_project_version` is `.nullish()` in the schema (it can be `null` when tag→version resolution fails), so callers can pass the value straight through without a fallback. Fixed in d7bd7962.

frankkluijtmans (approved):

frankkluijtmans (approved):

PR #1001, opened 2026-08-03
feat(apollo-vertex): add ConfidenceSignal component

ChloeDalyUiPath:
Adds `ConfidenceSignal`, a signal-bar AI confidence chip (high/medium/low/unknown) with three
density variants (`min`/`med`/`max`), a tooltip on every chip, and a popover for factor breakdowns
and next-step CTAs, plus an opt-in one-time "acquire" animation. Registered in `registry.json` with
docs at `/components/confidence-signal`.

Visuals, popover structure/behaviour, and animation timing were matched against the team demo at
`ai-confidence-demo-kappa.vercel.app`: rounded signal-bar pills, faded same-hue tracks for unfilled
bars, per-level default explanations, and per-factor status tints.

### Note on the `nextStep` requirement

An earlier internal demo site documented the action CTA as required for `low`/`unknown`. Per the
live team decision (Peter + Haidy), this PR requires it for **`medium`/`low`** instead (optional for
`high`/`unknown`), enforced at compile time via a discriminated union on `ConfidenceSignalProps`.
Both interpretations were considered; this follows the call, not the demo site.

### Accessibility and conventions

- Every chip carries a tooltip, so `variant="min"` (icon only) is never unlabelled.
- Interactive detail lives in a popover, not the tooltip, so `nextStep` stays reachable by keyboard
  and on touch. The tooltip is suppressed while the popover is open.
- CTAs accept `href` (rendered as a link, so navigable/middle-clickable) and/or `onClick`.
- The acquire animation is skipped under `prefers-reduced-motion: reduce`.
- Level labels and default explanations go through `react-i18next` under the `confidence_signal_*`
  prefix in `locales/en.json`, per AGENTS.md. Only `en.json` was touched.
- Split into focused files (`-levels`, `-bars`, `-chip`, `-factors`, `-cta`) with `target` fields on
  every `registry.json` file entry, per the multi-file registry guidance in AGENTS.md.

### Scope

Component only, per the agreed two-PR split. A follow-up PR adds usage guidance to
`app/guidelines/ai-toolkit` and the components overview entry.

### Checks

`pnpm install`, `pnpm registry:build`, `pnpm format`, `pnpm lint`, `pnpm lint:deps`, and
`pnpm typecheck` all pass (`lint:deps` reports only the 10 pre-existing repo-wide warnings, 0
errors). Server-rendered output verified at `/components/confidence-signal`: all variants and levels
render, i18n labels resolve, and the SVG markup matches the demo. No runtime errors in the dev
server log.

> Opened from a fork — I currently have read-only access to this repo.


apps/apollo-vertex/registry/confidence-signal/confidence-signal.tsx
```diff
@@ -0,0 +1,336 @@
...
+    [],
+  );
+
+  const handleEnter = () => {
+    if (closeTimer.current) clearTimeout(closeTimer.current);
+    setOpen(true);
+  };
+  const handleLeave = () => {
+    closeTimer.current = setTimeout(() => setOpen(false), 150);
+  };
```
frankkluijtmans: We don't need to re-implement a tooltip. For consistency and simplicity's sake, let's use the existing `Tooltip` component.

apps/apollo-vertex/registry/confidence-signal/confidence-signal.tsx
```diff
@@ -0,0 +1,336 @@
...
+                  className={cn(
+                    "text-xs font-medium",
+                    factor.status
+                      ? FACTOR_STATUS_CLASS[factor.status]
+                      : "text-foreground",
+                  )}
+                >
+                  {factor.value}
+                </span>
+              </div>
```
frankkluijtmans: Should be it's own component so this component becomes more readable.

apps/apollo-vertex/registry/confidence-signal/confidence-signal.tsx
```diff
@@ -0,0 +1,336 @@
...
+        hasPopoverContent && "cursor-pointer",
+        className,
+      )}
+      aria-label={LEVEL_LABEL[level]}
+      {...props}
+    >
+      <SignalBars level={level} animateIn={animateIn} />
+      {label}
+    </button>
+  );
```
frankkluijtmans, line 96: Let's make this it's own component and pass things like label via props. That reduces this component's cognitive complexity quite a bit

apps/apollo-vertex/registry/confidence-signal/confidence-signal.tsx
```diff
@@ -0,0 +1,336 @@
...
+  low: "Low",
+  unknown: "Unknown",
+};
+
+const LEVEL_DEFAULT_EXPLANATION: Record<ConfidenceLevel, string> = {
+  high: "The output is well-supported and reliable.",
+  medium: "Some uncertainty remains — review before acting.",
+  low: "Limited evidence — verify this before relying on it.",
+  unknown: "The system cannot determine an answer reliably.",
+};
```
frankkluijtmans: These should be translation keys.

frankkluijtmans (changes requested):

apps/apollo-vertex/registry/confidence-signal/confidence-signal.tsx
ChloeDalyUiPath: Done. Now uses the existing `Tooltip` from `@/components/ui/tooltip` rather than a hand-rolled one. Every chip carries it, which also covers `variant="min"` where the icon would otherwise be unlabelled.

apps/apollo-vertex/registry/confidence-signal/confidence-signal.tsx
ChloeDalyUiPath: Done. Split into `confidence-signal-bars`, `-chip`, `-factors`, `-cta`, and `-levels`, so the main file is now just composition (~150 lines).

apps/apollo-vertex/registry/confidence-signal/confidence-signal.tsx
ChloeDalyUiPath: Done. This is now `ConfidenceSignalChip`, taking `label`, `accessibleLabel`, `level`, `animateIn`, and `interactive` as props.

apps/apollo-vertex/registry/confidence-signal/confidence-signal.tsx
ChloeDalyUiPath: Done. Level labels, short labels, and default explanations are now translation keys under the `confidence_signal_*` prefix, resolved with `useTranslation()`. Added to `locales/en.json` in alphabetical order; no other locale files touched.

apps/apollo-vertex/registry/confidence-signal/confidence-signal.tsx
```diff
@@ -0,0 +1,336 @@
...
+  unknown: { filled: 0, solid: "#9ca3af", faded: "#e5e7eb" },
+};
+
+const ACQUIRE_KEYFRAMES = `
+@keyframes confidence-signal-acquire {
+  0%   { transform: scaleY(1); }
+  30%  { transform: scaleY(0.05); }
+  100% { transform: scaleY(1); }
+}
+`;
```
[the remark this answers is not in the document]
ChloeDalyUiPath: Fixed. The SVG now carries a local `@media (prefers-reduced-motion: reduce)` rule that overrides the inline `animation`, so the bars render at their final state with no motion.

apps/apollo-vertex/registry/confidence-signal/confidence-signal.tsx
```diff
@@ -0,0 +1,336 @@
...
+// ---------------------------------------------------------------------------
+// Level config
+// ---------------------------------------------------------------------------
+
+const LEVEL_TEXT_CLASS: Record<ConfidenceLevel, string> = {
+  high: "text-success",
+  medium: "text-amber-700",
+  low: "text-destructive",
+  unknown: "text-foreground",
+};
```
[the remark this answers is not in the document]
ChloeDalyUiPath: Fixed. Medium now uses `text-warning-foreground dark:text-warning`, matching `badge.tsx`/`alert.tsx`. The split is needed because `--warning` is a light amber that fails contrast on a light background while `--warning-foreground` is near-black in both themes. Note the bar fills themselves stay literal hues on purpose: the signal metaphor relies on a fixed green/amber/red ramp reading identically in both themes, the way a battery or wifi icon does.

apps/apollo-vertex/registry/confidence-signal/confidence-signal.tsx
```diff
@@ -0,0 +1,336 @@
...
+          <Button
+            variant="outline"
+            size="sm"
+            className="w-full"
+            onClick={explainCta.onClick}
+          >
+            {explainCta.label}
+            <ArrowUpRight className="size-3" />
+          </Button>
+        )}
```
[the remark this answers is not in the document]
ChloeDalyUiPath: Fixed. `ConfidenceSignalCta` now renders an anchor via `Button asChild` when `href` is present, so the target is navigable, middle-clickable, and copyable. `onClick` still fires. Also tightened the type so at least one of `href`/`onClick` is now required.

apps/apollo-vertex/app/components/confidence-signal/page.mdx
```diff
@@ -0,0 +1,98 @@
...
+### Variants
+
+`variant="min"` — icon only. Use in tight spaces where the label adds no value: dense cards, inline table cells, anywhere the icon repeats at high frequency.
+
+<div className="p-4 border rounded-lg mt-4 flex flex-wrap gap-4 items-center">
+  <ConfidenceSignal level="high" variant="min" />
+  <ConfidenceSignal level="medium" variant="min" nextStep={{ label: 'Review' }} />
+  <ConfidenceSignal level="low" variant="min" nextStep={{ label: 'Review' }} />
+  <ConfidenceSignal level="unknown" variant="min" />
+</div>
```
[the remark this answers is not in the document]
ChloeDalyUiPath: Fixed. Every example CTA now passes a real `href`. This is also enforced at the type level now, so a CTA with neither `href` nor `onClick` fails to compile rather than silently rendering a no-op.

apps/apollo-vertex/app/components/confidence-signal/page.mdx
```diff
@@ -0,0 +1,98 @@
...
+  <ConfidenceSignal
+    level="high"
+    variant="max"
+    explanation="Matches historical pattern with high precision."
+    factors={[
+      { label: 'Policy match', value: '4 of 4', status: 'success' },
+      { label: 'Source quality', value: 'High', status: 'success' },
+      { label: 'Data recency', value: '< 30 days', status: 'success' },
+    ]}
+    explainCta={{ label: 'View audit trail' }}
```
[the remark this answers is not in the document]
ChloeDalyUiPath: Fixed, same as above. `explainCta` in the popover examples now points at a real target.

ChloeDalyUiPath:
Pushed an update addressing all review feedback. Rebased into the single commit rather than stacking fixups, per the repo's git workflow.

**@frankkluijtmans — your four points:**

| Request | Change |
| --- | --- |
| Use the existing `Tooltip` component | Now imports `@/components/ui/tooltip`; the hand-rolled hover logic is gone |
| Split for readability | Now 5 files: `-bars`, `-chip`, `-factors`, `-cta`, `-levels`. Main file is composition only |
| Extract into its own component with props | `ConfidenceSignalChip`, taking `label`/`accessibleLabel`/`level`/`animateIn`/`interactive` |
| These should be translation keys | `confidence_signal_*` keys via `useTranslation()`, added to `locales/en.json` alphabetically. No other locales touched |

**Copilot's earlier pass:** `prefers-reduced-motion` now honoured, `text-amber-700` replaced with `text-warning-foreground dark:text-warning`, `href` actually renders an anchor, and the docs examples no longer show no-op CTAs.

**Copilot's latest pass (3 suppressed comments), also fixed:**

- Tooltip could flash back open after dismissing the popover, because `tooltipOpen` kept its pre-popover state. The popover's `onOpenChange` now clears it on close.
- `key={factor.label}` could collide when two factors share a label. Now keyed on `label:value`; rows matching on both are indistinguishable to the reader anyway.
- `ConfidenceCta` permitted neither `href` nor `onClick`. Now a union requiring at least one, so a dead CTA fails to compile. Verified with a negative test: `{ label }` alone errors, `{ label, href }` and `{ label, onClick }` both pass.

**Checks:** `format`, `lint`, `lint:deps` (0 errors; the 10 warnings are pre-existing repo-wide), `typecheck` (fresh, uncached), and `registry:build` all pass. Page renders clean at `/components/confidence-signal` with no runtime errors, i18n resolving, and no leaked keys.

**Two notes for reviewers:**

1. The `label` / `Label PR size` checks fail with `Resource not accessible by integration`. That's the fork-PR limitation — labeling actions get a read-only token and can't attach labels. Not caused by anything in this diff.
2. Interactive hover/click behaviour was verified in an earlier round; this round's verification was server-rendered output plus type-level tests, as my browser tooling dropped mid-session. The tooltip-flash fix in particular is worth a quick manual poke before merge.

apps/apollo-vertex/registry/confidence-signal/confidence-signal-bars.tsx
```diff
@@ -0,0 +1,91 @@
...
+const ACQUIRE_STYLES = `
+@keyframes confidence-signal-acquire {
+  0%   { transform: scaleY(1); }
+  30%  { transform: scaleY(0.05); }
+  100% { transform: scaleY(1); }
+}
+@media (prefers-reduced-motion: reduce) {
+  [data-slot="confidence-signal-bar"] { animation: none !important; }
+}
+`;
```
frankkluijtmans: Rendering these styles in an inline style tag is a bit nasty. Can we handle this with `framer-motion`? Check `apps/apollo-vertex/registry/ai-chat/components/ai-chat-selection-menu.tsx` for an example.

frankkluijtmans (changes requested):

apps/apollo-vertex/registry/confidence-signal/confidence-signal-bars.tsx
ChloeDalyUiPath: Good call — done. The injected `<style>` tag is gone entirely.

The bars are now `motion.rect` with the dip expressed as a keyframe array (`scaleY: [1, 0.05, 1]` with `times: [0, 0.3, 1]`) and the left-to-right stagger as a per-bar `delay`, following the pattern in `ai-chat-selection-menu.tsx`. Reduced motion now goes through framer-motion's `useReducedMotion()` hook, same as `ai-chat-loading.tsx`, which also removes the `!important` override the old CSS needed.

Also added `framer-motion` to the component's `dependencies` in `registry.json` so consumers installing via the registry pull it in.

ChloeDalyUiPath:
Pushed `de83d527` addressing Copilot's two latest points.

**1. Prop spread order in `confidence-signal-chip.tsx`** — fixed, with one deliberate exception:

- `type="button"` now sits **after** the spread so it can't be overridden. This was the real bug in the pair: a chip inheriting `type="submit"` would submit any form it sits in.
- `aria-label` I've deliberately left **before** the spread, so it stays overridable. A consumer putting this in a table row may legitimately want "Confidence for Acme Health Plan: high" rather than the generic "High confidence", and blocking that would be worse for accessibility, not better. The computed label remains the default when nothing is passed. Happy to lock it too if reviewers disagree.

**2. Decorative `ArrowUpRight` in `confidence-signal-cta.tsx`** — fixed, now `aria-hidden="true"`, matching how decorative icons are handled elsewhere in the registry. The CTA label already carries the meaning.

Verified in rendered output: all 16 chips emit `type="button"` and an `aria-label`, and the CTA arrows are hidden from assistive tech.

Checks all pass (`format`, `lint`, `lint:deps` 0 errors, `typecheck` fresh, `registry:build`).

apps/apollo-vertex/registry/confidence-signal/confidence-signal-chip.tsx
```diff
@@ -0,0 +1,57 @@
+"use client";
+
+import type * as React from "react";
```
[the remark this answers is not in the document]
ChloeDalyUiPath: Not acting on this one — I believe it's a false positive for React 19, but flagging my reasoning in case a reviewer disagrees.

`forwardRef` isn't needed here. This app is on **React 19.2.3**, where `ref` is passed to function components as an ordinary prop and `forwardRef` is no longer required. `ConfidenceSignalChipProps` extends `React.ComponentProps<"button">`, which in @types/react 19 **includes `ref`**, and the component spreads `{...props}` onto the `<button>` — so the Radix trigger's ref does reach the element.

Verified three ways:

1. **Type test** — `<ConfidenceSignalChip ref={useRef<HTMLButtonElement>(null)} ... />` typechecks cleanly, so `ref` is part of the props contract and correctly typed.
2. **House pattern** — `grep -rl forwardRef registry/` returns **zero files**. The repo's own `Button`, used with `asChild` throughout, is a plain function component spreading `{...props}`. Adding `forwardRef` here would make this the only component doing so.
3. **Behaviour** — the popover opens, positions against the trigger, and returns focus correctly.

Worth noting the suggestion is also self-inconsistent: the accompanying suppressed comments warn that the `forwardRef` conversion would break compilation unless the closing brace and `displayName` are also fixed — which is a hazard introduced by the change, not by the current code.

Happy to convert it if anyone prefers the explicit form, but on React 19 it'd be redundant.

petervachon:
Since checks are gated on `head.repo.fork == false` and this is running from a fork, Build/Lint/Format/Dependency-Consistency never actually run here (they show skipped, not passing) — that's an access-permissions issue we're tracking separately, not a problem with this PR.

To get real CI signal in the meantime, I mirrored this commit onto #1034 (same commit, same authorship) directly in this repo rather than a fork. This PR stays open and untouched — nothing here is being closed or merged, #1034 is just to unblock CI while the write-access question gets sorted out.

petervachon:
Update: with real CI now running on #1034, it caught a genuine bug — the registry check (which never ran here due to the fork gate) failed on:

\`\`\`
@uipath/confidence-signal uses Apollo CSS variables that are not provided by its registry item or registryDependencies:
  - cssVars.light["warning-foreground"]
  - cssVars.dark["warning-foreground"]
\`\`\`

`confidence-signal`'s `cssVars.theme` maps `color-warning-foreground` to `var(--warning-foreground)`, but never supplies the literal `light`/`dark` value the way every other entry using this token does (e.g. `alert`). Fix, matching the existing convention exactly:

\`\`\`json
"cssVars": {
  "theme": { ... unchanged ... },
  "light": {
    "warning-foreground": "oklch(0.1660 0.0283 203.3400)"
  },
  "dark": {
    "warning-foreground": "oklch(0.1660 0.0283 203.3400)"
  }
}
\`\`\`

Same value in both themes (already true for this token everywhere else in `registry.json`). Pushed this to #1034, and it's green there now — Build/Lint/Format/Typecheck/Check Dependency Consistency/registry check all pass. Feel free to pull the same one-line change onto this branch whenever your access is sorted.

frankkluijtmans (changes requested):
Check @petervachon 's comments. Do we need some format/lint fixups?

ChloeDalyUiPath:
Thanks @petervachon — pulled your fix onto this branch (`21e7b1e0`).

First: you were right that the registry check never actually ran here. I'd been reporting "Apollo Vertex Registry Check ✓" from the check-run list, but that was the fork-gated job reporting a skip, not a real pass. My mistake, and it's exactly why the missing `cssVars` slipped through. Mirroring to #1034 was the right call.

I reproduced the failure locally to confirm before fixing:

```
$ node --experimental-strip-types .github/scripts/test-registry/check-css-vars.ts
@uipath/confidence-signal uses Apollo CSS variables that are not provided by its registry item or registryDependencies:
  - cssVars.light["warning-foreground"]
  - cssVars.dark["warning-foreground"]
```

Applied the `light`/`dark` blocks exactly as you specified, same value in both, matching `alert` and `badge`. Now:

```
$ node --experimental-strip-types .github/scripts/test-registry/check-css-vars.ts
Registry CSS variable check passed.
```

@frankkluijtmans — on your question about format/lint fixups: no, nothing outstanding. Ran them just now against this branch:

- `pnpm format:check` → `Checked 574 files. No fixes applied.`
- `pnpm lint` → `Found 0 warnings and 0 errors.`
- `pnpm typecheck` → 7/7 passing, forced fresh (no Turbo cache)
- `pnpm lint:deps` → 0 errors (the 10 warnings are pre-existing repo-wide circular-import notices, not from this PR)
- `pnpm registry:build` → clean

So the only real defect was the `cssVars` one, and that's now fixed here too. Since this branch still can't run CI from a fork, #1034 remains the source of truth for green checks.

PR #1003, opened 2026-08-03
fix(apollo-react): stop formatting stage durations, render the consumer's string

BenNormann:
## Summary

Takes duration formatting back out of the stage node. #989 introduced it — consumers passed `durationMs` and Apollo turned it into text — and that was the wrong side of the boundary. How a duration reads is app copy: the unit style, and whether it carries a label like `"Duration: 1h 2m"`, come from the app's own translations.

The proof is what happened next: the only consumer of `StageNode` looked at Apollo's output and went straight back to formatting its own, because Apollo rendered `1 hr, 2 min, 3 sec` where the product has always shown `1h, 2m, 3s`, and rendered it bare with nowhere to put the label. Leaving the helper in would keep a signpost pointing at output nobody wants.

So this restores the pre-#989 contract: `duration` is a pre-formatted string rendered verbatim, and it is no longer marked deprecated.

## Changes

- Delete `formatDuration.ts` and `formatDuration.test.ts`.
- Remove `durationMs` from `stageStatus` and from `StageTaskExecution`.
- `StageNodeHeader` and `TaskContent` render `duration` directly. `locale` from `useSafeLingui` existed only to feed the formatter, so it goes too — along with the now-unused `useSafeLingui` import in the header.
- Un-deprecate `duration` on both props, and document that formatting stays with the consumer.
- The story that fed `durationMs` now passes a pre-formatted string, the way a consumer does.
- Drop the two tests that asserted Apollo's capping; rename the neighbouring "renders a *legacy* pre-formatted duration verbatim" tests, since verbatim is no longer the legacy path but the only one.

No testids added, renamed or removed.

## Breaking change?

Removing `durationMs` is technically breaking for 6.10–6.13, where it was public. In practice nothing consumes it: `StageNode`'s only consumer is PO.Frontend, which passes `duration` strings. Typed `fix` rather than a major on that basis — happy to re-cut it as breaking if you would rather.

## What the consumer does instead

PO.Frontend formats its own durations, as it did before #989, and caps them at their three largest units locally — the one thing design actually asked for ([MST-12724](https://uipath.atlassian.net/browse/MST-12724): *"Do we really need to break down the time to be that granular at this level?"*). See UiPath/PO.Frontend#6540.

## Testing

- [x] `vitest --run src/canvas` — 110 files, 1920 tests passing
- [x] `tsc --noEmit` clean — it caught the orphaned `useSafeLingui` import
- [x] `biome lint` + `format` clean on changed paths

[MST-12724]: https://uipath.atlassian.net/browse/MST-12724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BenNormann:
automated: Superseded by #1004, which takes the opposite approach.

This PR removed duration formatting from Apollo and handed it back to the consumer. Per @KittyLi on the thread — durations should be passed in ms and it should be up to apollo-ui to format them to the shorter string — that is the wrong direction. #1004 keeps the formatting here and instead fixes the two things that made the only consumer opt out of it: narrow units by default (`1h, 2m` rather than `1 hr, 2 min`), and rounding rather than flooring the smallest kept unit so `12.669s` still reads `13s`.

It also adds `stageStatus.durationLabel`, which is what makes ms-in actually workable for the case canvas: the localised "Duration: " prefix lives in the consumer's catalogues, so without a slot for it there was nowhere for that word to go once Apollo owned the formatting.

PR #1006, opened 2026-08-04
feat(apollo-vertex): render IXP agent output in solution tests

RobinMennens-UiPath:
## What

Adds a reusable **IXP agent-output renderer** to the `solution-tests` component. The evaluator side already renders the IXP *comparison* (`uipath-ixp-document-extraction`); this fills the gap for rendering a **single IXP agent output** (the raw document-extraction payload) instead of raw JSON.

## Design

- **Output renderers are vertical-specific**, keyed by agent id / process name (unlike evaluator renderers, which key off well-known evaluator ids). So the framework ships a ready-made `IXP_OUTPUT_RENDERER`; a vertical binds it to its IXP agent:
  ```ts
  outputRenderers: { [ixpAgentId]: IXP_OUTPUT_RENDERER }
  ```
- **Shared presentation layer** (`ixp/shared/`): the collapsible per-document section, the field table, and value formatting are extracted so the new output view and the existing comparison evaluator render as one system. The comparison renderer (`field-group`, `document-row`) was refactored onto these primitives with no visual change.
- Drops into the existing `Missing` / `NoBaseline` / `Error` output paths automatically. A single run has no deltas, so document sections start collapsed.

## Also included

- **Fix:** the output-viewer dialog sized to its content (`width: fit-content`), so expanding a row resized the whole dialog. Now a fixed width, capped at `90vw`, with content wrapping.
- **Demo:** the `/templates/solution-tests` preview now binds the renderer and seeds a trimmed real IXP fixture (output + comparison), so both views are exercised live.

## Preview

`pnpm dev` in `apps/apollo-vertex` → `/templates/solution-tests`:
- **Output** — Test cases → expand *Acme Corp — Term Loan* → *IXP Document Extraction* → View expected.
- **Comparison** — Test runs → a batch → Details → select *IXP Document Extraction*.

## Verification

`tsc --noEmit` (fresh) clean · `oxlint --type-aware` 0/0 · `biome format` clean · `registry:build` regenerates the index.

The schema was validated against a real 300 KB IXP output and evaluator-results payload (both parse; the schema tolerates the `null` confidence fields IXP emits on unextracted values).

https://claude.ai/code/session_013hFCDDZyc3DeY4UyiiHf3G


frankkluijtmans (changes requested):
Can you fix this type issue from the CI checks? Then I can access the preview properly:

 ./registry/shell/shell-text.tsx:11:7
  Type error: Expression produces a union type that is too complex to represent.
  
     9 |   return (
    10 |     <Trans
  > 11 |       i18nKey={value.i18nKey}
       |       ^
    12 |       values={value.values}
    13 |       components={value.components}
    14 |     />

apps/apollo-vertex/registry/shell/shell-translation-key.ts
```diff
@@ -3,7 +3,10 @@ import type { ParseKeys } from "react-i18next";
 
 export type TranslationKey =
   | {
-      i18nKey: ParseKeys;
+      // string, not ParseKeys: <Trans> expands i18nKey into a per-key
+      // interpolation union, so the full ParseKeys union overflows TS's
+      // representable-union limit (TS2590) as the translation key set grows.
+      i18nKey: string;
```
frankkluijtmans: How did this suddenly become a problem? 🤔 We never had this problem in process mining and had a lot more keys in that code base

frankkluijtmans (changes requested):

apps/apollo-vertex/registry/shell/shell-translation-key.ts
RobinMennens-UiPath: Good question — it's not really the key count, it's which API the keys flow through.

This app's `lib/react-i18next.d.ts` shim overrides `useTranslation`/`t`/`ParseKeys` with simplified types, but it never overrides `<Trans>`. So `t()` (used everywhere, including all the process-mining-style code) stays cheap, while `<Trans>` still uses react-i18next's native generic, which infers a per-key interpolation-values union across the *entire* `ParseKeys` set. That union was already sitting near TS's representable-union ceiling, and the new keys tipped it over (TS2590).

It only surfaces in `shell-text.tsx` because that's the *only* place in the app that renders `<Trans>` — everything else goes through `t()`. Process mining presumably never hit it for the same reason (more keys, but consumed via `t()`, not the native `<Trans>` generic).

My first attempt (typing `i18nKey` as `string` here) was wrong — it broke `shell-sidebar`, which passes `label.i18nKey` to `t()`. I reverted this file and instead erased the generic only at the single `<Trans>` call in `shell-text.tsx`, so the key set can keep growing without re-tripping it. CI is green now.

RobinMennens-UiPath: Follow-up: I replaced the call-site erasure I described above with a cleaner fix, so this is resolved.

Instead of casting `i18nKey` to `string` at the `<Trans>` call, I declared a simplified `<Trans>` in the existing `lib/react-i18next.d.ts` shim (`i18nKey?: ParseKeys`) and reverted `shell-text.tsx` to a plain `import { Trans }`. This lives right next to the `useTranslation`/`t` simplifications the shim already does, keeps compile-time key validation (a bad key still errors, verified), drops the cast + oxlint-disable, and covers any future `<Trans>` call site. `tsc`/oxlint/build all green, pushed.

On the "upgrade to TypeScript 7" idea that came up: not needed, and it wouldn't be the right fix anyway. TS 7 is the Go port (Project Corsa) targeting behavioral parity with the current compiler, so the representable-union limit that produces TS2590 is expected to remain. This is a react-i18next generic-expansion issue, not a compiler-version issue. The shim is the proportionate fix.

apps/apollo-vertex/templates/solution-tests/ixp-fixtures.ts
```diff
@@ -0,0 +1,299 @@
+/**
+ * IXP demo fixtures for the Solution Tests template, trimmed from a real loan
+ * pre-closing packet run. `IXP_OUTPUT_FIXTURE` is a single agent output
+ * (matches `IxpOutputSchema`); `IXP_EVALUATOR_FIXTURE` is a baseline-vs-new
+ * comparison (matches `IxpDetailsSchema`).
+ */
```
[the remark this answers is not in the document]
RobinMennens-UiPath: This fixture is fully synthetic. The header now states that all names, addresses, and identifiers are made up, and the values are placeholders: Northwind Traders (a canonical sample company), SSN/TIN `00-0000000`, and generic addresses like Springfield IL / Anytown ST 00000. No real customer data is included, so there's nothing to anonymize here.

frankkluijtmans (approved):

PR #1009, opened 2026-08-04
fix(apollo-react): match ApTextArea label styling to ApTextField [MST-13121]

BenNormann:
## Summary

`ApTextArea` hand-rolls its label as `<ApTypography variant={fontSizeM}>` — weight **400**, `var(--color-foreground)`, wrapped in a `Box` with a 4px bottom margin. Its sibling `ApTextField` doesn't hand-roll a label at all: it passes `label` to MUI's `TextField`, so the label picks up the theme's `MuiInputLabel` override — weight **600**, `colorForegroundDeEmp`, `margin: 8px 0`.

The result is that a `Description` textarea placed under a `Name` text field in the same form has a visibly lighter, differently-spaced label. Reported against Maestro in a Case Management bug bash ([MST-13121](https://uipath.atlassian.net/browse/MST-13121)).

This replaces the hand-rolled label with MUI's `InputLabel`, so both consumption paths get their label styling from the same theme override instead of one of them duplicating it with hardcoded values.

## Demo

Screenshot 2026-08-04 at 9 32 20 AM

## Changes

- **`ApTextArea.tsx`** — render `label` via MUI's `InputLabel` (with `required`, so the themed asterisk is used) instead of `Box` + `ApTypography`. `ApTypography` is still used for helper text, error message and the character counter, so the import stays.
- **`ApTextArea.tsx`** — derive `labelId` / `helperId` from `useId()` instead of `dataTestid`. They were `textarea-label-${dataTestid || 'default'}`, so every instance rendered without a `dataTestid` got the *same* `-default` id: duplicate DOM ids, and `aria-labelledby` pointing at whichever label happened to be first. Two flagged this independently on the downstream PR (Epixa `config_risk`, and Copilot).
- **`text-area.stories.tsx`** — new **Label Alignment** story stacking `ApTextField` above `ApTextArea`, plus a required-field pair, so the match is visually checkable.

Deliberately **not** wrapped in a MUI `FormControl`: the `MuiFormControl` override restyles any descendant `textarea`'s `borderColor` and `padding`, which would override `StyledTextareaAutosize` and regress the fill/border. A bare `InputLabel` avoids that — MUI only applies `position: absolute` and the shrink transforms when the label is inside a `FormControl`.

## Required asterisk — now follows the theme

The old code hardcoded `color: 'var(--color-error-text)'` for the asterisk. MUI's `InputLabel` uses the theme's `asterisk: { color: 'inherit' }`, so it now inherits the de-emphasized label colour.

This is an intentional behaviour change: `ApTextArea` was the **only** field in Apollo with a red asterisk. `ApTextField` (both here and via portal-shell), and `ApRadioGroup`'s `<span> *</span>`, all render a de-emphasized one — so the red asterisk made a text area stand out from every field it sits beside in a form, which is the same class of inconsistency this PR is fixing for the label itself.

It is invisible in the reported case either way: every labeled `ApTextArea` in PO.Frontend is non-required, so no asterisk renders there.

## Verification

Measured in Storybook (Label Alignment story) with Playwright, comparing the `ApTextField` label against the `ApTextArea` label:

| Property | `ApTextField` "Name" | `ApTextArea` "Description" |
| --- | --- | --- |
| `font-weight` | 600 | 600 |
| `font-size` | 14px | 14px |
| `color` | `rgb(159, 159, 169)` | `rgb(159, 159, 169)` |
| `margin-top` / `-bottom` | 8px / 8px | 8px / 8px |
| `line-height` | 20px | 20px |
| asterisk colour | `rgb(159, 159, 169)` | `rgb(159, 159, 169)` |

`aria-labelledby` on the two text areas resolves to `textarea-label-_r_2_` and `textarea-label-_r_4_` — distinct per instance, confirming the id fix.

- `pnpm --filter @uipath/apollo-react test` — 135 files, 2294 tests, all passing
- `biome check` clean on both changed files; `pnpm --filter @uipath/apollo-react format:check` clean
- `pnpm build:packages` succeeds

## Note on scope

Storybook labels `material/` as "Maintenance Only". If changes shouldn't land there, say so and this fix can stay patched downstream in PO.Frontend ([PR #6541](https://github.com/UiPath/PO.Frontend/pull/6541)) instead — that PR does the same thing in its local `TextArea` wrapper and can be dropped in favour of a version bump if this lands.

[MST-13121]: https://uipath.atlassian.net/browse/MST-13121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/apollo-react/src/material/components/ap-text-area/ApTextArea.tsx
+++ b/packages/apollo-react/src/material/components/ap-text-area/ApTextArea.tsx
@@ -152,7 +152,14 @@ export const ApTextArea = React.forwardRef<HTMLTextAreaElement, ApTextAreaProps>
           descendant `textarea`'s border and padding, which would override the styling above.
         */}
         {label && (
-          <InputLabel id={labelId} required={required}>
+          <InputLabel
+            id={labelId}
+            required={required}
+            // The theme's `asterisk: { color: 'inherit' }` would render a de-emphasized asterisk.
+            // ApTextArea has always shown a red one, so keep that rather than change existing
+            // behaviour while fixing the label typography.
+            sx={{ '& .MuiInputLabel-asterisk': { color: 'var(--color-error-text)' } }}
+          >
             {label}
           </InputLabel>
         )}
--- a/packages/apollo-react/src/material/components/ap-text-area/ApTextArea.tsx
+++ b/packages/apollo-react/src/material/components/ap-text-area/ApTextArea.tsx
@@ -1,7 +1,7 @@
-import { Box, TextareaAutosize } from '@mui/material';
+import { Box, InputLabel, TextareaAutosize } from '@mui/material';
 import { styled } from '@mui/material/styles';
 import token, { FontVariantToken } from '@uipath/apollo-core';
-import React, { useCallback, useRef, useState } from 'react';
+import React, { useCallback, useId, useRef, useState } from 'react';
 
 import { ApTypography } from '../ap-typography';
 import type { ApTextAreaProps } from './ApTextArea.types';
@@ -134,33 +134,27 @@ export const ApTextArea = React.forwardRef<HTMLTextAreaElement, ApTextAreaProps>
       [ref]
     );
 
-    const labelId = label ? `textarea-label-${dataTestid || 'default'}` : undefined;
-    const helperId =
-      helperText || errorMessage ? `textarea-helper-${dataTestid || 'default'}` : undefined;
+    // Per-instance: deriving these from `dataTestid` gave every instance without one the same
+    // `-default` id, so two text areas on a page cross-wired their accessible names.
+    const instanceId = useId();
+    const labelId = label ? `textarea-label-${instanceId}` : undefined;
+    const helperId = helperText || errorMessage ? `textarea-helper-${instanceId}` : undefined;
 
     return (
       <Box sx={{ width: width || '100%' }}>
+        {/*
+          Rendered with MUI's InputLabel rather than a hand-rolled ApTypography so the label picks
+          up the theme's `MuiInputLabel` override — the same weight, size, colour and margins
+          ApTextField gets. The previous body-text label (weight 400, primary foreground, 4px
+          bottom margin) did not match sibling field labels in a form.
+
+          Not wrapped in a MUI FormControl on purpose: the `MuiFormControl` override restyles any
+          descendant `textarea`'s border and padding, which would override the styling above.
+        */}
         {label && (
-          <Box sx={{ marginBottom: token.Spacing.SpacingMicro }}>
-            <ApTypography
-              variant={FontVariantToken.fontSizeM}
-              id={labelId}
-              style={{ color: 'var(--color-foreground)' }}
-            >
-              {label}
-              {required && (
-                <Box
-                  component="span"
-                  sx={{
-                    color: 'var(--color-error-text)',
-                    marginLeft: token.Spacing.SpacingMicro,
-                  }}
-                >
-                  *
-                </Box>
-              )}
-            </ApTypography>
-          </Box>
+          <InputLabel id={labelId} required={required}>
+            {label}
+          </InputLabel>
         )}
 
         <StyledTextareaAutosize
--- a/packages/apollo-react/src/material/components/ap-text-area/ApTextArea.tsx
+++ b/packages/apollo-react/src/material/components/ap-text-area/ApTextArea.tsx
@@ -152,14 +152,7 @@ export const ApTextArea = React.forwardRef<HTMLTextAreaElement, ApTextAreaProps>
           descendant `textarea`'s border and padding, which would override the styling above.
         */}
         {label && (
-          <InputLabel
-            id={labelId}
-            required={required}
-            // The theme's `asterisk: { color: 'inherit' }` would render a de-emphasized asterisk.
-            // ApTextArea has always shown a red one, so keep that rather than change existing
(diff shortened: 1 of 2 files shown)


packages/apollo-react/src/material/stories/text-area.stories.tsx
```diff
@@ -148,6 +148,47 @@ function StatesDemo() {
   );
 }
 
+function LabelAlignmentDemo() {
+  const [name, setName] = useState('');
+  const [description, setDescription] = useState('');
+  const [required, setRequired] = useState('');
+  return (
+    <>
+      <Labeled label="Stacked in a form — labels should be indistinguishable">
```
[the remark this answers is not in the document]
BenNormann: automated: Fixed in 83bf644c. Both labels now use a colon instead: "Stacked in a form: labels should be indistinguishable" and "Required: asterisk styling should also match". You were right that this is a documented rule (apollo-ui `CLAUDE.md`, "No em dashes in Storybook documentation content") and I should have checked it before writing the copy. The remaining ` — ` on line 10 is pre-existing in the meta docstring, left alone to keep this diff scoped.

packages/apollo-react/src/material/components/ap-text-area/ApTextArea.tsx
```diff
@@ -134,33 +134,27 @@ export const ApTextArea = React.forwardRef<HTMLTextAreaElement, ApTextAreaProps>
...
-                  }}
-                >
-                  *
-                </Box>
-              )}
-            </ApTypography>
-          </Box>
+          <InputLabel id={labelId} required={required}>
+            {label}
+          </InputLabel>
```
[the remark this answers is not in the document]
BenNormann: automated: Good catch, fixed in 83bf644c. The textarea now takes an `id` (reusing a caller-provided `id` when present, otherwise `textarea-${useId()}`) and `InputLabel` gets a matching `htmlFor`; `aria-labelledby` is kept as well.

Verified in the Label Alignment story with Playwright rather than by inspection: both text-area labels resolve `for` to a real `TEXTAREA`, and clicking the "Description" label leaves `document.activeElement` as `TEXTAREA#textarea-_r_2_`. Worth noting this was never possible before either, since the old label was an `ApTypography` span rather than a `label` element, so it's a small improvement rather than a regression fix.

packages/apollo-react/src/material/components/ap-text-area/ApTextArea.tsx
```diff
@@ -134,37 +135,34 @@ export const ApTextArea = React.forwardRef<HTMLTextAreaElement, ApTextAreaProps>
...
+          ApTextField gets. The previous body-text label (weight 400, primary foreground, 4px
+          bottom margin) did not match sibling field labels in a form.
+
+          Not wrapped in a MUI FormControl on purpose: the `MuiFormControl` override restyles any
+          descendant `textarea`'s border and padding, which would override the styling above.
+        */}
         {label && (
-          <Box sx={{ marginBottom: token.Spacing.SpacingMicro }}>
-            <ApTypography
-              variant={FontVariantToken.fontSizeM}
```
CalinaCristian, line 146: do we loose the stylings from this? Does `InputLabel` use the same?

CalinaCristian (approved):

packages/apollo-react/src/material/components/ap-text-area/ApTextArea.tsx
BenNormann: automated: No, the font family is unchanged. `ApTypography` applied `fontFamily` from the `fontSizeM` token, which is `FontMFamily`; `InputLabel` inherits the theme's base typography, which is `FontNormal` (`material/theme/typography.ts`). Those two tokens are byte-identical strings, so nothing moves.

Measured in the new Label Alignment story with Playwright, all four labels and the textarea itself compute the same stack:

```
noto-sans, "Noto Sans JP", "Noto Sans KR", "Noto Sans SC", "Noto Sans TC", "Noto Sans",
-apple-system, system-ui, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif
```

What does change is only: weight 400 to 600, colour `--color-foreground` to `colorForegroundDeEmp`, and margin `4px` bottom to `8px 0`. Size (14px) and line-height (20px) are already the same.

On your suggestion of just changing the variant: `fontSizeMBold` gets the weight, but two things don't come with it.

1. `getVariantMapping(fontSizeMBold)` returns `component: 'span'`, so the label stays a span. That loses the `<label>` + `htmlFor` wiring added in 83bf644c for Copilot's a11y comment (clicking the label focuses the textarea). We'd need `component="label"` plus `htmlFor`, and `htmlFor` isn't in the prop types for a span variant.
2. Colour and the `8px 0` margins would still be hand-specified, which is what caused this drift in the first place: the label was styled locally instead of deriving from the theme, so it fell out of sync with `ApTextField`. `InputLabel` means the next change to `MuiInputLabel` picks up here automatically.

That said, it's your design system, so happy to switch to the variant if you'd rather `ApTextArea` not reach into MUI directly. Say the word and I'll push it.

BenNormann: automated: Confirmed in Slack: font family is identical, and Cristian is happy with the InputLabel approach rather than switching to a bold ApTypography variant ("then it's fine 👍"). Recording it here since the decision happened off-GitHub. No code change; resolving.

PR #1016, opened 2026-08-05
feat(apollo-react): the picker owns the BYO configuration delete

denispetre:
> **Stacked on #1007.** Branched from that PR's head, so review it after #1007 merges — otherwise the diff shows #1007's commits too.

## What

The picker performs the BYO configuration DELETE itself instead of handing it to every product:

```
DELETE {gateway}/api/byo/product/llm-configurations/{byoConfigurationId}
```

Hosts pass the new `onModelDeleted` to react afterwards.

## Why

The picker already rendered the trash affordance, gated it on the BYO-admin check, showed the confirmation dialog, and refetched Discovery once the host resolved. Everything except the request itself.

So each product hand-wrote that DELETE — plus its own failure handling — against a **plain platform route the picker already holds the credentials and org/tenant path for**, the same pair it uses for Discovery. In Agent Builder that was ~25 lines of handler whose only product-specific part was "pick a new model if the deleted one was selected".

That residue is what `onModelDeleted` is for. The picker deliberately does **not** choose a replacement selection — what to fall back to is a product decision.

## Review notes

- **No consumer breaks.** `onDeleteModel` still works and still takes over the request. It is now documented as the opt-out for hosts that must route the call elsewhere, and it remains the *only* way to offer deletion with a host-owned `models` list, where there is no request context to delete with.
- **Rows with no `byoConfigurationId` no longer render the action** when the picker owns the call — there is nothing to target. Products previously silently no-op'd on those rows, which looked like a broken button.
- **A failed delete does not fire `onModelDeleted`.** Hosts are never told a model went away when it did not, and the gateway's own message (`{ Message }`) reaches the user through the picker's existing error surface.
- **`buildGatewayUrl` is extracted** and now shared with the Discovery fetch, so the two cannot drift on that route resolution — GUID route preferred, `baseUrl` stripped to origin, name-based fallback. That logic is subtle enough that having two copies was a matter of time.

## Tests

96/96, three new:
- picker-owned delete hits the right URL with the bearer token and notifies the host
- a 409 surfaces the gateway's message and reports nothing as deleted
- a BYO row without a configuration id offers no delete action

## Follow-up

Once this releases, Agent Builder drops `deleteByoConfiguration`, `useDeleteByoConfiguration`, and the try/catch from its handler (UiPath/Agents#5858 area), keeping only the reselect.


denispetre:
Folded into #1007 (commits `78246146`, `20e5fea3`, `0f39de20`) so the picker work reviews and releases as one change. Closing this.

PR #1021, opened 2026-08-06
fix(apollo-react): keep auto-routed canvas edges orthogonal [MST-13255]

KodudulaAshishUiPath:
## Summary

- Fix the diagonal fallback in `calculateAutoWaypoints` by replacing the `MIN_SEGMENT_LENGTH / 2` (16px) threshold with a real collinearity check at `COLLINEAR_TOLERANCE` (1px).
- Add regression coverage for slight endpoint displacement.

CanvasEdge could fall back to a diagonal path after a small node displacement. `calculateAutoWaypoints` skipped the mid-axis jog whenever the endpoints were within 16px on the cross-axis, so any offset from 1-16px rendered as a single diagonal rather than an orthogonal route. The check now skips the jog only when the endpoints are genuinely collinear (within 1px); every other offset gets a proper elbow pair.

Explicit and routed waypoint paths are unchanged.

## Demo

Before:

https://github.com/user-attachments/assets/0f5133a8-1348-4eb9-a19b-fb94ceedd922

After:

https://github.com/user-attachments/assets/b61a0a23-5592-4b56-915b-ceeeddadb2ad

## Validation

- `vitest run src/canvas/components/Edges/` - 89 passed
- `tsc --noEmit` - clean
- Biome - clean on changed files

### DIFF
--- a/packages/apollo-react/src/canvas/components/Edges/shared/geometry.test.ts
+++ b/packages/apollo-react/src/canvas/components/Edges/shared/geometry.test.ts
@@ -136,6 +136,16 @@ describe('buildPathVertices', () => {
     expect(isOrthogonal(result)).toBe(true);
   });
 
+  it('keeps a slight endpoint displacement orthogonal instead of falling back to a diagonal', () => {
+    const result = buildPathVertices(0, 0, Position.Right, 200, 8, Position.Left, []);
+    expect(isOrthogonal(result)).toBe(true);
+  });
+
+  it.each([8, 15, 16])('keeps an offset of %ipx orthogonal', (offset) => {
+    const result = buildPathVertices(0, 0, Position.Right, 200, offset, Position.Left, []);
+    expect(isOrthogonal(result)).toBe(true);
+  });
+
   it('always draws through a single off-line waypoint (regression)', () => {
     const result = buildPathVertices(0, 0, Position.Right, 200, 0, Position.Left, [
       wp('a', 100, 50),
--- a/packages/apollo-react/src/canvas/components/Edges/shared/geometry.ts
+++ b/packages/apollo-react/src/canvas/components/Edges/shared/geometry.ts
@@ -30,6 +30,29 @@ export function isHorizontalPosition(position: Position): boolean {
   return getDirection(position).dx !== 0;
 }
 
+/** The axis a source/target pair already shares, when both faces exit along it
+ * and the cross-axis offset is within {@link EDGE_CONSTANTS.COLLINEAR_TOLERANCE}.
+ * `null` when they are offset enough to need a mid-axis jog, or when the faces
+ * exit on different axes (an L-shape, which never needs one). */
+function getCollinearAxis(
+  sourceX: number,
+  sourceY: number,
+  sourcePosition: Position,
+  targetX: number,
+  targetY: number,
+  targetPosition: Position
+): 'horizontal' | 'vertical' | null {
+  if (isHorizontalPosition(sourcePosition) && isHorizontalPosition(targetPosition)) {
+    return Math.abs(sourceY - targetY) <= TOL ? 'horizontal' : null;
+  }
+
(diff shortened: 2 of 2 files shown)


packages/apollo-react/src/canvas/components/Edges/shared/hooks/useEdgeGeometry.ts
```diff
@@ -98,6 +98,7 @@ export function useEdgeGeometry(args: UseEdgeGeometryArgs): EdgeGeometry {
   const arrowOffset = ARROW_OFFSETS[targetPosition];
   const isWaypoint = routing === 'waypoint';
   const isHandle = routing === 'handle';
+  const isSmoothStepPath = isHandle || (isWaypoint && waypoints.length === 0 && !enableSegments);
```
snuziale: This hunk makes the rendered edge shape depend on `editingEnabled`, which I think is a worse inconsistency than the diagonal it's fixing.

`CanvasEdge.tsx:129` passes `enableSegments: editingEnabled`, and `CanvasEdge.tsx:88` is:

```ts
const editingEnabled = !!data?.enableEditing && routing === 'waypoint' && !isReadOnly;
```

So for the same graph at the same coordinates, a waypointless edge renders as `buildPathVertices` output when `enableEditing` is true, and as React Flow smooth-step when `enableEditing` is false or `isReadOnly` is true. Viewer mode, locked canvases, and diff views would each show a different route, different corner rounding, and different endpoints than the editable canvas.

Three specifics:

- **Endpoints move.** `buildPathVertices` insets both ends by `ARROW_OFFSETS` (`geometry.ts:212-216`). The smooth-step branch passes raw `sourceX`/`sourceY` and insets the target only when `hideArrowHead`. So a waypointless edge attaches to its nodes differently than a sibling edge that happens to have one waypoint.
- **Both pipelines run.** The `pathPoints` memo is still gated on `isWaypoint`, not `isSmoothStepPath`, so `buildPathVertices` and `createRoundedPath` recompute on every coordinate change and are then discarded by the `pathPoints: EMPTY_VERTICES` early return on line 161. If the hunk stays, that memo should be gated on `!isSmoothStepPath`.
- **The docblock on lines 71-78 is now inaccurate.** It still says handle-routed edges are the only ones that skip the waypoint pipeline, and that `useEdgePath` "is fed constant inputs, its memo never invalidates and no smooth-step path is recomputed on coordinate changes." Neither holds for the new branch.

Stepping back: this hunk and the `getCollinearAxis` change fix the same bug twice, in ways that disagree. `getCollinearAxis` already fixes the diagonal fallback for every auto-routed CanvasEdge. This hunk then routes the non-editing subset around that fix. The new geometry tests cover the path editing consumers get, while this hunk hands everyone else a different one. There's also no test for the switch itself, and `CanvasEdge.test.tsx:28` mocks `useEdgeGeometry`, so nothing exercises it end to end.

Suggestion: drop this hunk and ship the `geometry.ts` fix on its own. If smooth-step really is wanted for auto-routed edges, an explicit option on `CanvasEdgeData` would be clearer than keying it off `editingEnabled`.

KodudulaAshishUiPath: Agreed, dropping the hunk. `useEdgeGeometry.ts` is reverted, so the PR is now just `getCollinearAxis` plus its tests.

You were right that the two fixes were redundant: the geometry tests pass unchanged with the hook reverted, so `getCollinearAxis` alone covers the diagonal for every consumer. Reverting also resolves the other three points, since the discarded `buildPathVertices`/`createRoundedPath` recompute, the endpoint-inset mismatch, and the stale docblock on lines 71-78 all existed only because of that branch.

Also made `getCollinearAxis` module-private, since nothing outside `geometry.ts` consumes it.

On smooth-step alignment with SequenceEdge: not pursuing it here. If we do want it, I'll take your suggestion of an explicit `CanvasEdgeData` option rather than keying it off `editingEnabled`, as a separate ticket.

snuziale (approved):

PR #1030, opened 2026-08-09
fix(apollo-react): improve stage return-edge snapping

jevinkosasih:
## Summary

- Make the StageNode bottom target use the standard 24px top/bottom hit window and let xyflow connection validity control pointer interaction
  - Previously, it was difficult to get a return edge to attach to the target stage
 

https://github.com/user-attachments/assets/5c7b1252-3dd1-4521-834f-80956f6b36db

 
- Render persisted and preview stage edges from the actual selected handle coordinates
  - Previously, dragging from the bottom handle would create an edge that starts from the right handle.

https://github.com/user-attachments/assets/d033cdd5-60e1-4748-8072-5a2ba39e30b0

- Keep the connection preview above stage nodes while dragging

Connection compatibility remains consumer-owned through isValidConnection; this change does not restrict bottom handles to specific target handles.

https://github.com/user-attachments/assets/b61c1678-3aa3-47af-9494-67cd6ce79251

## Testing

- focused existing ButtonHandle and StageNode tests: 107 passed
- typecheck passed
- Apollo React lint passed
- Apollo React build passed
- format check passed


packages/apollo-react/src/canvas/components/StageNode/StageEdge.tsx
```diff
@@ -42,15 +38,6 @@ interface StageEdgeInnerProps extends Omit<Props, 'sourceX' | 'sourceY' | 'targe
   geometry: StageEdgeGeometry;
 }
 
-function stageEdgeGeometryEquality(previous: StageEdgeGeometry, next: StageEdgeGeometry): boolean {
```
jevinkosasih, line 45: Removed because this equality helper only supported the useStore selector that rebuilt fixed upper-right and upper-left node geometry. StageEdge now consumes the live sourceX, sourceY, targetX, and targetY coordinates React Flow calculates from the registered handles, then memoizes that geometry object for StageEdgeInnerMemo. With no local store selector remaining, the custom equality function is unused dead code and React Flow remains responsible for updating the edge when a handle or node moves.

packages/apollo-react/src/canvas/components/StageNode/StageEdge.tsx
```diff
@@ -173,17 +160,10 @@ function StageEdgeComponent({
...
-    const sourceNode = state.nodeLookup.get(rest.source);
-    const targetNode = state.nodeLookup.get(rest.target);
-
-    return {
-      sourceX: sourceNode ? sourceNode.position.x + (sourceNode.measured?.width ?? 0) : sourceX,
-      sourceY: sourceNode ? sourceNode.position.y + 32 : sourceY,
-      targetX: targetNode?.position.x ?? targetX,
-      targetY: targetNode ? targetNode.position.y + 32 : targetY,
-    };
-  }, stageEdgeGeometryEquality);
```
jevinkosasih, line 186: Removed because this selector discarded the handle coordinates supplied in EdgeProps and rebuilt every stage edge as upper-right header to upper-left header: source x plus width and y plus 32, then target x and y plus 32. That made bottom-to-bottom return edges render from the header even though React Flow had registered the correct bottom-handle coordinates. React Flow already updates sourceX, sourceY, targetX, and targetY when handles or nodes move, so StageEdge now memoizes those supplied values and no longer needs a direct nodeLookup subscription.

packages/apollo-react/src/canvas/components/StageNode/StageConnectionEdge.tsx
```diff
@@ -24,26 +24,18 @@ export function StageConnectionEdge({
...
   // Check if we have valid coordinates
   if (fromX === undefined || fromY === undefined || toX === undefined || toY === undefined) {
     return null;
   }
 
-  const sourceX = fromNode ? fromNode.position.x + (fromNode.measured?.width ?? 0) : fromX;
-  const sourceY = fromNode?.position.y ? fromNode.position.y + 32 : fromY;
-
-  const targetX = toNode ? toNode.position.x : toX;
-  const targetY = toNode ? toNode.position.y + 32 : toY;
```
jevinkosasih, line 39: Removed because these calculations forced the connection preview from the source stage upper-right header to the target stage upper-left header whenever node objects were available, regardless of which handles were active. For a bottom-to-bottom return edge, React Flow already provides the exact registered bottom-handle coordinates through fromX, fromY, toX, and toY; overriding them here made the ghost edge originate and land at the header. Passing those coordinates directly preserves the selected handles, the live cursor or snapped target position, and also avoids the position.y equals zero truthiness issue.

kittyyueli (approved):

joelasher-ap (approved):

PR #1033, opened 2026-08-10
feat(apollo-react): freeze edges between read-only nodes

david-rios-uipath:
## Summary

Completes per-node read-only support by freezing graph connections wholly inside the locked region and documenting the complete behavior in Storybook.

## Demo

https://github.com/user-attachments/assets/efaaf744-ad11-4fb7-a82a-ec3351f0caf3

## Scope

- Freezes edges only when both source and target nodes are locked
- Prevents frozen-edge deletion, reconnection, creation, and edge-toolbar splitting
- Keeps connections involving an unlocked endpoint editable
- Guards connection callbacks using the current lock set, including locks applied after a gesture starts
- Exports `useIsConnectionReadOnly(source, target)` for custom edge renderers
- Adds the interactive per-node read-only Storybook demo and final API documentation

## Stack

Part 4 of 4. Depends on #1053. This is the original PR, narrowed to the final edge-policy and documentation layer.

## Testing

- 5 focused edge/core test files
- 48 tests passing after the final API-boundary adjustment
- The cumulative 13-file read-only regression set passed 335 tests

## Out of scope

- GroupNode resize/collapse gating
- Per-node reason codes for tooltip and accessibility messaging

### DIFF
--- a/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.hooks.ts
+++ b/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.hooks.ts
@@ -3,6 +3,8 @@ import { useReactFlow, useStoreApi } from '@uipath/apollo-react/canvas/xyflow/re
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { BASE_CANVAS_DEFAULTS, FIT_VIEW_DELAY_MS } from './BaseCanvas.constants';
 import type { BaseCanvasFitViewOptions, EnsureNodesInViewOptions } from './BaseCanvas.types';
+import { isConnectionReadOnly } from './ReadOnlyNodesContext';
+import { EMPTY_SET } from './set-utils';
 
 const waitForNodeMeasurements = (getNodes: () => Node[]): Promise<void> => {
   return new Promise((resolve) => {
@@ -342,19 +344,29 @@ export const useMaintainNodesInView = (
 };
 
 /**
- * Composes an `onBeforeDelete` that vetoes deletion of locked nodes.
+ * Composes an `onBeforeDelete` that vetoes deletion of locked nodes and frozen
+ * connections.
  *
- * Chosen over stamping `deletable: false` onto the nodes handed to React Flow.
- * That clone round-trips out through `getNodes()` and `toObject()`, so a
- * consumer saving the graph persists our lock as their own data, and unlocking
- * the node could never undo it. A veto touches nothing the consumer owns and
- * covers keyboard delete, `deleteElements`, and toolbar actions in one place.
+ * Chosen over stamping `deletable: false` onto the elements handed to React
+ * Flow. That clone round-trips out through `getNodes()`, `getEdges()`, and
+ * `toObject()`, so a consumer saving the graph persists our lock as their own
+ * data and unlocking could never undo it. A veto touches nothing the consumer
+ * owns and covers keyboard delete, `deleteElements`, and toolbar actions in one
+ * place.
  *
- * Edges need the same care: React Flow gathers the edges of every node in the
- * delete set, so vetoing a node without pruning its edges would leave it
- * standing with its connections gone. An edge is dropped when it touches a
- * vetoed node and no node that is actually being deleted, which keeps
- * mixed selections from stranding a dangling edge.
+ * Cascading edges need care of their own: React Flow gathers the edges of every
+ * node in the delete set, so vetoing a node without pruning its edges would
+ * leave it standing with its connections gone. An edge is dropped when it
+ * touches a vetoed node and no node that is actually being deleted, which stops
+ * a mixed selection from stranding a dangling edge.
+ *
+ * Not enforced here, deliberately:
+ * - `connectable`: React Flow never consults it when validating a connection,
+ *   and connecting to a locked node is allowed anyway.
+ * - `draggable`: position is layout, not content. Freezing it would also strand
+ *   locked nodes when a mixed selection is dragged.
+ * - reconnection: guarded by `useReadOnlyConnectionCallbacks`, which rejects the
+ *   gesture with the lock set read at completion time.
  */
 export function useReadOnlyBeforeDelete<NodeType extends Node, EdgeType extends Edge>(
   readOnlyNodeIds: ReadonlySet<string>,
@@ -375,19 +387,45 @@ export function useReadOnlyBeforeDelete<NodeType extends Node, EdgeType extends 
       for (const node of candidate.nodes) {
         (readOnlyNodeIds.has(node.id) ? vetoedNodeIds : deletedNodeIds).add(node.id);
       }
-      if (vetoedNodeIds.size === 0) return candidate;
 
       const touchesVetoedNode = (edge: EdgeType) =>
         vetoedNodeIds.has(edge.source) || vetoedNodeIds.has(edge.target);
       const touchesDeletedNode = (edge: EdgeType) =>
         deletedNodeIds.has(edge.source) || deletedNodeIds.has(edge.target);
+      const isFrozen = (edge: EdgeType) =>
+        isConnectionReadOnly(readOnlyNodeIds, edge.source, edge.target);
+
+      const keptEdges = candidate.edges.filter(
+        (edge) => !isFrozen(edge) && (!touchesVetoedNode(edge) || touchesDeletedNode(edge))
+      );
+      if (vetoedNodeIds.size === 0 && keptEdges.length === candidate.edges.length) {
+        return candidate;
+      }
 
       return {
         nodes: candidate.nodes.filter((node) => !vetoedNodeIds.has(node.id)),
-        edges: candidate.edges.filter(
-          (edge) => !touchesVetoedNode(edge) || touchesDeletedNode(edge)
-        ),
+        edges: keptEdges,
       };
     };
   }, [readOnlyNodeIds, onBeforeDelete]);
+}
+
+/**
+ * Ids of the edges to freeze: those whose source and target are both locked.
+ * An edge with either end unlocked is left editable.
+ */
+export function useReadOnlyEdgeIds<EdgeType extends Edge>(
+  edges: EdgeType[],
+  readOnlyNodeIds: ReadonlySet<string>
+): ReadonlySet<string> {
+  return useMemo(() => {
+    if (readOnlyNodeIds.size === 0) return EMPTY_SET;
+    const frozen = new Set<string>();
+    for (const edge of edges) {
+      if (isConnectionReadOnly(readOnlyNodeIds, edge.source, edge.target)) {
+        frozen.add(edge.id);
+      }
+    }
+    return frozen.size === 0 ? EMPTY_SET : frozen;
+  }, [edges, readOnlyNodeIds]);
 }
--- a/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
+++ b/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
@@ -22,12 +22,14 @@ import {
   useEnsureNodesInView,
   useMaintainNodesInView,
   useReadOnlyBeforeDelete,
+  useReadOnlyEdgeIds,
 } from './BaseCanvas.hooks';
 import type { BaseCanvasProps, BaseCanvasRef } from './BaseCanvas.types';
 import { CanvasBackground } from './CanvasBackground';
 import { CanvasProviders } from './CanvasProviders';
 import { PanShortcutTeachingUI } from './PanShortcutTeachingUI';
 import { useStableNodeIdSet } from './ReadOnlyNodesContext';
+import { useReadOnlyConnectionCallbacks } from './useReadOnlyConnectionCallbacks';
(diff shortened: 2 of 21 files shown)


packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
```diff
@@ -110,6 +117,25 @@ const BaseCanvasInnerComponent = <NodeType extends Node = Node, EdgeType extends
...
+  const stableReadOnlyNodeIds = useStableNodeIdSet(readOnlyNodeIds);
+  const decoratedNodes = useReadOnlyDecoratedNodes(nodes, stableReadOnlyNodeIds);
+
+  // Safety net: drop structural changes targeting read-only nodes even if a
+  // change slips past the per-node flags (e.g. programmatic removals).
+  const handleNodesChange = useMemo(() => {
+    if (!onNodesChange || stableReadOnlyNodeIds.size === 0) return onNodesChange;
+    const guarded: typeof onNodesChange = (changes) => {
+      const allowed = changes.filter((change) =>
+        change.type === 'remove' || change.type === 'position'
```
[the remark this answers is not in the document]
david-rios-uipath: Added in `2efcb72a` (now `BaseCanvas.readonlyChanges.test.tsx`), plus one clarification: only `remove` is filtered, not `position`. Locked nodes stay repositionable, since position is layout rather than content, and freezing it stranded them when a mixed selection was dragged (React Flow drops non-draggable nodes from the drag set). Both facts are now pinned by tests.

The new file captures the `onNodesChange` handler BaseCanvas passes to ReactFlow and calls it with synthetic changes:

- `remove` for a locked node is dropped while `remove` for an unlocked node passes through
- the handler is not invoked at all when every change is filtered out (covers the `allowed.length > 0` guard)
- `position` on a locked node passes through
- `select` and `dimensions` on a locked node pass through
- with no `readOnlyNodeIds`, the original handler is passed by identity, so no wrapper is allocated

packages/apollo-react/src/canvas/components/Toolbar/EdgeToolbar/useEdgeToolbarState.ts
```diff
@@ -45,6 +46,7 @@ export function useEdgeToolbarState({
   const { getManifestForNode } = useCanvasNodeLayout();
   const { mode } = useBaseCanvasMode();
   const isDesignMode = mode === 'design';
+  const isFrozenConnection = useIsConnectionReadOnly(source, target);
```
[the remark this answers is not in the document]
david-rios-uipath: Fixed in 51984e7f. The toolbar action now reads the current read-only policy at call time and verifies that the edge still has the same source, target, and connections before opening the add-node preview. The deferred preview also revalidates against the authoritative edge snapshot inside the queued edge update before applying either nodes or edges. This uses the current policy store rather than the rendered deletable/reconnectable flags, which can lag for the same reason. Regression coverage includes locks changing after the action is created, reconnection and target-connection changes, and an earlier queued reconnection.

packages/apollo-react/src/canvas/components/Toolbar/EdgeToolbar/useEdgeToolbarState.ts
```diff
@@ -29,6 +33,17 @@ export interface EdgeToolbarState {
...
+type EdgeConnection = Pick<Edge, 'source' | 'sourceHandle' | 'target' | 'targetHandle'>;
+
+function hasSameConnection(edge: Edge | undefined, expected: EdgeConnection): edge is Edge {
+  return (
+    edge?.source === expected.source &&
+    edge.target === expected.target &&
+    (edge.sourceHandle ?? null) === (expected.sourceHandle ?? null) &&
+    (edge.targetHandle ?? null) === (expected.targetHandle ?? null)
+  );
+}
```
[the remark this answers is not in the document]
david-rios-uipath: Good catch. `hasSameConnection` now returns `false` up front when `edge` is falsy and compares `edge.source` / `edge.target` / handles directly. Squashed into the original commit.

packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.hooks.ts
```diff
@@ -1,8 +1,11 @@
-import type { Node } from '@uipath/apollo-react/canvas/xyflow/react';
+import type { Edge, Node } from '@uipath/apollo-react/canvas/xyflow/react';
 import { useReactFlow, useStoreApi } from '@uipath/apollo-react/canvas/xyflow/react';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { BASE_CANVAS_DEFAULTS, FIT_VIEW_DELAY_MS } from './BaseCanvas.constants';
 import type { BaseCanvasFitViewOptions, EnsureNodesInViewOptions } from './BaseCanvas.types';
+import { isConnectionReadOnly } from './ReadOnlyNodesContext';
+
+const EMPTY_EDGE_IDS: ReadonlySet<string> = new Set();
```
snuziale: object freeze

packages/apollo-react/src/canvas/components/BaseCanvas/ReadOnlyNodesContext.tsx
```diff
@@ -4,13 +4,15 @@ import {
...
   useEffect,
+  useLayoutEffect,
   useRef,
   useSyncExternalStore,
 } from 'react';
 
 type Listener = () => void;
 
-const EMPTY_SET: ReadonlySet<string> = new Set();
+const EMPTY_SET: ReadonlySet<string> = Object.freeze(new Set()) as ReadonlySet<string>;
```
Holychung: The freeze is back here, cast and all. You took it out on 1051 after Copilot flagged it, on the grounds that it's decorative on a Set and implies a guarantee it doesn't give. Rebase artifact or did something change your mind? EMPTY_EDGE_IDS in BaseCanvas.hooks.ts is unfrozen, so we've got both idioms sitting next to each other now.

packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
```diff
@@ -241,8 +275,10 @@ const BaseCanvasInnerComponent = <NodeType extends Node = Node, EdgeType extends
         panOnDrag={isInteractive ? PAN_ON_DRAG : false}
         onInit={handleInit}
         onNodesChange={isInteractive ? handleNodesChange : undefined}
-        onEdgesChange={isInteractive ? onEdgesChange : undefined}
-        onConnect={isDesignMode ? onConnect : undefined}
+        onEdgesChange={isInteractive ? handleEdgesChange : undefined}
+        onConnect={isDesignMode ? guardedOnConnect : undefined}
+        onReconnect={guardedOnReconnect}
```
Holychung: onReconnect and onReconnectEnd are the only connection callbacks here not gated on isDesignMode. edgesReconnectable defaults to true and xyflow only needs onReconnect to be present to render the reconnect anchors, so view and readonly mode end up with live edge endpoints. Intentional?

packages/apollo-react/src/canvas/utils/createPreviewGraph.ts
```diff
@@ -217,21 +224,36 @@ export function showPreviewGraph(options: CreatePreviewGraphOptions): PreviewGra
...
+  setTimeout(() => {
+    if (!canApply) {
+      applyNodes();
+      reactFlowInstance.setEdges(applyEdges);
+      return;
+    }
+
+    reactFlowInstance.setEdges((edges) => {
+      if (!canApply(edges)) return edges;
+      applyNodes();
```
Holychung: applyNodes() is calling setNodes from inside the setEdges updater. Updaters are meant to be pure, and getEdges() in the same setTimeout reads the same store, so I'm not sure the updater is buying extra atomicity here. What drove putting the check inside it?

packages/apollo-react/src/canvas/components/BaseCanvas/ReadOnlyNodesContext.tsx
```diff
@@ -4,13 +4,15 @@ import {
...
+const useIsomorphicLayoutEffect = typeof window === 'undefined' ? useEffect : useLayoutEffect;
```
Holychung: We've got two of these now, this one and hooks/useLatestRef.ts:3, spelled with the condition inverted. Both module-private. This PR imports useLatestRef over in useReadOnlyConnectionCallbacks, so worth pulling one out while you're in there?

packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.hooks.ts
david-rios-uipath: centralized `EMPTY_SET` and replaced `EMPTY_EDGE_IDS` w/ it.

packages/apollo-react/src/canvas/components/BaseCanvas/ReadOnlyNodesContext.tsx
david-rios-uipath: same as above comment

packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
david-rios-uipath: Fixed in `c93a416a`. Both are now gated on `isDesignMode`, same as `onConnect` / `onConnectStart` / `onConnectEnd`.

Added `BaseCanvas.reconnect.test.tsx` for it. It asserts on the props handed to `ReactFlow` rather than the anchor DOM: React Flow only renders reconnect anchors for edges it can measure, and jsdom reports no geometry, so a DOM assertion would pass vacuously. The prop is the actual switch, since `edgesReconnectable` defaults to true and the anchors appear purely because `onReconnect` is defined.

packages/apollo-react/src/canvas/utils/createPreviewGraph.ts
david-rios-uipath: You are right, and it is fixed in `c93a416a`. The guard now runs before either setter:

```ts
if (canApply?.(reactFlowInstance.getEdges()) === false) return;

applyNodes();
reactFlowInstance.setEdges(applyEdges);
```

`getEdges()` reads the same store the updater would have seen, so the check loses nothing, and the nested `setNodes` inside an updater is gone. It also fixes an ordering bug the old shape had: with a guard present, nodes were added from inside the edge updater, so edges could reach the store referencing a node that was not there yet. Added an `it.each` case asserting `setNodes` runs before `setEdges` both with and without a guard.

packages/apollo-react/src/canvas/components/BaseCanvas/ReadOnlyNodesContext.tsx
david-rios-uipath: Deduped. There is now a single `canvas/hooks/useIsomorphicLayoutEffect.ts`, and both `useLatestRef` and this file import it.

The extraction lands one PR lower, in #1051 (`2814a929`), next to the `set-utils` sharing commit. `useLatestRef.ts` is on `main` and has nothing to do with read-only nodes, so putting the refactor here would have mixed an unrelated file into this PR's diff. Kept module-private, matching `useLatestRef`, so it is not in the `hooks` barrel.

Holychung (approved):

packages/apollo-react/src/canvas/components/BaseCanvas/useReadOnlyConnectionCallbacks.ts
```diff
@@ -0,0 +1,104 @@
...
+      if (
+        !currentOnReconnectEnd ||
+        includesReadOnlyConnection(currentReadOnlyNodeIds, edgeBeingReconnected)
+      ) {
+        return;
+      }
+      currentOnReconnectEnd(event, edgeBeingReconnected, handleType, connectionState);
+    },
+    [onReconnectEndRef, readOnlyNodeIdsRef]
+  );
```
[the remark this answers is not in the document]
david-rios-uipath: Intentional asymmetry, keeping as-is.

The sequence is real, but it is not specific to read-only policy: dropping on empty canvas, or on a target the consumer's own `isValidConnection` rejects, takes the exact same path. Read-only just makes one more target invalid. Consumers who delete-on-no-reconnect opt into that, and `connectionState` tells them whether anything landed.

No read-only invariant is broken either: the edge deleted is editable, on editable nodes, by a gesture the user started on it. `guardedOnReconnect` guards what would be *created*; `guardedOnReconnectEnd` guards what would be *destroyed*, which is only ever the original edge.

Checking the proposed target in `onReconnectEnd` also is not reliably possible: `FinalConnectionState` is `Omit<ConnectionInProgress | NoConnection, 'inProgress'>`, so a drop-to-nowhere carries no `toNode` at all. And where it does, suppressing `onReconnectEnd` would swallow the consumer's end-of-gesture cleanup and leave an editable edge stuck mid-reconnect, which is worse.

Added a one-line comment above `guardedOnReconnectEnd` documenting why only the original edge is checked.

Holychung (approved):

Holychung (approved):

Holychung (approved):

PR #1036, opened 2026-08-11
fix(apollo-react): restore stage task tooltips and surface exact durations on hover

kittyyueli:
Tooltips missing on a stage task row: the task name, the run/rework chips, and the duration text. Three separate causes, fixed here.

## Root causes

**1. The name tooltip measured the wrong element**

`smartTooltip` measures its own trigger for overflow, and the trigger was a wrapper `Row`. A wrapper never overflows: the `truncate` span inside it ellipsises instead, so `scrollWidth === clientWidth` and truncation always evaluated to false. Every task name was suppressed, regardless of length. The stage rule labels and the collapsible section headers had the identical pattern.

**2. Refs were dropped on React 18 consumers**

Radix hands an `asChild` trigger its ref through `cloneElement`. `Badge` (the run/rework chips) and `Row`/`Column` were plain function components, and React 18 drops refs given to those, logging *&#34;Function components cannot be given refs&#34;*. Radix&#39;s popper was left without an anchor, so the chip tooltip never became visible.

This is why the neighbouring status-icon and play-button tooltips worked: `Button` already used `forwardRef`. It also explains why the bug does not reproduce in Storybook, which runs React 19, where the ref survives as an ordinary prop.

**3. Durations had no tooltip at all**

A duration renders only its **three largest units**, so anything below them (`2w, 19h, 8m` hiding 42 seconds; `Duration: 3w, 1d, 20h` hiding minutes and seconds) was readable nowhere. There was no tooltip unless the consumer passed `durationTooltip`, which PO.Frontend only does for wait-for-timer countdowns.

## Changes

- `apollo-wind` `Badge`: `forwardRef` so it can anchor a Radix popup.
- `apollo-react` `Stack`/`Row`/`Column`: `forwardRef`, so a stack can be a Radix trigger and a measurement target.
- `TaskContent`, `StageRuleContent`, `CollapsibleStageHeader`: the tooltip now wraps the truncating span itself instead of a wrapper row.
- `formatDuration`: new `formatExactDurationMs` (all units, spelled out) and `hasHiddenDurationParts` (did the cap drop anything?).
- `TaskContent` and `StageNodeHeader`: hovering a duration shows the exact value, e.g. `2 weeks, 19 hours, 8 minutes, 42 seconds`.

Nothing is added where it would only repeat what is already on screen: the name tooltip stays `smartTooltip` (truncated names only), and a duration the row already shows in full gains no tooltip. A consumer-supplied `durationTooltip` still wins, since a countdown says something the duration itself cannot. Legacy pre-formatted `duration` strings get no tooltip: without the milliseconds there is nothing to be exact about.

Read-only rows (monitoring, instance view) render the same `TaskContent`, so they are covered by the same fixes.

## Testing

- `TaskContent.integration.test.tsx` and `StageNodeHeader.integration.test.tsx` unmock the globally-stubbed Radix tooltip and assert real open/closed state: truncated name opens, untruncated stays closed, runs chip opens, rework chip opens, a read-only task row opens both, and a duration opens only when units were dropped.
- Verified the new tests fail on the pre-fix code (6 across the two rounds; the chip cases cannot fail under the repo&#39;s React 19 dev dependency, which is the React-18-specific half of the bug).
- `formatExactDurationMs` / `hasHiddenDurationParts` unit-tested including locale wording (German joins with &#34;und&#34;), sub-second rounding, and zero units between parts.
- `Row`/`Column`/`Badge` ref-forwarding tests added. `layouts/index.test.ts` now asserts the stacks render rather than asserting `typeof === &#39;function&#39;`, which no longer holds for `forwardRef` components.
- Full suites green: apollo-react 2393+ tests, apollo-wind 1073 tests. `tsc --noEmit` clean, Biome clean on all changed files.

## Consumer note

PO.Frontend pins `@uipath/apollo-react` 6.16.5 and needs a version bump to pick this up once apollo-react publishes.

## Demo

https://github.com/user-attachments/assets/d4c71117-00f5-42c2-8936-d8d59e7baf25

### DIFF
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.integration.test.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.integration.test.tsx
@@ -1,5 +1,6 @@
 import { ReactFlowProvider } from '@uipath/apollo-react/canvas/xyflow/react';
 import { afterEach, describe, expect, it, vi } from 'vitest';
+import { DAY, HOUR, MINUTE, SECOND, WEEK } from '../../../test/durations';
 import { act, fireEvent, render, screen } from '../../utils/testing';
 import type { StageNodeProps } from './StageNode.types';
 import { StageNodeHeader } from './StageNodeHeader';
@@ -8,12 +9,6 @@ import { StageNodeHeader } from './StageNodeHeader';
 // opens only when the header dropped units. Radix does not portal in happy-dom, so assertions
 // use the trigger's `data-state`.
 vi.unmock('@uipath/apollo-wind/components/ui/tooltip');
-
-const SECOND = 1000;
-const MINUTE = 60 * SECOND;
-const HOUR = 60 * MINUTE;
-const DAY = 24 * HOUR;
-const WEEK = 7 * DAY;
 
 const renderHeader = (durationMs: number) => {
   const props = {
--- a/packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.integration.test.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.integration.test.tsx
@@ -1,0 +1,72 @@
+import { ReactFlowProvider } from '@uipath/apollo-react/canvas/xyflow/react';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { act, fireEvent, render, screen } from '../../utils/testing';
+import type { StageNodeProps } from './StageNode.types';
+import { StageNodeHeader } from './StageNodeHeader';
+
+// Uses the real Radix tooltip (the shared setup stubs it out) to cover that the exact duration
+// opens only when the header dropped units. Radix does not portal in happy-dom, so assertions
+// use the trigger's `data-state`.
+vi.unmock('@uipath/apollo-wind/components/ui/tooltip');
+
+const SECOND = 1000;
+const MINUTE = 60 * SECOND;
+const HOUR = 60 * MINUTE;
+const DAY = 24 * HOUR;
+const WEEK = 7 * DAY;
+
+const renderHeader = (durationMs: number) => {
+  const props = {
+    id: 'stage-1',
+    stageDetails: { label: 'Stage 1', tasks: [] },
+    execution: { stageStatus: { durationMs }, taskStatus: {} },
+  } as unknown as StageNodeProps;
+
+  return render(
+    <ReactFlowProvider>
+      <StageNodeHeader
+        props={props}
+        isReadOnly={true}
+        status={undefined}
+        handleTaskAddClick={vi.fn()}
+      />
+    </ReactFlowProvider>
+  );
+};
+
+function hover(element: HTMLElement) {
+  fireEvent.mouseEnter(element);
+  fireEvent.pointerMove(element, { pointerType: 'mouse' });
+  act(() => {
+    vi.advanceTimersByTime(500);
+  });
+}
+
+describe('StageNodeHeader duration tooltip (real CanvasTooltip)', () => {
+  afterEach(() => {
+    vi.useRealTimers();
+    vi.restoreAllMocks();
+  });
+
+  it('opens the exact duration when the header dropped units', () => {
+    vi.useFakeTimers();
+    renderHeader(3 * WEEK + 1 * DAY + 20 * HOUR + 14 * MINUTE + 3 * SECOND);
+
+    const duration = screen.getByTestId('stage-duration-stage-1');
+    expect(duration).toHaveTextContent('3w, 1d, 20h');
+
+    hover(duration);
+    expect(duration).toHaveAttribute('data-state', 'delayed-open');
+  });
+
+  it('stays closed when the header already shows every unit', () => {
+    vi.useFakeTimers();
+    renderHeader(2 * HOUR + 4 * MINUTE);
+
+    const duration = screen.getByTestId('stage-duration-stage-1');
+    expect(duration).toHaveTextContent('2h, 4m');
+
+    hover(duration);
+    expect(duration).toHaveAttribute('data-state', 'closed');
+  });
+});
--- a/packages/apollo-react/src/canvas/components/StageNode/rules/StageRuleContent.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/rules/StageRuleContent.tsx
@@ -36,23 +36,13 @@ export const StageRuleContent = memo(({ rule, ruleType }: StageRuleContentProps)
       >
         <StageItemIcon data-testid={`stage-rule-icon-${rule.id}`}>{icon}</StageItemIcon>
         <CanvasTooltip content={rule.label} placement="top" smartTooltip>
-          <Row
-            gap={'2px'}
-            align="center"
-            style={{
-              overflow: 'hidden',
-              textOverflow: 'ellipsis',
-              whiteSpace: 'nowrap',
-            }}
+          <span
+            data-testid={`stage-rule-label-${rule.id}`}
+            className="text-sm truncate"
+            style={{ minWidth: 0 }}
           >
-            <span
-              data-testid={`stage-rule-label-${rule.id}`}
-              className="text-sm truncate"
-              style={{ minWidth: 0 }}
-            >
-              {rule.label}
-            </span>
-          </Row>
+            {rule.label}
+          </span>
         </CanvasTooltip>
       </Row>
     </Row>
--- a/packages/apollo-react/src/canvas/components/StageNode/shared/CollapsibleStageHeader.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/shared/CollapsibleStageHeader.tsx
@@ -65,23 +65,13 @@ export const CollapsibleStageHeader = ({
           />
         </span>
         <CanvasTooltip content={label} placement="top" smartTooltip>
-          <Row
-            gap={'2px'}
-            align="center"
-            style={{
-              overflow: 'hidden',
-              textOverflow: 'ellipsis',
-              whiteSpace: 'nowrap',
-            }}
+          <span
+            data-testid={`${testId}-label`}
+            className="text-sm truncate"
+            style={{ minWidth: 0 }}
           >
-            <span
-              data-testid={`${testId}-label`}
-              className="text-sm truncate"
-              style={{ minWidth: 0 }}
-            >
-              {label}
-            </span>
-          </Row>
+            {label}
+          </span>
         </CanvasTooltip>
       </StyledRow>
       <div
--- a/packages/apollo-react/src/canvas/components/StageNode/tasks/TaskContent.tsx
+++ b/packages/apollo-react/src/canvas/components/StageNode/tasks/TaskContent.tsx
@@ -7,7 +7,7 @@ import { memo, useCallback, useMemo, useState } from 'react';
 import { useSafeLingui } from '../../../../i18n';
 import { CanvasTooltip } from '../../CanvasTooltip';
 import { ExecutionStatusIcon } from '../../ExecutionStatusIcon';
-import { formatDurationMs } from '../formatDuration';
+import { formatDurationMs, formatExactDurationMs, hasHiddenDurationParts } from '../formatDuration';
 import { StageItemIcon } from '../StageNode.styles';
 import type { StageTaskExecution, StageTaskItem } from '../StageNode.types';
 import { useExecutionStatusLabel } from '../useExecutionStatusLabel';
@@ -133,6 +133,11 @@ export const TaskContent = memo(
         {durationText}
       </span>
     ) : null;
+    const exactDuration =
+      taskExecution?.durationMs !== undefined && hasHiddenDurationParts(taskExecution.durationMs)
+        ? formatExactDurationMs(taskExecution.durationMs, locale)
+        : undefined;
+    const durationTooltip = taskExecution?.durationTooltip || exactDuration;
 
     return (
       <Row
@@ -195,12 +200,8 @@ export const TaskContent = memo(
           data-testid={`stage-task-actions-${task.id}`}
         >
           {durationLabel &&
-            (taskExecution?.durationTooltip ? (
-              <CanvasTooltip
-                content={taskExecution.durationTooltip}
-                placement="top"
-                hide={isDragging}
-              >
+            (durationTooltip ? (
+              <CanvasTooltip content={durationTooltip} placement="top" hide={isDragging}>
                 {durationLabel}
               </CanvasTooltip>
             ) : (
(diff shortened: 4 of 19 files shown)


packages/apollo-react/src/canvas/components/StageNode/shared/CollapsibleStageHeader.tsx
```diff
@@ -64,24 +64,17 @@ export const CollapsibleStageHeader = ({
             color="var(--color-foreground-de-emp)"
           />
         </span>
+        {/* The tooltip wraps the truncating span itself: `smartTooltip` measures its own
+            trigger, and a wrapper around the span never overflows because the span
+            ellipsises instead, which would always read as "not truncated". */}
```
kittyyueli: remove

packages/apollo-react/src/canvas/components/StageNode/rules/StageRuleContent.tsx
```diff
@@ -35,24 +35,17 @@ export const StageRuleContent = memo(({ rule, ruleType }: StageRuleContentProps)
         style={{ overflow: 'hidden', textOverflow: 'ellipsis', whiteSpace: 'nowrap' }}
       >
         <StageItemIcon data-testid={`stage-rule-icon-${rule.id}`}>{icon}</StageItemIcon>
+        {/* The tooltip wraps the truncating span itself: `smartTooltip` measures its own
+            trigger, and a wrapper around the span never overflows because the span
+            ellipsises instead, which would always read as "not truncated". */}
```
kittyyueli: remove

packages/apollo-react/src/canvas/components/StageNode/tasks/TaskContent.tsx
```diff
@@ -155,20 +164,22 @@ export const TaskContent = memo(
...
-              gap={'2px'}
-              align="center"
-              style={{
-                overflow: 'hidden',
-                textOverflow: 'ellipsis',
-                whiteSpace: 'nowrap',
-              }}
+            {/* The tooltip wraps the truncating span itself. `smartTooltip` measures its own
+                trigger, and this wrapper row never overflows: the span inside it ellipsises
+                instead, so anchoring the tooltip on the row would always read "not truncated". */}
```
kittyyueli: remove comment

packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.integration.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+// Exercises the real Radix tooltip (the shared test setup stubs it out) so the stage duration's
+// hover behaviour is covered: the exact duration opens only when the header had to drop units.
+// Radix does not portal its content in happy-dom, so the trigger's `data-state` is the signal.
+vi.unmock('@uipath/apollo-wind/components/ui/tooltip');
+
+const SECOND = 1000;
+const MINUTE = 60 * SECOND;
+const HOUR = 60 * MINUTE;
+const DAY = 24 * HOUR;
+const WEEK = 7 * DAY;
```
kittyyueli: these consts are duplicated everywhere. just export and reuse

packages/apollo-wind/src/components/ui/badge.tsx
```diff
@@ -32,8 +32,14 @@ export interface BadgeProps
   extends React.HTMLAttributes<HTMLDivElement>,
     VariantProps<typeof badgeVariants> {}
 
-function Badge({ className, variant, ...props }: BadgeProps) {
-  return <div className={cn(badgeVariants({ variant }), className)} {...props} />;
-}
+// forwardRef so the badge can be a Radix `asChild` trigger (tooltip, popover, …).
+// Without it React 18 consumers silently drop the ref, leaving the tooltip
+// without an anchor to position against, so it never becomes visible.
```
kittyyueli, line 37: Is this going to break anything?

kittyyueli: Checked rather than assumed. Short answer: one narrow difference, and nothing in this repo depends on it.

**Unchanged:** the rendered output is byte-identical. Same `div`, same `cn(badgeVariants({ variant }), className)`, same prop spread. No Badge story moved in the visual diff, and all 1073 apollo-wind tests pass.

**The one real difference:** `typeof Badge` is now `'object'` instead of `'function'`, because a `forwardRef` component is an exotic object rather than a plain function. That is not hypothetical: the same change to `Row`/`Column` in this PR broke `layouts/index.test.ts`, which asserted `typeof Row === 'function'` (fixed here by asserting it renders instead). So the risk is specifically code that introspects the component rather than rendering it. I grepped the monorepo: no `typeof Badge` checks, and Badge is only ever used as JSX, never called directly.

**For consumers:** strictly additive. No prop was added, removed, or renamed. A consumer passing `ref` on React 18 previously got a dev warning and had the ref silently dropped (the bug this fixes); now it reaches the `div`. On React 19, `ref` already arrived via the props spread, so behaviour there is unchanged.

**Precedent:** 45 of the 156 components in `apollo-wind/src/components/ui` already use `forwardRef`, including `Button`, which is why the neighbouring status-icon and play-button tooltips worked while the chips did not.


---

packages/apollo-react/src/canvas/components/StageNode/tasks/TaskContent.tsx
kittyyueli: Removed, in all three places it appeared (`TaskContent`, `StageRuleContent`, `CollapsibleStageHeader`), folded into `317f3e4`. The rationale now lives only in the commit message and the PR description.


---

packages/apollo-react/src/canvas/components/StageNode/StageNodeHeader.integration.test.tsx
kittyyueli: Done in `09bdcaf`: they come from `src/test/durations.ts` now (`SECOND`/`MINUTE`/`HOUR`/`DAY`/`WEEK`).

While there I also took out the millisecond literals in the two files that were spelling the same durations out by hand, so all four use the shared constants:

```diff
-durationMs: 2 * 86400000 + 3 * 3600000 + 4 * 60000 + 5000,
+durationMs: 2 * DAY + 3 * HOUR + 4 * MINUTE + 5 * SECOND,
```


---

kittyyueli:
Note on the visual diff, since it is a review gate here: the 10 flagged stories are not caused by this PR, and I would not hold merge on them.

**No StageNode story has changed in any of the three runs.** The task row and stage header are what this PR modifies, and they render pixel-identically, which is the intended outcome.

**None of the flagged stories live in a file this PR touches.** The changed set is canvas edges, LoopNode, the node property panel, and two canvas templates. The diff touches only `StageNode/*`, `canvas/layouts/Stack.tsx`, `src/test/durations.ts`, and `apollo-wind` `badge.tsx`.

**The flagged set is unstable across runs on identical code**, which is what identifies it as nondeterministic rather than real:

| Story | pre-rebase run 1 | pre-rebase run 2 | post-rebase run 3 |
| --- | --- | --- | --- |
| `canvasedge--execution-states` [light] | 0.013% | 0.014% | not flagged |
| `loopnode--nested-outer-output-insert` [dark] | not flagged | 1.973% | not flagged |
| `loopnode--nested-outer-output-append` [light] | not flagged | 1.505% | not flagged |
| `canvas-with-panels--properties` [dark] | not flagged | not flagged | 1.014% |

Since the rebase, this branch's LoopNode, agent-flow, canvas-with-panels, edge and panel code is identical to `main`, so a nonzero delta there cannot be a code difference. The ~1-2% entries are all dark-theme stories with animated or async content (timeline player, edge routing); the sub-0.05% ones are anti-aliasing on edge strokes.

The count also moved 6 → 13 → 10 partly because `main` advanced 8 commits (a release plus two canvas fixes) while this branch sat at its original base. The rebase onto `e8f2f47` removed that source of drift.

Both of this PR's production changes are ref plumbing only: same elements, same classes, same inline styles. Nothing in them can move a pixel, in these stories or anywhere else.


---

kittyyueli:
`Deploy apollo-design` failed on `79b6591` for an infrastructure reason, not a code one. I have re-run the failed jobs.

Everything up to the deploy succeeded:

```
Storybook build completed successfully      (Vite ✓ built in 3m 7s, Tasks: 5 successful, 5 total)
✔ Package uploaded to Orchestrator
✔ Coded app registered successfully!        (Deploy Version: 5)
==> Deploying Coded App
Error: app upgrade failed with HTTP 504 ... staging.uipath.com | 504: Gateway time-out
```

The Cloudflare error page names the origin as the failing hop: Browser `Working`, Cloudflare `Working`, Host `staging.uipath.com` `Error`, with "The web server reported a gateway time-out error. Please try again in a few minutes." So the build and the package were fine; the staging app-upgrade endpoint timed out.

The knock-on effect is that the Storybook visual diff skipped, since it needs the apollo-design preview to be up. The re-run should restore both.

Worth noting this failure mode is already known here: `afea0a5` bumped uip-go to 0.3.2 for "gateway-timeout deploy recovery", and this job did run 0.3.2, so the retry budget was simply exhausted by a longer outage. Nothing in this PR touches the deploy path.


---

kittyyueli:
Follow-up to my earlier note on the visual diff: this run settles it. 8 changed, still **no StageNode story** (the string does not appear anywhere in the run log), and the flagged set shifted again on code that did not change:

- `node-property-panel--full-editor` was flagged in all three previous runs and is **gone** in this one.
- `loopnode--nested-outer-output-append [light]` is **back** at 1.505%, having been absent last run.
- `canvas-with-panels--properties [dark]` moved from **1.014% to 0.639%**. A code-caused diff does not change magnitude between runs on identical bytes.
- A brand-new entry appeared, `apollo-wind-templates-future-layout-generator--visual [light]` at 0.937%, in a package where the only change in this PR is `badge.tsx` and which was untouched between the last two runs.

So the pattern is nondeterministic rendering in animated/async stories, not regressions. The components this PR actually modifies have produced zero pixel delta in all four runs.

Separately: `Deploy apollo-design` went green on the retry after the 504, and the remaining red job is `Update Apollo Coded App Preview Comment`, which hit `HttpError: status 500` from `GET /issues/1036/comments`. That job only refreshes the preview-links table, so it does not affect correctness. I have re-run it.


---

kittyyueli:
Heads-up on the last commit, `d77c9b6`, because it touches shared build config rather than this PR's feature. Happy to split it into its own PR if you would rather review it separately.

**Why it is here.** The duration tooltip story added in `74186d1` did not appear in the preview Storybook. The cause is not the story: the shared `build` task excluded `src/**/*.stories.*` from its inputs, and `storybook:build` depends on `^build`, so a stories-only change left every hash untouched and turbo replayed a cached Storybook.

Evidence from CI on that commit:

```
Tasks:    5 successful, 5 total
Cached:    5 cached, 5 total
  Time:    2.487s >>> FULL TURBO
Vite storybook-static/assets/StageNode.stories-DMSDpLrD.js   119.86 kB
```

That chunk hash is byte-identical to the run before the story existed, and the visual diff reported `ADDED: 0` with the same 267 stories and 534 comparisons as the previous run.

**Reproduced and fixed with turbo's own hashes**, no guessing:

| | `storybook:build` hash |
| --- | --- |
| before, clean tree | `56b74d4526267fda` |
| before, after editing a story | `56b74d4526267fda` (unchanged, hence the stale preview) |
| after, clean tree | `08cb874ef70d9364` |
| after, after editing a story | `0d6872779768f48e` (rebuilds) |

The fix drops only the stories exclusion. Test and spec files stay excluded, and I verified a test edit still does not trigger a rebuild, so the part of the optimisation that matters for CI time is intact. The cost is that stories-only changes now rebuild the libraries they belong to.

Worth knowing beyond this PR: until this lands, any stories-only change in this repo deploys a stale preview and gets a visual diff computed against the previous bundle.


---

packages/apollo-react/src/canvas/components/StageNode/tasks/TaskContent.tsx
```diff
@@ -133,6 +133,15 @@ export const TaskContent = memo(
         {durationText}
       </span>
     ) : null;
+    // The row shows only the largest units, so anything below them is readable nowhere else.
+    // A consumer-supplied `durationTooltip` (e.g. a wait-for-timer countdown) says something the
+    // duration itself cannot, so it wins; the exact duration fills in when there is no such text
+    // and parts were actually dropped. Nothing is added when the row already shows every part.
```
kittyyueli: remove

turbo.json
```diff
@@ -8,7 +8,6 @@
       "env": ["APOLLO_CODED_APP", "APOLLO_CODED_APP_PATH", "UIP_GO_PLATFORM_AUTH_*", "UIP_GO_AICHAT_*"],
       "inputs": [
         "$TURBO_DEFAULT$",
-        "!src/**/*.stories.*",
```
kittyyueli, line 11: PR preview doesn't rebuild if we only update stories in a commit

packages/apollo-react/src/canvas/components/StageNode/tasks/TaskContent.tsx
kittyyueli: Removed. I took the whole four-line block rather than only line 139, since the last sentence does not stand alone, and did the same to its counterpart above `exactStageDuration` in `StageNodeHeader.tsx`, which said the same thing in the same position.

I left the comment just above it, on lines 122-123 about `durationMs` winning over the legacy string, because `git blame` puts it on `6890d1b`, before this branch. Say the word if you want that one gone too and I will take it in a separate commit.

While here I also shortened the remaining comments this PR adds, per your other note: the two `formatDuration` JSDoc blocks, both integration-test headers, the `durations.ts` doc, and the story description. Each is one or two lines now. Folded into the commits that introduced them, so the history stays one commit per idea.


---

turbo.json
kittyyueli: That is exactly the bug, and it is what this line fixes. It bit this PR: the duration-tooltip story was added, and the preview deployed without it.

The cause is that `storybook:build` depends on `^build`, so the Storybook app's hash is built from its dependencies' `build` hashes, and `build` excluded `src/**/*.stories.*` from its inputs. A stories-only change therefore moved no hash at all and turbo replayed a cached Storybook. On the story commit the deploy log read `5 cached, 5 total`, `2.487s >>> FULL TURBO`, with an unchanged `StageNode.stories-DMSDpLrD.js`, and the visual diff reported `ADDED: 0`.

Verified with turbo's own hashes rather than by reasoning about it:

| | `storybook:build` hash |
| --- | --- |
| before, clean tree | `56b74d4526267fda` |
| before, after editing a story | `56b74d4526267fda`, unchanged |
| after, clean tree | `08cb874ef70d9364` |
| after, after editing a story | `0d6872779768f48e`, rebuilds |

Confirmed end to end on the next run: Vite spent 2m31s, turbo reported `1 cached, 5 total`, and the visual diff logged `NEW apollo-react-canvas-components-nodes-stagenode--duration-tooltips`. I re-ran the hash check after today's rebase and it still holds.

Test and spec files stay excluded, and I checked that a test-only edit still does not trigger a rebuild, so the part of the optimisation that saves CI time is intact. The cost is that a stories-only change now rebuilds the libraries it belongs to.

One note on this hunk: I collapsed the array onto one line because that is what `biome format` wants at `lineWidth: 100` once the third entry is gone. The `env` line above it is a pre-existing format deviation on `main`, which I left alone. Neither is caught today, since `turbo run lint` runs per package and root `turbo.json` is outside every package.


---

Miyuki-L (approved):

PR #1038, opened 2026-08-11
fix(apollo-wind): resolve the portal container in Dialog, Sheet and AlertDialog

rares-baesu-uipath:
## Problem

`PortalContainerProvider` / `useResolvedPortalContainer` already exist in this package, and
`Popover`, `Select` and `DropdownMenu` already consume them. `Dialog`, `Sheet` and `AlertDialog`
do not, so their overlays always portal to `document.body`.

For a consumer rendering inside a **shadow root** that is unreachable: the stylesheet it injects
into its own root cannot reach `document.body`, so every dialog and sheet renders unstyled —
transparent backdrop, no tokens, no layout.

The consumer cannot work around it either. `DialogContent`, `SheetContent` and
`AlertDialogContent` each render their own portal and forward `{...props}` to the inner Radix
`Content`, **not** to the `Portal`. The only escape is to stop using `*Content` and re-implement
its composition (overlay, close button, the class-name strings) at every call site — which then
drifts with every apollo-wind release.

## Change

Each `*Content` takes the same optional prop `PopoverContent` already exposes, resolves it through
the shared hook, and hands the result to its portal:

```tsx
const AlertDialogContent = React.forwardRef<…, … & { container?: PortalContainerOverride }>(
  ({ className, container, ...props }, ref) => {
    const resolvedContainer = useResolvedPortalContainer(container);
    return <AlertDialogPortal container={resolvedContainer}>…
  }
);
```

Three files, ~3 lines each. No new API concepts — this is the existing provider reaching the three
overlays that were skipped.

## Compatibility

Unchanged for every existing consumer. With no provider mounted the resolver returns `undefined`,
which is Radix's `document.body` default; `container='body'` still forces body even under a
provider, and `container={null}` still inherits (ref-safe).

## Tests

12 added to `portal-container.test.tsx`, table-driven across the three overlays:

- portals into the ambient provider's in-tree boundary
- falls back to `document.body` with no provider
- honours an explicit element override
- honours `container='body'` under a provider

Each was verified to fail when its component's change is reverted. The file's comment claiming the
cases cover "all three overlays" now says six.

```
pnpm test    → 8/8 tasks, 211 test files (apollo-wind 62, apollo-react 143, apollo-core 4, ap-chat 2)
pnpm build   → 7/7 tasks
tsc --noEmit → clean
biome check  → clean on the changed files
```

## Who is waiting on this

Two dependency patches carry exactly this diff today, both hosting a React eval surface inside a
shadow root from a non-React host:

- [UiPath/flow-workbench#2956](https://github.com/UiPath/flow-workbench/pull/2956) — `patches/@uipath__apollo-wind@2.32.2.patch`
- [UiPath/StudioWeb#20491](https://github.com/UiPath/StudioWeb/pull/20491) — `patches/@uipath%2Fapollo-wind@2.31.1.patch`

Both get deleted once this ships. The patch applied unchanged from 2.31.1 to 2.32.2, so the surface
has been stable across releases.

## Incidental

biome's organize-imports fixed a pre-existing violation in `alert-dialog.tsx` (its Radix import sat
below the local ones). `main` currently fails `biome check` on that file; I did not want to leave a
file I touched non-compliant.

## Separately, worth a look

`dist/tailwind.css` imports `tailwindcss`, apollo-core tokens, sonner and this package's utilities —
but never registers the animate plugin. So `animate-in` / `fade-in-0` / `zoom-in-95` cannot be
generated by any consumer compiling from that entry, while `AlertDialogContent`'s own className is
full of them. Not touched here; happy to file it separately if it is news.

### DIFF
--- a/packages/apollo-wind/src/components/ui/alert-dialog.tsx
+++ b/packages/apollo-wind/src/components/ui/alert-dialog.tsx
@@ -1,8 +1,11 @@
-import * as React from 'react';
-
-import { buttonVariants } from '@/components/ui/button';
-import { cn } from '@/lib';
 import * as AlertDialogPrimitive from '@radix-ui/react-alert-dialog';
+import * as React from 'react';
+import { buttonVariants } from '@/components/ui/button';
+import {
+  type PortalContainerOverride,
+  useResolvedPortalContainer,
+} from '@/components/ui/portal-container';
+import { cn } from '@/lib';
 
 const AlertDialog = AlertDialogPrimitive.Root;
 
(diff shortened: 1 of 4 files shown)


packages/apollo-wind/src/components/ui/alert-dialog.tsx
```diff
@@ -1,8 +1,11 @@
+import * as AlertDialogPrimitive from '@radix-ui/react-alert-dialog';
```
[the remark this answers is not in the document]
rares-baesu-uipath: Correct — fixed in 42d3978.

Every other Radix wrapper here carries the directive (`dialog`, `sheet`, `popover`, `select`, `dropdown-menu`), and so does `portal-container`, which this file now imports; importing a client module from a file without the directive is what breaks the RSC consumer. It was missing on `main` too — the new import is what makes it matter.

packages/apollo-wind/src/components/ui/alert-dialog.tsx
```diff
@@ -1,8 +1,13 @@
-import * as React from 'react';
+'use client';
```
CalinaCristian, line 1: why?

CalinaCristian (approved):

packages/apollo-wind/src/components/ui/alert-dialog.tsx
rares-baesu-uipath: i just saw it as an existing pattern in files such as `packages/apollo-wind/src/components/ui/dialog.tsx` . Also received a comment about it (related to nextjs consumers).

Will remove if necessary

PR #1040, opened 2026-08-11
feat(apollo-react): add CanvasBottomPanel to Storybook

dbacomputer:
## Summary

- add the reusable `CanvasBottomPanel` component and public exports to `@uipath/apollo-react`
- add collapsed, expanded, canvas-composition, and Flow Workbench consumer-contract stories to Storybook
- support grouped tabs, persistent shared overlay content, contextual header actions, docked/floating treatments, and configurable collapse behavior
- preserve mounted tab content and accessible keyboard/tab relationships

Screenshot 2026-08-13 at 3 14 28 PM

## Why

Flow Workbench, Studio, and VSIX need a shared bottom-panel primitive that matches their existing execution and evaluation workflows. The earlier implementation in #980 was closed without merging, so this PR rebases that work onto current `main` and includes the consumer requirements identified during review.

## Consumer impact

Consumers can use a controlled panel without duplicating expensive evaluation views or losing in-flight state when switching tabs. Existing hosts can provide either setter-style collapse handling or an expand-only callback and can select a floating canvas or docked application treatment.

## Validation

- `pnpm --filter @uipath/apollo-react exec vitest --run src/canvas/components/CanvasBottomPanel/CanvasBottomPanel.test.tsx` — 15 tests passed
- Biome lint on all five component files
- `pnpm exec tsc --noEmit -p packages/apollo-react/tsconfig.json`
- `pnpm --dir apps/storybook run storybook:build`


packages/apollo-react/src/canvas/components/CanvasBottomPanel/CanvasBottomPanel.tsx
```diff
@@ -0,0 +1,251 @@
...
+              </ActionsScope>
+            </div>
+          );
+        })}
+        {overlay && (
+          <div
+            ref={setOverlayHost}
+            id={`${idPrefix}-overlay-tabpanel`}
+            role="tabpanel"
+            aria-labelledby={overlayActive ? `${idPrefix}-tab-${activeTabId}` : undefined}
```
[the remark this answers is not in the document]
dbacomputer: Addressed in 3e65f0ea. The persistent overlay tabpanel now remains labelled while hidden by falling back to the first valid overlay tab, and switches to the active overlay tab when shown. Added regression coverage for the hidden state; all 16 focused tests, lint, and TypeScript checks pass.

packages/apollo-react/src/canvas/components/CanvasBottomPanel/CanvasBottomPanel.tsx
```diff
@@ -0,0 +1,254 @@
...
+  const [actionsContainer, setActionsContainer] = useState<HTMLDivElement | null>(null);
+  const [overlayHost, setOverlayHost] = useState<HTMLDivElement | null>(null);
+  const overlayActive = overlay?.tabIds.includes(activeTabId) ?? false;
+  const overlayLabelTabId = overlayActive
+    ? activeTabId
+    : tabs.find((tab) => overlay?.tabIds.includes(tab.id))?.id;
+  const focusTab = useCallback(
+    (tabId: string) => {
+      requestAnimationFrame(() => {
+        document.getElementById(`${idPrefix}-tab-${tabId}`)?.focus();
```
CalinaCristian: is this really necessary? I usually prefer avoiding `document` queries

packages/apollo-react/src/canvas/components/CanvasBottomPanel/CanvasBottomPanel.tsx
```diff
@@ -0,0 +1,254 @@
...
+                    className="mx-1.5 h-4 w-px shrink-0 bg-border-subtle"
+                    data-testid="canvas-bottom-panel-tab-separator"
+                  />
+                )}
+                <button
+                  id={`${idPrefix}-tab-${tab.id}`}
+                  type="button"
+                  role="tab"
+                  aria-label={tab.ariaLabel}
+                  aria-selected={isActive}
```
CalinaCristian: ### 🔴 C1 — Collapsed state still reports a selected tab

Could we factor `isCollapsed` into the selected state here? Right now the active tab keeps `aria-selected="true"` while the content container is `display: none`. Flow Workbench explicitly clears selection in this state for AE-2036 ([current contract](https://github.com/UiPath/flow-workbench/blob/e1dd3c1a0b740927978e24b39fac9e6aab96263f/packages/canvas/src/components/bottom-panel/BottomPanel.test.tsx#L97-L105)), so replacing the component would regress both that behavior and the tab/tabpanel accessibility relationship.

**Fix:** derive selection as `!isCollapsed && tab.id === activeTabId`, while keeping the active tab as the roving keyboard tab stop so the collapsed strip remains reachable. Please add the collapsed-state assertion to this suite as well.

packages/apollo-react/src/canvas/components/CanvasBottomPanel/CanvasBottomPanel.types.ts
```diff
@@ -0,0 +1,46 @@
...
+  tabs: CanvasBottomPanelTab[];
+  activeTabId: string;
+  onTabChange: (tabId: string) => void;
+  isCollapsed?: boolean;
+  /** Called with false when a collapsed panel is expanded by selecting a tab. */
+  onCollapsedChange?: (isCollapsed: boolean) => void;
+  /** Convenience callback for hosts that expose an expand action rather than a setter. */
+  onExpand?: () => void;
+  /** Consumer-owned controls rendered at the end of the panel header. */
+  headerActions?: ReactNode;
```
CalinaCristian, line 38: ### 🟠 H1 — The Flow consumers need a trailing-tab slot

Could we carry over a `tabsTrailingSlot`-style prop? Both current consumers use it for the evaluation `Preview` badge ([Studio](https://github.com/UiPath/flow-workbench/blob/e1dd3c1a0b740927978e24b39fac9e6aab96263f/packages/studio/src/components/CanvasExecutionPanel.tsx#L543-L545), [VSIX](https://github.com/UiPath/flow-workbench/blob/e1dd3c1a0b740927978e24b39fac9e6aab96263f/packages/vsix/src/webview/components/flow/components/FlowEditor.tsx#L2479-L2483)). `headerActions` is not equivalent because it moves the badge into the right-hand control group. As written, a direct migration fails type-checking, while a spread silently drops the badge.

**Fix:** add `tabsTrailingSlot?: ReactNode`, render it beside—but outside—the `role="tablist"` element, and exercise it in the Flow consumer story/test.

packages/apollo-react/src/canvas/components/CanvasBottomPanel/CanvasBottomPanel.tsx
```diff
@@ -0,0 +1,254 @@
...
+  activeTabId,
+  onTabChange,
+  isCollapsed,
+  onCollapsedChange,
+  onExpand,
+  headerActions,
+  overlay,
+  variant = 'floating',
+  collapsedHeight = CANVAS_BOTTOM_PANEL_COLLAPSED_HEIGHT,
+  idPrefix = 'canvas-bottom-panel',
```
CalinaCristian: ### 🟠 H2 — Default instances generate duplicate IDs

Could the default prefix come from React `useId()` instead of a constant? Two panels rendered without an explicit `idPrefix` produce identical tab and tabpanel IDs. Because keyboard navigation later calls `document.getElementById`, arrowing in the second panel can focus the first panel's tab; the ARIA relationships are duplicated too. The override is useful, but consumers should not need to know that it is required for a safe default.

**Fix:** generate a stable per-instance prefix with `useId()` when `idPrefix` is absent, preserve the explicit override, and add a two-instance regression test.

packages/apollo-react/src/canvas/components/CanvasBottomPanel/CanvasBottomPanel.tsx
```diff
@@ -0,0 +1,254 @@
...
+    },
+    [idPrefix]
+  );
+
+  const activateTab = useCallback(
+    (tabId: string) => {
+      clearNativeSelection();
+      onTabChange(tabId);
+      if (isCollapsed) {
+        onCollapsedChange?.(false);
```
CalinaCristian: ### 🟡 M1 — Both alternative expansion APIs fire

The props describe `onExpand` as the alternative for hosts that do not expose a setter, but providing both currently runs both side effects. A host bridging controlled state and an imperative resizable panel can therefore expand twice.

**Fix:** prefer `onCollapsedChange(false)` when it exists; otherwise call `onExpand()`. A small test with both callbacks would make the precedence explicit.

CalinaCristian (commented):
I checked this against the current Flow Workbench implementations in Studio, VSIX, and the Workbench shell. The shared primitive is close, but there are four gaps to address before it can safely replace the existing `@flow/canvas` panel: C1 (collapsed tab semantics), H1 (missing trailing-tab slot), H2 (duplicate default IDs), and M1 (both expansion callbacks fire). See the inline comments for details.

Compatibility-wise, after C1/H1/H2 this can replace `BottomPanel`, `BottomPanelActions`, and their types in Studio and VSIX. Those consumers will still need `variant="docked"`, `className="h-full"`, and `collapsedHeight={36}`. It is not a wholesale replacement for the Workbench shell's `PanelContent`; the view registry, resizer/maximize state, conditional evaluation header, and active-only mounting remain host responsibilities.

Process note: `Validate Commit Messages` is currently failing on the non-conventional merge commit message.

What is working well: the controlled state model, persistent overlay, contextual action portal, grouped tabs, and keyboard navigation all match the important shared behavior, and the focused component coverage is useful.

packages/apollo-react/src/canvas/components/CanvasBottomPanel/CanvasBottomPanel.tsx
dbacomputer: Addressed in f15ed430. When collapsed, every tab now exposes aria-selected=false while the active tab retains tabIndex=0 for the roving focus contract. Added regression coverage.

packages/apollo-react/src/canvas/components/CanvasBottomPanel/CanvasBottomPanel.tsx
dbacomputer: Addressed in f15ed430. Expansion now prefers onCollapsedChange(false) and falls back to onExpand only when the setter is absent. Added callback-precedence coverage.

CalinaCristian (approved):

PR #1052, opened 2026-08-13
feat(apollo-react): enforce read-only BaseCanvas nodes

david-rios-uipath:
## Summary

Applies per-node read-only state to generic `BaseCanvas` and `BaseNode` behavior.

## Scope

- Marks locked React Flow nodes as undeletable
- Filters direct locked-node removal requests as a defensive backstop
- Preserves selection, dragging, and resize/layout changes
- Hides BaseNode add buttons and toolbars while locked
- Prevents label editing and discards an in-progress label draft when a lock activates
- Leaves unlocked nodes unchanged by identity

## Stack

Part 2 of 4. Depends on #1051. The next PR adapts Apollo's specialized node renderers and higher-level canvas wrappers.

## Testing

- 7 focused test files
- 119 tests passing

### DIFF
--- a/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.hooks.ts
+++ b/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.hooks.ts
@@ -1,6 +1,6 @@
-import type { Node } from '@uipath/apollo-react/canvas/xyflow/react';
+import type { Edge, Node, OnBeforeDelete } from '@uipath/apollo-react/canvas/xyflow/react';
 import { useReactFlow, useStoreApi } from '@uipath/apollo-react/canvas/xyflow/react';
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { BASE_CANVAS_DEFAULTS, FIT_VIEW_DELAY_MS } from './BaseCanvas.constants';
 import type { BaseCanvasFitViewOptions, EnsureNodesInViewOptions } from './BaseCanvas.types';
 
@@ -340,3 +340,54 @@ export const useMaintainNodesInView = (
     };
   }, [flowStoreApi, reactFlowInstance, fitViewOptions]); // Only depend on flowStoreApi and reactFlowInstance
 };
+
+/**
+ * Composes an `onBeforeDelete` that vetoes deletion of locked nodes.
+ *
+ * Chosen over stamping `deletable: false` onto the nodes handed to React Flow.
+ * That clone round-trips out through `getNodes()` and `toObject()`, so a
+ * consumer saving the graph persists our lock as their own data, and unlocking
+ * the node could never undo it. A veto touches nothing the consumer owns and
+ * covers keyboard delete, `deleteElements`, and toolbar actions in one place.
+ *
+ * Edges need the same care: React Flow gathers the edges of every node in the
+ * delete set, so vetoing a node without pruning its edges would leave it
+ * standing with its connections gone. An edge is dropped when it touches a
+ * vetoed node and no node that is actually being deleted, which keeps
+ * mixed selections from stranding a dangling edge.
+ */
+export function useReadOnlyBeforeDelete<NodeType extends Node, EdgeType extends Edge>(
+  readOnlyNodeIds: ReadonlySet<string>,
+  onBeforeDelete: OnBeforeDelete<NodeType, EdgeType> | undefined
+): OnBeforeDelete<NodeType, EdgeType> | undefined {
+  return useMemo(() => {
+    if (readOnlyNodeIds.size === 0) return onBeforeDelete;
+
+    return async ({ nodes, edges }) => {
+      // The consumer runs first, so our policy is applied to whatever they
+      // allowed rather than being overridden by it.
+      const upstream = onBeforeDelete ? await onBeforeDelete({ nodes, edges }) : true;
+      if (upstream === false) return false;
+      const candidate = upstream === true ? { nodes, edges } : upstream;
+
+      const vetoedNodeIds = new Set<string>();
+      const deletedNodeIds = new Set<string>();
+      for (const node of candidate.nodes) {
+        (readOnlyNodeIds.has(node.id) ? vetoedNodeIds : deletedNodeIds).add(node.id);
+      }
+      if (vetoedNodeIds.size === 0) return candidate;
+
+      const touchesVetoedNode = (edge: EdgeType) =>
+        vetoedNodeIds.has(edge.source) || vetoedNodeIds.has(edge.target);
+      const touchesDeletedNode = (edge: EdgeType) =>
+        deletedNodeIds.has(edge.source) || deletedNodeIds.has(edge.target);
+
+      return {
+        nodes: candidate.nodes.filter((node) => !vetoedNodeIds.has(node.id)),
+        edges: candidate.edges.filter(
+          (edge) => !touchesVetoedNode(edge) || touchesDeletedNode(edge)
+        ),
+      };
+    };
+  }, [readOnlyNodeIds, onBeforeDelete]);
+}
--- a/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
+++ b/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
@@ -17,7 +17,12 @@ import {
 } from 'react';
 import { useToolbarActionStore } from '../../hooks/ToolbarActionContext';
 import { BASE_CANVAS_DEFAULTS, PAN_ON_DRAG } from './BaseCanvas.constants';
-import { useAutoLayout, useEnsureNodesInView, useMaintainNodesInView } from './BaseCanvas.hooks';
+import {
+  useAutoLayout,
+  useEnsureNodesInView,
+  useMaintainNodesInView,
+  useReadOnlyBeforeDelete,
+} from './BaseCanvas.hooks';
 import type { BaseCanvasProps, BaseCanvasRef } from './BaseCanvas.types';
 import { CanvasBackground } from './CanvasBackground';
 import { CanvasProviders } from './CanvasProviders';
@@ -77,6 +82,7 @@ const BaseCanvasInnerComponent = <NodeType extends Node = Node, EdgeType extends
     onPaneClick,
     onInit,
     onSelectionChange,
+    onBeforeDelete,
 
     // Additional ReactFlow props
     proOptions = BASE_CANVAS_DEFAULTS.pro,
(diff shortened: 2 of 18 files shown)


packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.hooks.ts
```diff
@@ -340,3 +340,20 @@ export const useMaintainNodesInView = (
...
+export function useReadOnlyEnforcedNodes<NodeType extends Node>(
+  nodes: NodeType[],
+  readOnlyNodeIds: ReadonlySet<string>
+): NodeType[] {
+  return useMemo(() => {
+    if (readOnlyNodeIds.size === 0) return nodes;
+    return nodes.map((node) =>
+      readOnlyNodeIds.has(node.id) ? { ...node, deletable: false } : node
+    );
+  }, [nodes, readOnlyNodeIds]);
```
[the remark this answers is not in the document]
david-rios-uipath: Fixed in 4575eb76. useReadOnlyEnforcedNodes now copies the array only when a locked node still needs deletable: false. Unmatched lock ids return the original array, already-undeletable locked nodes retain object identity, and mixed inputs clone only nodes that need enforcement. Added regression coverage for all three cases; the 7 focused test files pass (115 tests), along with strict type-checking and Biome checks.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
```diff
@@ -680,14 +690,15 @@ const BaseNodeComponent = (props: NodeProps<Node<BaseNodeData>>) => {
...
+          readonly={!canEdit}
+          discardDraftOnReadonly={isNodeReadOnly}
           onChange={handleLabelChange}
         />
         {displayFooter && (
           <div className="basis-full pt-0.5 min-w-0 overflow-hidden">{displayFooter}</div>
         )}
       </BaseContainer>
-      {toolbarConfig && (
+      {toolbarConfig && !isNodeReadOnly && (
```
akraminakib: How come we're hiding the entire toolbar for the read-only node, instead of just disabling all items?

packages/apollo-react/src/canvas/components/BaseNode/NodeLabel.tsx
```diff
@@ -273,20 +276,25 @@ const NodeLabelInternal = ({
...
+  // mutate data as a side effect. Every other exit commits.
+  useEffect(() => {
+    if (!isEditing) return;
+    if (readonly && discardDraftOnReadonly) {
+      handleCancel();
+    } else if (!selected || dragging || readonly) {
+      handleSave();
+    }
+  }, [selected, dragging, isEditing, readonly, discardDraftOnReadonly, handleSave, handleCancel]);
+
```
akraminakib, line 301: `readonly` drops `onDoubleClick` but the placeholder is still a real `<button aria-label="Add node label">`, and `opacity-0` doesn't take it out of the tab order. So on a locked node you can tab to a button that announces "Add node label" and pressing it does nothing.

Already true for canvas readonly mode, so not something you introduced, but this widens it to per-node locks. Mind adding `disabled` while you're here?

```tsx
<EmptyLabelPlaceholder
  disabled={readonly}
  onDoubleClick={readonly ? undefined : handleDoubleClick(labelInputRef)}
/>
```

packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
```diff
@@ -112,6 +117,20 @@ const BaseCanvasInnerComponent = <NodeType extends Node = Node, EdgeType extends
   const isInteractive = mode !== 'readonly';
   const isDesignMode = mode === 'design';
   const stableReadOnlyNodeIds = useStableNodeIdSet(readOnlyNodeIds);
+  const readOnlyEnforcedNodes = useReadOnlyEnforcedNodes(nodes, stableReadOnlyNodeIds);
+
+  // React Flow honors `deletable: false`; this guard also protects consumers
+  // that invoke the change handler directly. Layout changes pass through.
+  const handleNodesChange = useMemo(() => {
```
akraminakib, line 126: Think this guard needs to cover `replace` as well as `remove`. The nodes we pass to `<ReactFlow>` are our own clones, so React Flow's store holds the version with `deletable: false` on it. Any host calling `updateNodeData` on a locked node then gets a `replace` change built from that clone, and it sails straight through the filter into their state.

Checked it with real React Flow rather than our mock in the tests:

```
REPLACE_CHANGES: [{"id":"a","item":{...,"data":{"v":99},"deletable":false},"type":"replace"}]
CONSUMER_NODE_A:  {"id":"a",...,"data":{"v":99},"deletable":false}
```

Node went in with no `deletable` key and came out with one. The bit that worries me is that it sticks: `useReadOnlyEnforcedNodes` skips anything already at `deletable === false`, so taking the id back out of `readOnlyNodeIds` won't undo it, and anyone saving their node list persists the flag.

Would it work to remember which ids we flagged and strip that key off `change.item` on the way through? Also worth one test that skips the global mock in `canvas-mocks.ts`, since mocking the whole xyflow module is why our suite is blind to this.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
david-rios-uipath: hm that's a good point. disabling all items would be an improvement here. will look into it

packages/apollo-react/src/canvas/components/BaseNode/NodeLabel.tsx
david-rios-uipath: Added `disabled={readonly}` on the placeholder, plus a test asserting it. Locked nodes no longer expose the tab stop. Committed as a fixup, so it lands on the next stack push.

Two things it doesn't cover, both pre-existing:

* The placeholder's only handler is `onDoubleClick`. There's no `onClick` and no `onKeyDown`, so Enter/Space is a no-op on it *even on an editable node*. `disabled={readonly}` shrinks the dead-button surface rather than removing it.
* `:hover` still matches on disabled buttons, so `hover:opacity-100` keeps revealing the dashed "add label" box on locked nodes even though it's now inert.

Both point at wiring real keyboard activation on the placeholder, which feels like its own a11y change rather than something to grow a read-only enforcement PR into. Happy to file a follow-up if you agree.

packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.tsx
david-rios-uipath: Confirmed, and fixed. I reproduced it independently before changing anything: node authored with no `deletable` key, `updateNodeData` on the locked node, real React Flow.

```
{ "id": "a", "type": "replace",
  "item": { "id": "a", "data": { "v": 99 }, "position": {...},
            "deletable": false } }
```

The path, for the record: `updateNode` does `{ ...node, data }` over the *store* node, so the spread picks up our flag, and `getElementsDiffChanges` emits that object as `change.item`. The guard only special-cased `remove`, so it went straight through.

Your point about it sticking is the part that made this worth fixing properly rather than just adding `'replace'` to the filter. Once `deletable: false` is in consumer state we can no longer tell it apart from one they set deliberately.

What I did, slightly more than "strip the key":

* `useReadOnlyEnforcedNodes` now returns `{ nodes, originalDeletable }`, where `originalDeletable` maps each id it flagged to the value it replaced. Nodes already at `deletable === false` are skipped and stay out of the map, so a consumer's own flag is never mistaken for ours.
* The guard restores that original value on `replace` instead of unconditionally deleting the key. Absent originally means absent again; an explicit `deletable: true` comes back as `true` rather than silently losing the key. React Flow defaults to deletable so the behaviour is the same either way, but silently rewriting consumer data is the bug class we're fixing here, so it seemed wrong to introduce a smaller version of it.

On the mock: you're right that it's why the suite was blind. `vi.unmock` in a single file does work, so there's a new `BaseCanvas.readonlyLeak.test.tsx` running against real React Flow, covering the leak, the explicit-`true` case, the consumer-owned-`false` case, unlocked passthrough, and that deletion is still blocked. I verified the two leak tests fail without the guard. I also switched `BaseCanvas.readonlyChanges.test.tsx` to keep the real enforcement hook, so the guard is tested against the flags the hook actually sets rather than a stub.

Unrelated, but found on the way: `AgentFlow.test.tsx` was mocking `BaseCanvas.hooks` without `useReadOnlyEnforcedNodes`, so all 26 tests in it were failing on this branch. It's now a partial mock via `importActual`. Full canvas suite is 2061 passing.

Ripping out the global xyflow mock in `setup.ts` is the real root fix, but that's a much wider change than this PR should carry.

packages/apollo-react/src/canvas/components/BaseNode/BaseNode.tsx
david-rios-uipath: toolbar actions are now shown but disabled.

david-rios-uipath: Done in `3ef64af5`. The toolbar now stays mounted for locked nodes with every action (primary and overflow) force-disabled, instead of being unmounted.

One gotcha worth flagging: naively setting `disabled: true` would have been *worse* than hiding. `ToolbarButton` sets the native `disabled` attribute, and a disabled button emits no pointer events, so `CanvasTooltip`'s trigger never fires. You'd get a row of unlabelled grey icons, exactly when the label matters most. Disabled actions now render inside a hoverable wrapper so the tooltip survives. `EdgeToolbar` shares that component, so its disabled actions get their labels back too.

Disabling is also real enforcement here, not just cosmetic: `deletable: false` only stops React Flow's own delete path, while a consumer's `remove` action calls their handler directly. Both `ToolbarButton` and the overflow items guard `onClick` on `disabled`.

Two deliberate calls:

- **Add buttons and handle notches are still hidden**, not disabled. The toolbar only appears on hover or selection, so a disabled state is a response to user intent; add buttons sit inline in the layout, and a grey `+` on every handle of a locked node is just clutter.
- **The overflow `⋮` trigger stays enabled** so you can still see what the node would offer. Every item inside it is disabled.

Known limitation: `BaseNode` can't tell which actions mutate, so non-mutating ones get disabled too. `ResourceNode`, for example, ships `go-to-source` and `toggle-breakpoint` alongside `remove` and `add-guardrail`. Hiding had the same effect, so this isn't a regression. If it becomes a problem, an opt-out flag on `ToolbarActionItem` (something like `allowWhenReadOnly`) is the non-breaking fix. Left out for now as YAGNI.

Tests and docs updated: the BaseNode read-only test now asserts the disabled config instead of an absent toolbar, `NodeToolbar` gained guards for the hover wrapper, and the story spec table splits "Toolbar / Disabled" from "Add buttons / Hidden". Full canvas suite green (2122 tests).

akraminakib (approved):

akraminakib:
Small one on the new leak test, and it is only a one-word fix.

`BaseCanvas.readonlyLeak.test.tsx:29` passes `readOnlyNodeIds={["locked"]}`, but the prop is `ReadonlySet<string>`. Nothing flags it: `tsconfig.json` excludes `**/*.test.tsx` and vitest does not typecheck, so it slips past both.

It passes by accident. `useStableNodeIdSet` calls `new Set(next)` when contents differ, which coerces the array fine. The catch is `setsEqual`, which opens with `a.size !== b.size`, and an array has no `.size`. So the comparison is `1 !== undefined` on every render, the stabilizer never reaches its reuse path, and it hands back a brand new Set identity each time. That is the exact memoization the hook exists to provide, so the test guarding your leak fix is not exercising the real code path.

`new Set(["locked"])` sorts it.

Might also be worth normalizing the input in `useStableNodeIdSet`. A plain JS consumer passing an array would hit the same silent per-render churn with nothing to warn them, and taking `Iterable<string>` then wrapping in `new Set()` up front would make it impossible to get wrong.

Holychung (approved):

Holychung (approved):

PR #1053, opened 2026-08-13
feat(apollo-react): lock specialized canvas node controls

david-rios-uipath:
## Summary

Applies per-node read-only behavior to Apollo's specialized node renderers and forwards lock state through higher-level canvas consumers.

## Demo

https://github.com/user-attachments/assets/a8ddd7da-7d00-4e36-80ce-2d5baa493ece

## Scope

- `AgentNode`: hides resource/instruction editing controls
- `LoopNode`: hides structural controls while preserving resize behavior
- `StageNode`: disables task/rule mutations and closes add/replace toolboxes when locking activates
- `StickyNoteNode`: disables editing, deferred formatting commits, and resize controls
- `AgentFlow` and `HierarchicalCanvas`: forward `readOnlyNodeIds` to `BaseCanvas`

## Stack

Part 3 of 4. Depends on #1052. The final PR freezes edges wholly inside a locked region and adds the complete Storybook demo.

## Testing

- 6 focused renderer test files
- 232 tests passing

### DIFF
--- a/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.stories.independence.test.ts
+++ b/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.stories.independence.test.ts
@@ -1,0 +1,98 @@
+import { composeStories } from '@storybook/react';
+import { render, screen } from '@testing-library/react';
+import { type ComponentProps, createElement } from 'react';
+import { describe, expect, it, vi } from 'vitest';
+import * as stories from './BaseCanvas.stories';
+
+const { specializedNodeStores } = vi.hoisted(() => ({
+  specializedNodeStores: new Map<number, { getState: () => unknown }>(),
+}));
+
+// Cancels the global xyflow stub from `src/test/canvas-mocks.ts` (loaded for every
+// file by `test/setup.ts`): these stories need the real store, not the stub.
+vi.mock('@uipath/apollo-react/canvas/xyflow/react', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('@uipath/apollo-react/canvas/xyflow/react')>()),
+}));
+
+vi.mock('../AgentCanvas/nodes/AgentNode', async () => {
+  const [{ useNodeTypeRegistry }, { createElement }] = await Promise.all([
+    import('../../core'),
+    import('react'),
+  ]);
+
+  return {
+    AgentNodeElement: () => {
+      const nodeTypeRegistry = useNodeTypeRegistry();
+      const hasAgentManifest =
+        nodeTypeRegistry.getManifest('uipath.agent.autonomous') !== undefined;
+
+      return createElement(
+        'span',
+        undefined,
+        hasAgentManifest ? 'Agent manifest available' : 'Manifest Undefined'
+      );
+    },
+  };
+});
+
+vi.mock('./BaseCanvas', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('./BaseCanvas')>();
+  const [{ useStoreApi }, { createElement }] = await Promise.all([
+    import('@uipath/apollo-react/canvas/xyflow/react'),
+    import('react'),
+  ]);
+
+  return {
+    ...actual,
+    BaseCanvas: (props: ComponentProps<typeof actual.BaseCanvas>) => {
+      const store = useStoreApi();
+
+      if (props.nodes?.[0]?.type === 'comparison-agent') {
+        specializedNodeStores.set(props.readOnlyNodeIds?.size ?? 0, store);
+
+        const node = props.nodes[0];
+        const NodeComponent = props.nodeTypes?.[node.type ?? ''];
+        if (NodeComponent) {
+          return createElement(NodeComponent, {
+            id: node.id,
+            type: node.type,
+            data: node.data,
+            selected: node.selected ?? false,
+            dragging: node.dragging ?? false,
+            draggable: node.draggable ?? true,
+            selectable: node.selectable ?? true,
+            deletable: node.deletable ?? true,
+            isConnectable: node.connectable ?? true,
+            positionAbsoluteX: node.position.x,
+            positionAbsoluteY: node.position.y,
+            zIndex: node.zIndex ?? 0,
+          });
+        }
+      }
+
+      return createElement('div');
+    },
+  };
+});
+
+const { PerNodeReadOnly } = composeStories(stories);
+
+describe('specialized node comparison previews', () => {
+  it('keeps locked and unlocked previews independent when a node tab renders', () => {
+    specializedNodeStores.clear();
+
+    render(createElement(PerNodeReadOnly));
+
+    expect([...specializedNodeStores.keys()].sort()).toEqual([0, 1]);
+    expect(new Set([...specializedNodeStores.values()].map((store) => store.getState)).size).toBe(
+      2
+    );
+  });
+
+  it('provides the agent manifest when AgentNode comparisons render', () => {
+    render(createElement(PerNodeReadOnly));
+
+    expect(screen.getAllByText('Agent manifest available')).toHaveLength(2);
+    expect(screen.queryByText('Manifest Undefined')).not.toBeInTheDocument();
+  });
+});
--- a/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.types.ts
+++ b/packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.types.ts
@@ -77,10 +77,14 @@ export interface BaseCanvasProps<NodeType extends Node = Node, EdgeType extends 
 
   /**
    * Node ids whose content is read-only while the rest of the canvas remains editable.
-   * Locked nodes cannot be deleted or label-edited and hide BaseNode add buttons.
-   * Their toolbar stays visible with every action disabled, so the lock is
-   * discoverable. They remain selectable, draggable, and resizable.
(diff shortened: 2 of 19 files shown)


packages/apollo-react/src/canvas/components/LoopNode/LoopNode.tsx
```diff
@@ -381,8 +384,8 @@ function LoopNodeComponent(props: LoopNodeProps) {
         statusBorder,
         isHovered && 'shadow-(--canvas-node-shadow-hover)',
         isHovered && !hasStatusBorder && 'border-border-hover',
-        selected && 'outline outline-2 outline-foreground-accent-muted',
-        isDropTarget && 'bg-surface-hover outline outline-2 outline-brand',
+        selected && 'outline-2 outline-foreground-accent-muted',
+        isDropTarget && 'bg-surface-hover outline-2 outline-brand',
```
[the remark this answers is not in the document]
david-rios-uipath: `outline-2` contains the styles that `outline` does.

packages/apollo-react/src/canvas/components/LoopNode/LoopNode.tsx
```diff
@@ -434,7 +437,7 @@ function LoopNodeComponent(props: LoopNodeProps) {
           <EmptyState label={addNodeToLoopLabel} onAddFirstChild={handleEmptyClick} />
         </div>
       ) : null}
-      {toolbarConfig && (
+      {toolbarConfig && !isNodeReadOnly && (
```
Holychung: This still hides the whole toolbar on a locked loop, but the prop doc in this PR now says the opposite: "Toolbars are the exception: they stay visible with every action disabled, so the lock is discoverable." BaseNode does that through disableToolbarActions, LoopNode doesn't.

StickyNoteNode:621 is in the same spot and the doc lists it as honoring the contract too, but that one predates the PR. This line is new.

packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.types.ts
```diff
@@ -77,10 +77,14 @@ export interface BaseCanvasProps<NodeType extends Node = Node, EdgeType extends
...
   /**
    * Node ids whose content is read-only while the rest of the canvas remains editable.
-   * Locked nodes cannot be deleted or label-edited and hide BaseNode add buttons.
-   * Their toolbar stays visible with every action disabled, so the lock is
-   * discoverable. They remain selectable, draggable, and resizable.
-   * Custom node renderers can observe their state with `useIsNodeReadOnly`.
+   * Locked nodes cannot be deleted or label-edited and hide their inline
+   * content-editing controls. Toolbars are the exception: they stay visible with
+   * every action disabled, so the lock is discoverable. Locked nodes remain
+   * selectable and draggable. Resizable nodes generally remain resizable, except
```
Holychung, line 83: LoopNode keeps resize when locked and the comment there makes the case well: size is layout, not content. StickyNote loses it, and this doc ends up calling that an exception rather than a rule.

Looks incidental rather than chosen. StickyNoteNode:624 already gated resize on readOnly, and the node lock now folds into that same flag, so the sticky note gives it up for free while the loop deliberately keeps it. Worth picking one?

packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.stories.independence.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    render(createElement(PerNodeReadOnly));
+
+    expect([...specializedNodeStores.keys()].sort()).toEqual([0, 1]);
+    expect(new Set([...specializedNodeStores.values()].map((store) => store.getState)).size).toBe(
+      2
+    );
+  });
+
+  it('uses the available page width when specialized comparisons render', () => {
```
Holychung: The store-independence test above is a good one. These two are asserting the docs page's layout though: data-story-layout, w-[calc(100vw-2rem)], max-w-[1600px], reached through parentElement.parentElement, with BaseCanvas mocked out. data-story-layout exists in exactly two places, the attribute on StoryPage and the assertion here. Feels like it breaks on the next styling tweak and tells us nothing about the product when it does.

Separately, the xyflow mock at the top of the file reads as dead code, it spreads importOriginal straight back out. Took me a minute to see it's cancelling the global mock in src/test/canvas-mocks.ts that setup.ts loads for every file. One line of comment would save the next person.

packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.types.ts
david-rios-uipath: This is intentional. I chose to preserve resize for the loop node since it doesn't affect its execution behavior.

Sticky notes don't affect execution at all. Their entire functionality is display. So if a sticky note is targeted to be read-only, it shouldn't be resizable either, since resizing is a core part of its functionality.

packages/apollo-react/src/canvas/components/LoopNode/LoopNode.tsx
david-rios-uipath: Fixed. `disableToolbarActions` moved to `Toolbar/NodeToolbar/NodeToolbar.utils.ts` as `lockToolbarConfig` (config in, config out) and LoopNode now builds an `effectiveToolbarConfig` the same way BaseNode does, so a locked loop keeps its toolbar with every action disabled. `BaseNode` uses the shared helper too, so the two cannot drift again.

StickyNoteNode is in as well, since 14f6990d folded `isNodeReadOnly` into the same `readOnly` flag that line 621 reads. The per-node lock now keeps the toolbar with disabled actions; the sticky-note-level `readOnly` prop/option predates the contract and still hides it outright, which the code comments. `handleDelete`, `handleColorChange`, and `handleToggleColorPicker` picked up read-only guards, and the color picker closes when the note locks, since the toolbar is now reachable in that state. Resize controls stay unmounted while locked, per the existing test.

Tests: 3 in `LoopNode.test.tsx` (mocking `NodeToolbar` and asserting on the config, mirroring `BaseNode.test.tsx`) and 4 in `StickyNoteNode.test.tsx` covering locked/unlocked/sticky-level-read-only toolbars and blocked resizing. The prop doc is now accurate as written, so it is unchanged.

packages/apollo-react/src/canvas/components/BaseCanvas/BaseCanvas.stories.independence.test.ts
david-rios-uipath: Both dropped. The layout test is gone and `data-story-layout` came off `StoryWideContent` with it, since the assertion was its only consumer. The store-independence and agent-manifest tests stay.

The xyflow mock has a comment now:

```ts
// Cancels the global xyflow stub from `src/test/canvas-mocks.ts` (loaded for every
// file by `test/setup.ts`): these stories need the real store, not the stub.
```

Same pattern existed twice: `BaseCanvas.stories.test.ts` on #1033 had its own copy of the layout assertion and the uncommented mock. Fixed there too, so the attribute has no remaining references anywhere.

snuziale (approved):

Holychung (approved):

packages/apollo-react/src/canvas/components/LoopNode/LoopNode.tsx
```diff
@@ -347,7 +352,6 @@ function LoopNodeComponent(props: LoopNodeProps) {
 
   if (!manifest) {
     return (
-      // biome-ignore lint/a11y/noStaticElementInteractions: canvas node hover state is mouse-driven
       <div
         className="relative"
```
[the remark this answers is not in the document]
david-rios-uipath: Checked this rather than taking it on faith: the `Lint` job is green on this PR, and `biome lint` over the changed files reports no new diagnostics. Biome does not flag these divs, so the suppressions were dead comments (it was reporting `Suppression comment has no effect` on the pair in BaseNode for the same reason). Removing them is the fix, not the regression. Resolving.

packages/apollo-react/src/canvas/components/LoopNode/LoopNode.tsx
```diff
@@ -365,7 +377,6 @@ function LoopNodeComponent(props: LoopNodeProps) {
   }
 
   return (
-    // biome-ignore lint/a11y/noStaticElementInteractions: canvas node hover state is mouse-driven
     <div
       data-loop-container
       data-selected={selected ? 'true' : 'false'}
```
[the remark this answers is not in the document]
david-rios-uipath: Checked this rather than taking it on faith: the `Lint` job is green on this PR, and `biome lint` over the changed files reports no new diagnostics. Biome does not flag these divs, so the suppressions were dead comments (it was reporting `Suppression comment has no effect` on the pair in BaseNode for the same reason). Removing them is the fix, not the regression. Resolving.

Holychung (approved):

Holychung:
shouldShowMenu dropped the !isReadOnly gate, so a locked stage now shows the host menu. Not blocking it, the stage's own mutation paths (task add/replace, entry/exit rules) and the task-level TaskMenu are all still gated.

One ask: the contract is now asymmetric, apollo gates its own affordances and host menuItems are the host's to gate, and nothing says so. StageNodeTaskUtilities builds MOVE_UP/MOVE_DOWN style items itself, so "host items are inspect/debug only" is not a safe assumption for a Case designer. Worth a line in the readOnlyNodeIds doc block.

Holychung (approved):

PR #1104, opened 2026-08-31
fix(apollo-react): add aria-label to suggestion navigator chevrons

abegu:
## Summary
- The chevron-up/chevron-down icon buttons in `SuggestionGroupNavigator` (`packages/apollo-react/src/canvas/components/AgentCanvas/components/SuggestionGroupPanel.tsx`) had no accessible name at all — their only content is a decorative Lucide icon (`chevron-up`/`chevron-down`), which contributes nothing to the accessible name.
- Screen reader users focusing these controls got no indication of what "previous suggestion" / "next suggestion" navigation they performed.
- Confirmed this is unfixed as of `main` (not just the pinned `6.20.1` a downstream consumer had) before filing.

## Changes
- Added `aria-label="Previous suggestion"` / `aria-label="Next suggestion"` to the two navigator `Button`s.
- Added `SuggestionGroupPanel.test.tsx` (new file — none existed for this component) covering: empty-state renders nothing, both buttons expose accessible names, both buttons invoke their respective callbacks, and the position counter renders correctly.

## Test plan
- [x] `vitest run` on the new test file — 4/4 passing
- [x] `biome check` — clean
- [x] `tsc --noEmit` — no new errors

## Source
Found while investigating a WCAG 2.4.6 report (PC-4063) filed against a downstream product consuming this package — the bug turned out to be upstream here, not in the consumer.

### DIFF
--- a/packages/apollo-react/src/canvas/components/AgentCanvas/components/SuggestionGroupPanel.tsx
+++ b/packages/apollo-react/src/canvas/components/AgentCanvas/components/SuggestionGroupPanel.tsx
@@ -88,6 +88,7 @@ const SuggestionGroupNavigator = ({
         variant="ghost"
         size="icon"
         className="h-8 w-8"
+        aria-label="Previous suggestion"
         onMouseEnter={() => setIsHoveringUp(true)}
         onMouseLeave={() => setIsHoveringUp(false)}
         onClick={onNavigatePrevious}
@@ -105,6 +106,7 @@ const SuggestionGroupNavigator = ({
         variant="ghost"
         size="icon"
         className="h-8 w-8"
+        aria-label="Next suggestion"
         onMouseEnter={() => setIsHoveringDown(true)}
         onMouseLeave={() => setIsHoveringDown(false)}
         onClick={onNavigateNext}
--- a/packages/apollo-react/src/canvas/components/AgentCanvas/components/SuggestionGroupPanel.test.tsx
+++ b/packages/apollo-react/src/canvas/components/AgentCanvas/components/SuggestionGroupPanel.test.tsx
@@ -1,0 +1,67 @@
+import { render, screen } from '@testing-library/react';
+import { ReactFlowProvider } from '@uipath/apollo-react/canvas/xyflow/react';
+import { describe, expect, it, vi } from 'vitest';
+import type { AgentFlowSuggestionGroup } from '../../../types';
+import { SuggestionGroupPanel } from './SuggestionGroupPanel';
+
+const suggestionGroup: AgentFlowSuggestionGroup = {
+  id: 'group-1',
+  suggestions: [
+    { id: 'suggestion-1', type: 'add' },
+    { id: 'suggestion-2', type: 'add' },
+  ],
+};
+
+const renderWithProvider = (ui: React.ReactElement) =>
+  render(<ReactFlowProvider>{ui}</ReactFlowProvider>);
+
+describe('SuggestionGroupPanel', () => {
+  it('renders nothing when there is no suggestion group', () => {
+    renderWithProvider(<SuggestionGroupPanel suggestionGroup={null} />);
+    expect(screen.queryByRole('button')).not.toBeInTheDocument();
+  });
+
+  it('exposes accessible names for the previous/next navigator buttons', () => {
+    renderWithProvider(
+      <SuggestionGroupPanel
+        suggestionGroup={suggestionGroup}
+        currentIndex={0}
+        onNavigateNext={vi.fn()}
+        onNavigatePrevious={vi.fn()}
+      />
+    );
+
+    expect(screen.getByRole('button', { name: 'Previous suggestion' })).toBeInTheDocument();
+    expect(screen.getByRole('button', { name: 'Next suggestion' })).toBeInTheDocument();
+  });
+
+  it('calls onNavigatePrevious/onNavigateNext when the navigator buttons are clicked', async () => {
+    const onNavigatePrevious = vi.fn();
+    const onNavigateNext = vi.fn();
+
+    renderWithProvider(
+      <SuggestionGroupPanel
+        suggestionGroup={suggestionGroup}
+        currentIndex={0}
+        onNavigateNext={onNavigateNext}
+        onNavigatePrevious={onNavigatePrevious}
+      />
+    );
+
+    screen.getByRole('button', { name: 'Previous suggestion' }).click();
+    expect(onNavigatePrevious).toHaveBeenCalledTimes(1);
+
+    screen.getByRole('button', { name: 'Next suggestion' }).click();
+    expect(onNavigateNext).toHaveBeenCalledTimes(1);
+  });
+
+  it('shows the current position among the suggestions', () => {
+    renderWithProvider(<SuggestionGroupPanel suggestionGroup={suggestionGroup} currentIndex={0} />);
+
+    expect(
+      screen.getByText(
+        (_, element) => element?.tagName === 'SPAN' && element.textContent === '1 of 2'
+      )
+    ).toBeInTheDocument();
+  });
+});


robert-trandafir:
Hi Andrei — I reviewed this one in depth ahead of lining up an approval. **No blocking findings**; I'd be happy to see it go in. A few notes below, tiered so it's clear what's optional.

I also traced it back to the ticket, and the mapping is exact: **PC-4063** ([2.4.6] *Autopilot result screen with prompt — Programmatic label does not convey purpose of control*, `VPAT26.10`) names the element as *"Up and Down buttons (indicating previous suggestion and next suggestion)"*, which is precisely this `SuggestionGroupNavigator`. The ticket asks for a programmatic label conveying the control's purpose, and since these are icon-only with no visible text, `aria-label` is the right technique. This fix closes it cleanly.

---

## 🟠 Worth knowing before merge

**No lint, typecheck, test or build actually ran on this PR.** `.github/workflows/pr-checks.yml:32` gates the whole `checks` matrix on `if: github.event.pull_request.head.repo.fork == false`. Everything green here is CodeQL / actionlint / Socket / zizmor / "Detect changes" — nothing executes a test. The two red checks are just the fork token lacking write access for the labeler, not a problem with your code. Entirely a consequence of the permissions situation rather than anything you did, but it does mean the `vitest` / `biome` / `tsc` results in the description are local-only and unverified by CI. If someone can push this as a branch inside the repo, CI would cover it properly.

## 🟡 Small things

- `SuggestionGroupPanel.test.tsx:44`/`:47` use raw DOM `.click()` rather than `fireEvent.click` / `userEvent`, which is the convention everywhere else in the canvas suites (e.g. `HandleButton.test.tsx`). It works here only because the callbacks are bare `vi.fn()` with no state update — the moment the component sets state on click, this becomes an unwrapped-`act` warning. Cheap to switch now.
- The fix names the controls, but the `1 of 2` counter (`SuggestionGroupPanel.tsx:102-104`) isn't a live region, so activating a chevron announces nothing to a screen reader. That's WCAG 4.1.3 rather than the 2.4.6 this PR targets, so I'd call it a genuine follow-up rather than a gap here — but `role="status"` on that span is a one-liner if you want it covered in the same change.

## 🔵 Optional

- The aria-labels are hardcoded English. Canvas does localise aria-labels elsewhere via a `translations: CanvasTranslations` prop (`CanvasPositionControls.tsx:21`, `:31`, `:56`), but this whole component is un-localised already ("Reject all" / "Accept all" are hardcoded too), so being consistent with the file seems right. Flagging only because it means non-English screen-reader users get English names.
- `jest-axe`'s `toHaveNoViolations` is already wired up in `src/test/setup.ts` and used in ~10 suites (the Guardrails ones are a good template) — an `axe(container)` assertion would be a natural fit for an a11y PR.
- Worth linking PC-4063 in the description so the fix stays traceable for the 26.10 VPAT.
- PC-4063 doesn't carry the `apollo-a11y-issue` label, even though the root cause turned out to be here in apollo-ui. That label is already in use for exactly this (e.g. PC-1783, PC-1786, PC-4558) — adding it would stop the Agent Builder side chasing it and make the upstream dependency visible in VPAT reporting.

---

## What I verified rather than assumed

- `aria-label` genuinely reaches the DOM: apollo-wind `Button` spreads `...props` onto the native `<button>` (`packages/apollo-wind/src/components/ui/button.tsx`).
- Nothing else breaks: `git grep` across the whole tree confirms no other test or story queries these buttons by name.
- The fixture typechecks — `AgentFlowSuggestion` only requires `id` and `type` (`canvas/types.ts:244`).
- `ReactFlowProvider` is exported from the aliased path (`canvas/xyflow/react.ts` re-exports `@xyflow/react`), and four other canvas suites use the same pattern.
- No cross-test DOM leakage from the four renders — `src/test/setup.ts` registers a global `cleanup()` and `vitest.config.ts` sets `globals: true`.
- The custom `getByText` matcher in the last test is a legitimate workaround for React's split text nodes, not a smell. (If you add `role="status"` per the 🟡 above, `getByRole('status')` would replace it and read better.)
- `MERGEABLE`, and `main` hasn't touched `SuggestionGroupPanel.tsx` since your fork point.

*Reviewed with Claude Code.*

abegu:
Thanks for the deep review, Robert — much appreciated.

**Fixed:** swapped the raw DOM `.click()` for `fireEvent.click()` in `SuggestionGroupPanel.test.tsx`, matching the rest of the canvas suites. Pushed in 7f759c98. Re-ran vitest/biome/tsc locally (same three checks the fork-gate skips) — 4/4 tests still pass, clean.

On the rest, agreeing with your own tiering rather than expanding scope:
- **`role="status"` on the counter** — leaving as a genuine follow-up, as you said. It's WCAG 4.1.3, this PR targets 2.4.6/PC-4063, and I'd rather not scope-creep a ticket-mapped fix.
- **Localization / `axe(container)` assertion** — agreed these are reasonable, but both are optional per your own note and consistent with the file's existing (already-unlocalized) state, so leaving them out of this PR too.
- **Linking PC-4063 in the description, and the `apollo-a11y-issue` Jira label** — skipping both here. I don't have Jira access from this session, and didn't want to touch ticket metadata without someone actually confirming it — flagging both to Andrei to handle directly rather than guessing.

Really appreciate the "what I verified rather than assumed" section — that's exactly the kind of review that's useful even without CI running the suite.

packages/apollo-react/src/canvas/components/AgentCanvas/components/SuggestionGroupPanel.tsx
kittyyueli: these aria-label should be translated~ I think we have that set up in this repo now that can be consumed, but might be too big a change for this PR

kittyyueli (approved):

packages/apollo-react/src/canvas/components/AgentCanvas/components/SuggestionGroupPanel.tsx
abegu: Fixed - and it wasn't too big a change after all. Routed both aria-labels through `useSafeLingui` (`_({ id, message })`), matching the exact pattern sibling canvas controls already use (e.g. `NodeOutputModeSelect`). It falls back to the English message when no `I18nProvider` is mounted, so this is a drop-in change. Added the two new ids to `src/canvas/locales/en.json` for the l10n team to pick up. Pushed in a8ac3e58.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abegu:
Re: Copilot's note about adding translations for the new canvas locale catalog entries (`canvas.suggestion_group_panel.previous_suggestion` / `next_suggestion`) to the non-English locale files.

Leaving this out of this PR intentionally: real translations for this repo's canvas catalog come from the l10n team/pipeline (translators return updated `de.json`, `ja.json`, etc.), not from an engineer hand-writing guesses across 13 languages. `useSafeLingui` is specifically designed to fall back to the English `message` for any id missing from a locale's active catalog, so behavior is correct in every locale in the meantime — no user sees a raw message id or broken string. Once the l10n process picks up the new `en.json` entries, the other catalogs will get real translations through the normal flow.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

abegu:
Closing in favor of #1208 — re-opened as a branch inside UiPath/apollo-ui (not a fork) so the required CI checks (Lint, Build, Typecheck, Test Coverage, Format, etc.) actually run. On a fork PR, GitHub always issues a read-only GITHUB_TOKEN to pull_request-triggered workflows regardless of the author's permissions, so those checks were silently skipped here and this PR could never satisfy branch protection — confirmed via the ruleset's required_status_checks list, none of which ever appeared in this PR's checks. Same diff, same commit history, now at #1208.

_🤖 Closed by [Claude Code](https://claude.com/claude-code)_

PR #1109, opened 2026-09-01
fix(apollo-wind): removes sr-only default text for required indicator

BenGSchulz:
Supplying default screen reader text was overkill for majority of cases. It leads to a duplicated required announcement for assistive tech when labelled fields correctly use `required`/`aria-required`. The props remains as an escape hatch for consumers to provide alternative assistive labels or label controls that do not support `required`/`aria-required`.

### DIFF
--- a/packages/apollo-wind/src/components/ui/label.test.tsx
+++ b/packages/apollo-wind/src/components/ui/label.test.tsx
@@ -106,7 +106,7 @@ describe('RequiredIndicator', () => {
     expect(indicator).not.toHaveClass('text-error');
   });
 
-  it('hides the asterisk glyph from assistive tech but announces it via sr-only text', () => {
+  it('hides the asterisk glyph from assistive tech and has no sr label by default', () => {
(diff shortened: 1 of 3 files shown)


packages/apollo-wind/src/components/ui/label.test.tsx
```diff
@@ -133,7 +133,7 @@ describe('RequiredIndicator', () => {
       <div>
         <Label htmlFor="name">
           Name
-          <RequiredIndicator />
+          <RequiredIndicator srLabel="(required)" />
```
david-rios-uipath, line 136: should we localize these?

BenGSchulz: This is just in a test file. There is also no localization support in apollo-wind at the moment. Consumers need to provide translated strings

david-rios-uipath (approved):

david-rios-uipath (approved):

PR #1112, opened 2026-09-02
fix(apollo-wind): stop FormField overflowing on truncating controls

BenGSchulz:
Two related layout fixes in `apollo-wind`, reported against 2.43.1.

## 1. `FormField`'s grid floors at min-content

`FormField` rendered `grid gap-1.5`. Its implicit column track is `auto`, whose *minimum* is the item's automatic minimum size, and grid items get `min-width: auto`. So the track could never be narrower than its widest child's min-content.

A control containing a `truncate` span is `white-space: nowrap`, which makes its min-content the **entire string**. Any such control forced the track wider than the form and overflowed its container instead of ellipsizing.

This was not limited to custom fields: `field-renderer` wraps every standard-typed field in `FormField`, so `MetadataForm` carried it too. It showed up wherever a form lives in a width-constrained surface, such as a properties panel, sidebar, or narrow dialog.

**Fix:** pin the single column to `minmax(0, 1fr)`.

### Measured, against the compiled Tailwind build

Same markup, same text, in a 320px host (304px content box):

| host | `grid` (before) | `grid-cols-[minmax(0,1fr)]` (after) |
| --- | --- | --- |
| definite 320px | control **712.8px**, no ellipsis | **304px**, ellipsizes |
| `inline-block` (shrink-to-fit) | 712.8px | 712.8px |
| flex item, `flex: 0 1 auto` | 712.8px | 712.8px |
| block, 1000px | 1000px | 1000px |

An `fr` track resolves against the max-content constraint when the container's width is indefinite, so shrink-to-fit hosts are byte-identical. The two behaviours diverge **only** when the width is definite, which is exactly the broken case, so no consumer depends on today's behaviour who isn't already looking at an overflowing control.

The escape hatch survives, since `className` is merged last:

```
cn('grid grid-cols-[minmax(0,1fr)] gap-1.5', 'grid-cols-2')
  -> 'grid gap-1.5 grid-cols-2'
```

`flex flex-col` would also have fixed this, and arguably more honestly, since the automatic-minimum-size rule applies only to the flex main axis. It was rejected because it silently breaks any consumer passing `grid-cols-2` / `grid-cols-[auto_1fr]`, as those utilities do not restore `display: grid`. One extra class is the cheaper trade.

## 2. `Label` had no display, so `space-y-*` silently did nothing

`Label` rendered a bare Radix `<label>` with typography classes and no display utility, so it was `display: inline`. Tailwind v4 changed `space-y-*` to emit `margin-block-end` on every child but the last (v3 emitted `margin-top` on every child but the first). Vertical margins have no effect on non-replaced inline boxes, so the margin landed on the `Label` and was discarded.

Confirmed in the browser: the `space-y-1.5` rule does emit `margin-block-end`, the `Label` computed to `display: inline`, `margin-block-end` computed to `6px`, and the measured gap was `3.5px` (pure line-box leading).

**Fix:** `inline-block`. It preserves inline-flow use, such as checkbox rows and inline hints, in a way `block` would not.

### ⚠️ Expected visual change

Anywhere a consumer wrapped a bare `Label` in `space-y-*`, the previously swallowed gap now appears. One instance exists in our own stories (`Basic Contact Form`, the "Urgent Request" toggle under a `space-y-0.5` wrapper).

Inside `FormField` nothing changes: grid items are blockified, so the label already computed to `block`. Verified in the DOM. The `datetime-picker` label that passes `flex items-center gap-2` stays `flex` via tailwind-merge. No consumer puts `truncate` on a `Label`.

## Verification

- Full `apollo-wind` suite: **1437 tests / 93 files passing**. Four new regression tests covering the pinned column, the consumer column override, the label display, and the label display override.
- `biome check` clean on all four files; `tsc --noEmit` clean.
- Rendered in Storybook at desktop and 375px, no breakage.
- `test:visual` is declared in `turbo.json` but no package implements it, so there is no visual suite to run. The pixel diff noted above is worth a manual look if snapshots exist elsewhere.

## Notes for the reporter

The local `@layer base` workaround can be deleted once this lands:

```css
@layer base {
  :where([data-slot='form-field']) {
    grid-template-columns: minmax(0, 1fr);
  }
}
```

It composes with this fix rather than fighting it, so there is no version gate on removing it.

### DIFF
--- a/packages/apollo-wind/src/components/ui/form-field.test.tsx
+++ b/packages/apollo-wind/src/components/ui/form-field.test.tsx
@@ -19,6 +19,18 @@ describe('FormField', () => {
     expect(screen.getByText('The URL to call.')).toBeInTheDocument();
     expect(screen.getByText('Endpoint is required.')).toBeInTheDocument();
   });
+
+  it('pins the column so a truncating control cannot widen the field', () => {
+    render(<FormField data-testid="field" />);
+    expect(screen.getByTestId('field')).toHaveClass('grid', 'grid-cols-[minmax(0,1fr)]');
+  });
+
+  it('lets a consumer override the column template', () => {
+    render(<FormField data-testid="field" className="grid-cols-2" />);
+    const field = screen.getByTestId('field');
+    expect(field).toHaveClass('grid-cols-2');
+    expect(field).not.toHaveClass('grid-cols-[minmax(0,1fr)]');
+  });
 });
 
 describe('FormFieldLabel', () => {
--- a/packages/apollo-wind/src/components/ui/form-field.tsx
+++ b/packages/apollo-wind/src/components/ui/form-field.tsx
@@ -10,10 +10,23 @@ export type FormFieldProps = React.ComponentPropsWithoutRef<'div'>;
  *
  * These parts are presentational and hold no form state, so they compose the
  * same way inside a metadata-driven form as they do in hand-built panels.
+ *
+ * The single column is pinned to `minmax(0, 1fr)` rather than left implicit.
+ * An implicit `auto` track floors at its widest child's min-content, and a
+ * control that ellipsizes is `white-space: nowrap`, so its min-content is the
+ * whole string: the field would push past a width-constrained host instead of
+ * truncating. A `1fr` track still resolves to max-content when the host width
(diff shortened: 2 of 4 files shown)


david-rios-uipath (approved):
thanks for fixing these!

PR #1116, opened 2026-09-03
feat(apollo-wind): align modal takeover and run actions

dbacomputer:
## Summary

This PR aligns Modal and takeover examples with the Flow visual language, consolidates takeover behavior on the Apollo Wind core Modal, and standardizes execution actions as `Run`.

### Apollo Wind — additions and adjustments

- Rename the Storybook component category to `Modal` and explain that it is implemented with the shadcn/ui Dialog component and naming conventions, while Apollo’s internal name is Modal.
- Add takeover support to the core Modal with sidebar and no-sidebar examples, centered layout, Flow semantic surfaces, theme-aware curtain overlay, expand/collapse controls, and consistent radius behavior.
- Add public `Modal*` aliases for the existing Dialog primitives.
- Update the Sticky Footer example so its header, body, footer, borders, background surfaces, and corner radius demonstrate the same Flow styling used by Modal.
- Add focused takeover tests for title handling, sidebar rendering, expansion, and backdrop-dismiss behavior.

### Apollo React — removals and adjustments

- Remove the obsolete `CanvasTakeoverModal` Storybook page.
- Migrate Flow Standalone takeover examples to use Apollo Wind `Modal` and `ModalContent variant="takeover"` directly.
- Retain the exported `CanvasTakeoverModal` component temporarily as a deprecated compatibility wrapper. This means existing consumers can keep importing the old component name while its implementation delegates to the new Apollo Wind Modal; it can be removed in a future breaking release after consumers migrate.
- Update execution action labels from `Debug`, `Run debug`, and `Run test` to `Run`, including the Play icon where the action is presented as a button. Debug mode and debug-result panel terminology remain unchanged.

### Storybook and layout patterns

- Update Node Property Panel, Canvas toolbar, bottom-panel action, layout-pattern, and Flow examples to use consistent `Run` labels and icon treatment.
- Fix the UX inventory draggable list markup and make its Move down boundary depend on the current list length.

## Verification

- Apollo Wind Dialog tests: 15 passed.
- CanvasTakeoverModal compatibility tests: 5 passed.
- Biome checks pass for changed files.
- `git diff --check` passes.
- Local Storybook preview: http://localhost:6007/


packages/apollo-react/src/canvas/components/CanvasTakeoverModal/CanvasTakeoverModal.tsx
```diff
@@ -1,157 +1,77 @@
...
+        sidebarClassName={sidebarClassName}
+        closeOnBackdropClick={closeOnBackdropClick}
+        overlayTestId="canvas-takeover-backdrop"
+        data-expanded={expanded ?? defaultExpanded ?? false}
+        onOpenAutoFocus={(event) => {
+          event.preventDefault();
+          (event.currentTarget as HTMLElement)
+            .querySelector<HTMLElement>('[aria-label="Expand modal"]')
+            ?.focus();
+        }}
```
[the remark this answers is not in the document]
dbacomputer: Fixed in 0c8307ca: autofocus now targets either the Expand or Collapse control, so expanded takeovers also receive focus correctly.

packages/apollo-react/src/canvas/components/CanvasTakeoverModal/CanvasTakeoverModal.tsx
```diff
@@ -1,157 +1,77 @@
...
+        overlayProps={{
+          onMouseDown: (event) => {
+            if (closeOnBackdropClick && event.target === event.currentTarget) onOpenChange?.(false);
+          },
+        }}
```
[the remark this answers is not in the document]
dbacomputer: Fixed in 0c8307ca: removed the wrapper-level mouse handler and now rely on Radix outside-pointer dismissal, with closeOnBackdropClick=false handled by onPointerDownOutside.

packages/apollo-wind/src/components/ui/dialog.tsx
```diff
@@ -51,39 +51,145 @@ const DialogOverlay = React.forwardRef<
 
 const DialogContent = React.forwardRef<
   React.ComponentRef<typeof DialogPrimitive.Content>,
-  React.ComponentPropsWithoutRef<typeof DialogPrimitive.Content> & {
+  Omit<React.ComponentPropsWithoutRef<typeof DialogPrimitive.Content>, 'title'> & {
     showCloseButton?: boolean;
     container?: PortalContainerOverride;
+    variant?: 'default' | 'takeover';
+    title?: React.ReactNode;
+    headerActions?: React.ReactNode;
```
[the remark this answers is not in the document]
dbacomputer: Fixed in 0c8307ca: renamed the takeover heading prop to headerTitle while preserving DialogContent’s native title attribute.

packages/apollo-react/src/canvas/components/NodePropertyPanel/NodePropertyPanel.stories.tsx
```diff
@@ -4388,7 +4376,7 @@ function PanelUIInventoryStory() {
                     </PatternNote>
                     <div className="flex flex-wrap items-center gap-2">
                       <RunButton />
-                      <DebugButton />
+                      <RunButton />
```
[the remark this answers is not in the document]
dbacomputer: Fixed in 0c8307ca: the header-actions pattern now demonstrates the full Run button alongside the icon-only Run control.

packages/apollo-wind/src/components/ui/dialog.tsx
```diff
@@ -51,39 +51,145 @@ const DialogOverlay = React.forwardRef<
...
+        className={cn(
+          isTakeover
+            ? 'bg-surface-raised text-foreground data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 relative z-50 flex min-h-0 w-full flex-col overflow-hidden border border-border-subtle shadow-xl duration-200'
+            : 'bg-surface-raised text-foreground data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 relative z-50 grid w-full max-w-[calc(100%-2rem)] gap-4 rounded-2xl border border-border-subtle p-6 shadow-xl duration-200 sm:max-w-lg',
+          isTakeover && (expanded ? 'h-full w-full' : 'h-[95%] w-[95%] rounded-2xl'),
+          className
+        )}
+        onPointerDownOutside={(event) => {
+          if (isTakeover && !closeOnBackdropClick) event.preventDefault();
+        }}
```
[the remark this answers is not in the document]
dbacomputer: Fixed in 0c8307ca: added focused takeover tests covering header/sidebar rendering, expansion, native title preservation, and disabled backdrop dismissal.

dbacomputer:
Follow-up cleanup completed in `9db98deb`:

- Replaced the standalone draggable `<li>` elements with a valid `<ul>/<li>` structure for accessibility.
- Made the “Move down” disabled state derive from `compositionFields.length - 1` instead of a hard-coded index.
- Biome and diff checks pass.

The five prior review threads remain resolved.

apps/storybook/src/patterns/LayoutPatterns.stories.tsx
```diff
@@ -1,3 +1,4 @@
+import MonacoEditor from '@monaco-editor/react';
 import type { Meta, StoryObj } from '@storybook/react-vite';
```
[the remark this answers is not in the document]
dbacomputer: Fixed in `5f9d0b3f`: switched the layout-pattern Monaco editor to the same lazy dynamic-import wrapper used by the Node Property Panel stories, avoiding the Rolldown CJS default-import issue.

packages/apollo-wind/src/components/ui/dialog.tsx
```diff
@@ -54,36 +54,155 @@ const DialogContent = React.forwardRef<
...
+        ref={ref}
+        {...props}
+        {...((isTakeover && (headerTitle === undefined || headerTitle === null)) || ariaLabelledBy
+          ? {
+              'aria-labelledby':
+                isTakeover && (headerTitle === undefined || headerTitle === null)
+                  ? ''
```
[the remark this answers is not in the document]
dbacomputer: Fixed in `7b0930e4`: title-less takeovers no longer force an empty aria-labelledby attribute; a consumer-provided aria-labelledby is preserved, otherwise aria-label can name the dialog.

packages/apollo-wind/src/components/ui/dialog.tsx
```diff
@@ -54,36 +54,155 @@ const DialogContent = React.forwardRef<
...
+            ? 'bg-surface-raised text-foreground data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 relative z-50 flex min-h-0 w-full flex-col overflow-hidden border border-border-subtle shadow-xl duration-200'
+            : 'bg-surface-raised text-foreground data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 relative z-50 grid w-full max-w-[calc(100%-2rem)] gap-4 rounded-2xl border border-border-subtle p-6 shadow-xl duration-200 sm:max-w-lg',
+          isTakeover && (expanded ? 'h-full w-full' : 'h-[95%] w-[95%] rounded-2xl'),
+          className
+        )}
+        onPointerDownOutside={(event) => {
+          if (isTakeover && !closeOnBackdropClick) event.preventDefault();
+        }}
+        ref={ref}
+        {...props}
```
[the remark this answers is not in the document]
dbacomputer: Fixed in `7b0930e4`: the takeover handler is now composed with any consumer-provided `onPointerDownOutside` callback, so the built-in closeOnBackdropClick policy cannot be bypassed.

packages/apollo-wind/src/components/ui/dialog.tsx
```diff
@@ -54,36 +54,155 @@ const DialogContent = React.forwardRef<
...
-              className="ring-offset-background focus:ring-ring data-[state=open]:bg-accent data-[state=open]:text-muted-foreground absolute top-4 right-4 cursor-pointer rounded-xs opacity-70 transition-opacity hover:opacity-100 focus:ring-2 focus:ring-offset-2 focus:outline-hidden disabled:cursor-default disabled:pointer-events-none [&_svg]:pointer-events-none [&_svg]:shrink-0 [&_svg:not([class*='size-'])]:size-4"
-            >
-              <XIcon />
-              <span className="sr-only">Close</span>
-            </DialogPrimitive.Close>
-          )}
-        </DialogPrimitive.Content>
-      </DialogOverlay>
+      {content}
     </DialogPortal>
```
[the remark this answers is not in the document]
dbacomputer: Fixed in `5702973c`: takeover content now consistently uses DialogPortal, honoring the configured provider/body container semantics for undefined, null, and attached ref containers.

packages/apollo-wind/src/components/ui/dialog.tsx
```diff
@@ -54,36 +54,152 @@ const DialogContent = React.forwardRef<
   React.ComponentPropsWithoutRef<typeof DialogPrimitive.Content> & {
     showCloseButton?: boolean;
     container?: PortalContainerOverride;
+    variant?: 'default' | 'takeover';
```
BenGSchulz: I wonder if we should use `cva` to handle the variants and classes? We use that in other wind components.

dbacomputer: Applied in 849854eb: DialogContent visual variant classes now use class-variance-authority (cva), matching the established Apollo Wind pattern while preserving the existing default and takeover styles. Focused dialog tests (16) and Biome checks pass.

packages/apollo-wind/src/components/ui/dialog.tsx
```diff
@@ -52,38 +69,152 @@ const DialogOverlay = React.forwardRef<
...
+              {headerTitle !== undefined && headerTitle !== null ? (
+                <DialogPrimitive.Title className="min-w-0 flex-1 truncate text-sm font-semibold">
+                  {headerTitle}
+                </DialogPrimitive.Title>
+              ) : (
+                <div className="min-w-0 flex-1" aria-hidden="true" />
+              )}
+              {headerActions}
+              <div className="-mr-1 flex shrink-0 items-center gap-1">
+                <button
```
BenGSchulz: Should this be using apollo wind Button component?

packages/apollo-wind/src/components/ui/dialog.tsx
BenGSchulz, line 1: This component might honestly read easier / be easier to edit each variant if we split into an early return for the takeover rather than interspersing a bunch of checks for `isTakeover`. It would cost some duplication, but could be worth the tradeoff.

What do you think?

packages/apollo-wind/src/components/ui/dialog.tsx
dbacomputer: Applied in eee36127: the takeover expand and close controls now use Apollo Wind Button primitives with ghost/icon variants, preserving the existing 32px control sizing and accessibility labels.

packages/apollo-wind/src/components/ui/dialog.tsx
dbacomputer: I kept a shared render path here because the portal/container behavior, overlay dismissal policy, accessibility labeling, and controlled expansion state are shared between the default and takeover variants. The variant-specific class composition is now isolated in cva, while the explicit takeover branch keeps the structural differences easy to follow. Resolved after review.

packages/apollo-wind/src/components/ui/dialog.tsx
```diff
@@ -52,38 +70,156 @@ const DialogOverlay = React.forwardRef<
...
+                  className="size-8 rounded-md text-foreground-muted hover:bg-surface-hover hover:text-foreground focus-visible:ring-border-focus"
+                >
+                  {expanded ? <Minimize2 size={16} /> : <Maximize2 size={16} />}
+                </Button>
+                <DialogPrimitive.Close asChild>
+                  <Button
+                    variant="ghost"
+                    size="icon"
+                    aria-label="Close modal"
+                    className="size-8 rounded-md text-foreground-muted hover:bg-surface-hover hover:text-foreground focus-visible:ring-border-focus"
```
BenGSchulz: This styling is inconsistent. I think we should either:
- Use same styling as the standard dialog close
- Do not override our own Button styling, just use built in variant/sizes options

packages/apollo-wind/src/components/ui/dialog.tsx
```diff
@@ -52,38 +70,156 @@ const DialogOverlay = React.forwardRef<
...
+              )}
+              {headerActions}
+              <div className="-mr-1 flex shrink-0 items-center gap-1">
+                <Button
+                  variant="ghost"
+                  size="icon"
+                  aria-label={expanded ? 'Collapse modal' : 'Expand modal'}
+                  title={expanded ? 'Collapse modal' : 'Expand modal'}
+                  onClick={() => setExpanded(!expanded)}
+                  className="size-8 rounded-md text-foreground-muted hover:bg-surface-hover hover:text-foreground focus-visible:ring-border-focus"
```
BenGSchulz: Same with the overrides here.

packages/apollo-wind/src/components/ui/dialog.tsx
dbacomputer: Applied in 7bad614c: removed the custom class overrides from both takeover header controls. They now rely on the standard Apollo Wind Button `ghost` and `icon` variants.

packages/apollo-wind/src/components/ui/dialog.tsx
dbacomputer: Applied in 7bad614c: both takeover header controls now use the standard Button variants without custom styling overrides.

apps/storybook/src/patterns/LayoutPatterns.stories.tsx
```diff
@@ -534,6 +668,900 @@ function DapValidationPanel({ onClose }: { onClose: () => void }) {
...
+                      }}
+                      onDrop={() => {
+                        if (!draggedCompositionField || draggedCompositionField === field) return;
+                        setCompositionFields((fields) => {
+                          const nextFields = [...fields];
+                          const fromIndex = nextFields.indexOf(draggedCompositionField);
+                          const toIndex = nextFields.indexOf(field);
+                          nextFields.splice(fromIndex, 1);
+                          nextFields.splice(toIndex, 0, draggedCompositionField);
+                          return nextFields;
```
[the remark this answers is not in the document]
dbacomputer: Fixed in 62d7a80c: after removing the dragged item, the target index is decremented by one when moving downward (fromIndex < toIndex), since the removal shifts every later index down by one. Downward drops now land on the intended slot.

packages/apollo-wind/src/components/ui/dialog.tsx
```diff
@@ -52,38 +70,150 @@ const DialogOverlay = React.forwardRef<
...
+    if (controlledExpanded === undefined) setUncontrolledExpanded(nextExpanded);
+    onExpandedChange?.(nextExpanded);
+  };
+
+  const content = (
+    <DialogOverlay
+      {...overlayProps}
+      data-testid={overlayTestId}
+      className={cn(
+        isTakeover
```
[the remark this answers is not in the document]
dbacomputer: Fixed in cad5ac00: overlayProps click/mouse/pointer handlers are now guarded to only fire when the event target is the overlay element itself, so interactions inside DialogContent (which is nested for centering) no longer bubble into backdrop handlers.

packages/apollo-react/src/canvas/stories/templates/Flow.stories.tsx
```diff
@@ -4381,52 +4382,54 @@ export function FullWorkbenchComposition({
...
-        </CanvasTakeoverModal>
+        <Modal open={nodeTakeoverOpen} onOpenChange={setNodeTakeoverOpen}>
+          <ModalContent
+            variant="takeover"
+            headerTitle="Send Email"
+            headerActions={
+              <Button size="sm" variant="secondary">
+                <Play size={14} /> Run node
+              </Button>
+            }
```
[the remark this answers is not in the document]
dbacomputer: Fixed: both takeover/property-panel action buttons now render just "Run" (with the Play icon), matching the standardized label used elsewhere in this PR.

dbacomputer:
Fixed in d958e615: the Component Inventory date/date-time picker defaults no longer hardcode a fixed calendar date. They're now computed relative to the current date (7 days out) so the example stays stable and doesn't drift into stale-looking or time-dependent behavior.

packages/apollo-wind/src/components/ui/dialog.tsx
```diff
@@ -52,38 +94,150 @@ const DialogOverlay = React.forwardRef<
...
+                  {headerTitle}
+                </DialogPrimitive.Title>
+              ) : (
+                <div className="min-w-0 flex-1" aria-hidden="true" />
+              )}
+              {headerActions}
+              <div className="-mr-1 flex shrink-0 items-center gap-1">
+                <Button
+                  variant="ghost"
+                  size="icon"
```
BenGSchulz: Is `size="icon"` right here?

The buttons seem a bit too big to me.

Image

packages/apollo-wind/src/components/ui/dialog.stories.tsx
```diff
@@ -147,6 +147,71 @@ export const ComplexDialog = {
...
+          </div>
+        </DialogContent>
+      </Dialog>
+    </div>
+  );
+}
+
+export const TakeoverWithSidebar = {
+  name: 'Takeover w/ Sidebar',
+  render: () => <TakeoverModalStory withSidebar />,
```
BenGSchulz, line 208: This doesn't appear to play well with the docs overview story for dialog.

https://github.com/user-attachments/assets/94b2a9a8-464b-4d4e-bfa1-29bdcaeb6bb4

packages/apollo-wind/src/components/ui/dialog.tsx
```diff
@@ -52,38 +94,150 @@ const DialogOverlay = React.forwardRef<
...
+                  aria-label={expanded ? 'Collapse modal' : 'Expand modal'}
+                  title={expanded ? 'Collapse modal' : 'Expand modal'}
+                  onClick={() => setExpanded(!expanded)}
+                >
+                  {expanded ? <Minimize2 size={16} /> : <Maximize2 size={16} />}
+                </Button>
+                <DialogPrimitive.Close asChild>
+                  <Button variant="ghost" size="icon" aria-label="Close modal">
```
BenGSchulz: Same sizing here

packages/apollo-wind/src/components/ui/dialog.tsx
dbacomputer: Good catch, fixed in 5f99cf2d: switched to size="xs" icon (32px), matching the sizing agreed on earlier in review for these header controls. size="icon" is the 40px default, which was too big for the 48px header.

packages/apollo-wind/src/components/ui/dialog.tsx
dbacomputer: Fixed in 5f99cf2d: same 32px sizing applied here (size="xs" icon) for consistency with the Expand/Collapse control.

packages/apollo-wind/src/components/ui/dialog.stories.tsx
dbacomputer: Fixed in 5f99cf2d: added tags: ['!autodocs'] to both Takeover stories so they no longer auto-embed in the component's Docs overview page (they're full-viewport interactive demos that don't render well as a static inline block there). They're still available in the sidebar/canvas as before.

BenGSchulz (approved):

PR #1118, opened 2026-09-03
chore(storybook): bump Storybook and addon-mcp to 10.6.0

dbacomputer:
## What

Bumps Storybook core and all @storybook/addon-* packages from 10.4.1 to 10.6.0, and @storybook/addon-mcp from 0.2.2 to 10.6.0. Three files are affected: apps/storybook/package.json, packages/apollo-wind/package.json, and packages/apollo-react/package.json.

## Why

We were 8 months and roughly 12 releases behind on the MCP addon (0.2.2, released Jan 26). The addon has since dropped its own independent versioning and now tracks Storybook's core version directly, so addon-mcp and storybook need to move together going forward.

## Notes for reviewers

Verified peer-dependency compatibility for every bumped package directly against npm metadata for 10.6.0, all satisfied.

@storybook/addon-mcp@10.6.0 adds a new optional peer dependency on @storybook/addon-vitest, which is not currently installed anywhere in this repo. It is optional so it will not block install, but it powers addon-mcp's test-running MCP tool. Flagging in case we want to add it in a follow-up.

Did not run a full pnpm install in CI-equivalent conditions locally due to sandbox memory limits, so please confirm CI installs and Storybook builds cleanly on this branch before merging.

## Testing checklist

CI install passes. pnpm storybook boots cleanly in apps/storybook and packages/apollo-wind. The /mcp endpoint is reachable on the local dev server.


dbacomputer:
Converting to draft: this needs the lockfile updated before it can merge.

Copilot's review and the CI failures both point to the same root cause: I only bumped the version strings in package.json, but pnpm-lock.yaml was not regenerated to match. CI runs pnpm install --frozen-lockfile, so it fails at the install step, which cascades into Build, Lint, Typecheck, Test Coverage, and the Coded App preview deployments all failing.

I wasn't able to fix this myself: regenerating the lockfile requires running pnpm install, and I don't have a path to do that against this repo (no write-capable local environment, and a full install attempt hit a memory limit in my sandbox). This needs someone with the repo checked out locally to run pnpm install on this branch, commit the updated pnpm-lock.yaml, and push.

Opening this for discussion in the meantime. Marking as Ready for review once the lockfile is updated and checks are green.

PR #1121, opened 2026-09-04
feat(apollo-wind): add PromptEditor extension points for host integration

akraminakib:
## Summary

Adds host-integration extension points to apollo-wind's `PromptEditor` so Flow Workbench (MST-13998) can consume the published component instead of maintaining its diverged fork, and so connector rich-text fields can embed a value-mode control inside the editor toolbar per the design guidance. All changes are additive — every new prop defaults to today's behavior.

## Demo

https://github.com/user-attachments/assets/9c4fbb7c-824b-4762-b27a-2fe19becc363

## Changes

- **`validationOptions`** — validate chips against a wider set than `autoCompleteOptions` (runtime-only paths) without polluting the `$`-trigger menu
- **Index-normalized chip validation** — `records[0].id` in the option set now validates user-entered indices like `records[1].id` (`normalizeVariablePath`, also applied to free-form type inference; `VARIABLE_PATH_REGEX` accepts `[n]` segments)
- **`renderAutocompleteMenu`** — swap the built-in cmdk menu for a host picker; receives the exact props the built-in menu gets, and enables the `$` trigger even with empty options
- **`renderTokenPill`** — replace or decorate token chips (the built-in pill is passed through as `defaultPill`); delivered to decorator nodes via a new `PromptEditorConfigContext`
- **`children`** — mount extra Lexical plugins inside the composer (host drag-drop, chip-replace commands)
- **`strings`** — overridable UI strings for localizing hosts (apollo-wind has no i18n runtime); menu strings read the same config context
- **`showModeToggle={false}` + `toolbarTrailing`** — widget-style toolbar: Edit/Preview hidden, formatting cluster left-aligned, host control (e.g. a value-mode menu) right-aligned; new `WithTrailingModeControl` story
- **Barrel exports** — bless the helpers host plugins need (`$insertTokenAtCursor`, `WORD_JOINER`, `createTokenNodeForOption`, `isPromptTokenNode`, `getEditorTokens`/`setEditorTokens`, `TokenPill`, token node classes, …)

## Flow

```mermaid
flowchart TD
    Host[Host app e.g. Flow Workbench] -->|props| PE[PromptEditor]
    PE -->|strings + renderTokenPill| CTX[PromptEditorConfigContext]
    PE --> TB[EditorToolbar<br/>showModeToggle / trailing / strings]
    PE --> AC[AutocompletePlugin<br/>renderMenu override]
    PE --> VT[ValidateTokensPlugin<br/>validationOptions + index-normalized]
    PE -->|children| XP[Host Lexical plugins]
    CTX --> SLOT[TokenPillSlot in decorator nodes]
    SLOT -->|default| PILL[TokenPillWithTooltip]
    SLOT -->|override| HP[Host pill renderer]
```

## Testing

- [x] `pnpm lint` passes (JS, CSS, deps)
- [x] `pnpm typecheck` passes
- [x] `pnpm test` passes (apollo-wind 93 files / prompt-editor 70 tests incl. 13 new extension-point specs; one pre-existing, unrelated `apollo-react` `formatDuration` failure from local Node ICU data)
- [ ] `pnpm test:visual` — no local turbo task; runs in CI
- [x] Manual testing / Storybook checked (`WithTrailingModeControl` story)

## Dependency audit overrides

Unrelated to the editor work but required for green CI: fresh `pnpm audit` advisories surfaced on this PR's runs, all in transitive dependencies. `pnpm-workspace.yaml` gains a vetted override batch (`browserslist`, `fast-uri` (kept on ajv's 3.x line), `sanitize-html`, `qs`, `@tiptap/core` + `@tiptap/pm` (bumped together, they peer in lockstep), `@xmldom/xmldom`, `fflate`, `postcss-selector-parser`). Each version still inside the 14-day quarantine has a `minimumReleaseAgeExclude` entry with its vetting notes (provenance/publisher, tag+diff, install scripts, integrity) inline in the file.

Refs: MST-13998


packages/apollo-wind/src/components/ui/prompt-editor/plugins/NodeSelectionFixPlugin.tsx
```diff
@@ -22,8 +23,16 @@ import {
...
-import { isPromptTokenNode, type PromptTokenNode } from './shared/token-nodes';
+import { useEffect } from 'react';
 import { WORD_JOINER } from '../utils';
+import { isPromptTokenNode, type PromptTokenNode } from './shared/token-nodes';
+
+/** Non-inline element that can host inline children directly (paragraph in plain mode; list items
+ *  and paragraphs in rich mode). The caret-vs-pill fixes below apply inside any of them. */
+const $isBlockElement = (
+  node: LexicalNode | null | undefined
+): node is import('lexical').ElementNode =>
```
BenGSchulz: Move this to standard imports?

packages/apollo-wind/src/components/ui/prompt-editor/prompt-editor-config.tsx
```diff
@@ -0,0 +1,79 @@
...
+  defaultPill: ReactNode;
+}
+
+export type PromptEditorRenderTokenPill = (props: PromptEditorTokenPillSlotProps) => ReactNode;
+
+/**
+ * Overridable user-facing strings. Apollo-wind has no i18n runtime of its own; hosts that localize
+ * (e.g. via react-i18next) pass translated values here. Every key defaults to the built-in English.
+ */
+export interface PromptEditorStrings {
```
BenGSchulz, line 30: Claude helped find some missing strings:
```
- PromptEditorAutocompleteMenu.tsx:200 → getPromptEditorTokenTypeLabel(option.type) (reads strings on line 79, then hardcodes on 200)
- TokenPillWithTooltip.tsx:24 → same function, in the chip tooltip
- types.ts:103-113 → 'Input variable', 'Output variable', 'State variable', 'Resource'
- TokenPill.tsx:95 → aria-label="Remove token"
- MarkdownPreview.tsx:190 → 'Nothing to preview'
```
```
// prompt-editor-config.tsx — add to PromptEditorStrings + DEFAULT_PROMPT_EDITOR_STRINGS
  tokenTypeInput: string;      // 'Input variable'
  tokenTypeOutput: string;     // 'Output variable'
  tokenTypeState: string;      // 'State variable'
  tokenTypeResource: string;   // 'Resource'
  removeToken: string;         // 'Remove token'
  nothingToPreview: string;    // 'Nothing to preview'
```

packages/apollo-wind/src/components/ui/prompt-editor/components/PromptEditorAutocompleteMenu.tsx
```diff
@@ -197,7 +208,7 @@ export const PromptEditorAutocompleteMenu = ({
             <CommandItem value={freeFormPath} onSelect={() => onCommitFreeForm?.(freeFormPath)}>
               <Variable className="h-3.5 w-3.5 shrink-0 text-muted-foreground" />
               <span className="truncate">
-                Insert <span className="font-medium">{freeFormPath}</span>
+                {strings.insertFreeForm} <span className="font-medium">{freeFormPath}</span>
```
BenGSchulz: This may not be sufficiently translated in languages that use verb last. I guess tradeoff between translation complexity vs UI treatment.

packages/apollo-wind/src/components/ui/prompt-editor/utils/rich-serialization.ts
```diff
@@ -0,0 +1,211 @@
...
+  return {
+    dependencies: [InputTokenNode, OutputTokenNode, StateTokenNode, ResourceTokenNode],
+    regExp: PILL_SENTINEL_IMPORT,
+    export: (node) => {
+      if (!isPromptTokenNode(node)) {
+        return null;
+      }
+      const index = pills.length;
+      pills.push({
+        type: node.getType().replace('-token', '') as PillRef['type'],
```
BenGSchulz: `NODE_TYPE_TO_TOKEN_TYPE` exists that could be typed to probably avoid this cast.

pnpm-lock.yaml
```diff
@@ -659,35 +667,35 @@ importers:
         specifier: ^3.14.3
         version: 3.14.3(react-dom@19.2.3(react@19.2.3))(react@19.2.3)
       '@tiptap/core':
-        specifier: ^3.19.0
-        version: 3.19.0(@tiptap/pm@3.19.0)
+        specifier: '>=3.30.4'
+        version: 3.30.4(@tiptap/pm@3.19.0)
       '@tiptap/extension-document':
```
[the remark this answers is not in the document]
BenGSchulz: Curious about this, is there a reason we would not want to update /pm too?

BenGSchulz (commented):
These items can be addressed as follow-ups most likely.

packages/apollo-wind/src/components/ui/prompt-editor/plugins/NodeSelectionFixPlugin.tsx
akraminakib: Done. `ElementNode` is now a regular type import from `lexical` and the inline `import('lexical')` type reference is gone. (1e8442fa)

packages/apollo-wind/src/components/ui/prompt-editor/prompt-editor-config.tsx
akraminakib: Nice catch, added all six to `PromptEditorStrings` with English defaults: the four token type labels, `removeToken`, and `nothingToPreview`. `getPromptEditorTokenTypeLabel` now takes an optional labels object (structurally satisfied by `PromptEditorStrings`, kept structural to avoid a config/types import cycle), and the chip tooltip, autocomplete hint, remove button, and empty preview all read from the config context now. (1e8442fa)

packages/apollo-wind/src/components/ui/prompt-editor/components/PromptEditorAutocompleteMenu.tsx
akraminakib: Agreed, concatenation does not survive verb-last languages. `insertFreeForm` is now a template: the default is `Insert {path}` and the menu renders the label around the `{path}` placeholder (path stays bold), so a translation can put the verb after the path. A string without `{path}` is still treated as a prefix, so existing consumers passing a plain prefix keep working. (1e8442fa)

packages/apollo-wind/src/components/ui/prompt-editor/utils/rich-serialization.ts
akraminakib: Good idea. Moved `NODE_TYPE_TO_TOKEN_TYPE` to `shared/token-nodes.ts`, typed as `Record<string, Exclude<PromptEditorTokenType, 'text'>>`, and both `ValidateTokensPlugin` and the rich serialization now share it. The cast is gone; the export transformer just skips a node whose type is not in the map. (1e8442fa)

pnpm-lock.yaml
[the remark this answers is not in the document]
akraminakib: No reason, updating pm is the right call and it is done in a985860c: `@tiptap/pm` is overridden to 3.30.4 with its own vetted quarantine exemption (npm provenance attestation, no install scripts, it is just the prosemirror re-export bundle). Core and pm release in lockstep, so the lockfile now resolves the pair together, and the extensions' `^3.19` peer range on pm is still satisfied.

packages/apollo-wind/src/components/ui/prompt-editor/prompt-editor.tsx
```diff
@@ -433,15 +565,39 @@ export const PromptEditor = ({
...
+      return false;
+    }
+    return !!richTextProp;
+  });
+
   const [internalMode, setInternalMode] = useState<PromptEditorMode>('edit');
-  const mode = controlledMode ?? internalMode;
+  // Without the Edit/Preview switcher there is no way (or reason) to enter preview mode; in rich
+  // mode the editor IS the preview, so preview mode is inert there too.
+  const mode = showModeToggle && !richText ? (controlledMode ?? internalMode) : 'edit';
```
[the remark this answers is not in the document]
BenGSchulz: I guess this seems good to address.

packages/apollo-wind/src/components/ui/prompt-editor/utils/autocomplete-segments.ts
```diff
@@ -46,7 +54,7 @@ export const inferTokenTypeFromPath = (
  * paths can be committed the same as input/output ones.
  */
 export const VARIABLE_PATH_REGEX =
-  /^(?:vars|metadata|agent|state|resource)\.[a-zA-Z_][a-zA-Z0-9_.]*$/;
+  /^(?:vars|metadata|agent|state|resource)\.[a-zA-Z_](?:[a-zA-Z0-9_.]|\[\d+\])*$/;
```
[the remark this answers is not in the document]
akraminakib: Fixed in 712f5e60. The regex is now per-segment (identifier, then optional `[n]` indexes), so `vars.records[0]id`, consecutive dots, and unclosed brackets are all rejected; added those as test cases.

packages/apollo-wind/src/components/ui/prompt-editor/prompt-editor.stories.tsx
```diff
@@ -91,6 +91,61 @@ export const WithToolbar: Story = {
...
+    autoCompleteOptions: AUTOCOMPLETE_OPTIONS,
+    ariaLabel: 'Body',
+  },
+};
+
+/**
+ * WYSIWYG mode: formatting renders live while editing (real bold/strike and lists — no Edit/Preview
+ * switcher, the editor IS the preview). Text tokens still carry markdown; try typing `**bold**` or
+ * `- ` for a live markdown shortcut.
+ */
```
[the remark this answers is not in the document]
akraminakib: Fixed in dcea7ca5, both story doc comments now use a comma/colon instead.

pnpm-workspace.yaml
```diff
@@ -44,6 +48,16 @@ overrides:
...
+  fast-uri: "^3.1.6"                    # GHSA-5jgf-p345-68v8/f65p/fph4/jqff (high, host confusion + SSRF) — ajv transitive, kept on the 3.x line ajv expects (4.x has its own unpatched-in-range advisories); quarantine exemption above
+  sanitize-html: ">=2.17.7"             # GHSA-g8qq-57p8-ggw5 (moderate, stored XSS via SVG SMIL URI-list) — 2.17.7 published 2026-08-13 (aged)
+  qs: ">=6.16.0"                        # GHSA-4mjr-xmp4-gh2g + GHSA-x5fp-wj9c-mxmx (moderate) — quarantine exemption above
+  "@tiptap/core": ">=3.30.4"            # GHSA-cp6q-959q-f8rh (moderate, mergeAttributes __proto__ pollution) — ap-chat dep, within apollo-react's ^3.19 range; quarantine exemption above
+  "@tiptap/pm": ">=3.30.4"              # no advisory — @tiptap/core@3.30.4 peers on @tiptap/pm 3.30.4 exactly (lockstep releases); quarantine exemption above
+  "@xmldom/xmldom": ">=0.9.12"          # GHSA-6gmq-8vp8-gcm6 (moderate) — quarantine exemption above
+  fflate: ">=0.8.3"                     # GHSA-px8p-9vwx-vf98 (moderate) — 0.8.3 published 2026-05-16 (aged)
+  postcss-selector-parser: ">=7.1.3"    # GHSA-w9m9-85wc-3x92 (low) — 7.1.3 published 2026-06-11 (aged)
```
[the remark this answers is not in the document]
akraminakib: Fair point. Added a "Dependency audit overrides" section to the PR description covering the batch and where the per-version vetting notes live (inline in pnpm-workspace.yaml next to each quarantine exemption).

packages/apollo-wind/src/components/ui/prompt-editor/prompt-editor.tsx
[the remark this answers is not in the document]
akraminakib: Fixed in 712f5e60. The mode now locks to edit only in rich mode or when a toolbar renders without the Edit/Preview switcher; with no toolbar the controlled mode prop is respected, so preview-only hosts work again. Added a test for that case.

BenGSchulz (approved):

akraminakib:
Addressed the "needs a closer look" note from the last Copilot review: the invalid-chip tooltip title and body are now in `PromptEditorStrings` (the body is a `{path}` template), read from `usePromptEditorConfig()` like the rest. That was the last hardcoded user-facing copy in the editor. (See the latest commit.)

BenGSchulz (approved):

PR #1123, opened 2026-09-04
perf(apollo-wind): re-render only the edited field in MetadataForm

snuziale:
`MetadataForm`'s top-level `watch()` re-rendered the entire form on every keystroke, rebuilding the Zod schema and form context and re-rendering every section and field. Values now flow through a ref fed by one subscription, the schema reference is stabilized by deep equality, and the sub-components are memoized.

Measured on a 10-field, 2-section form, one keystroke:

| | before | after |
|---|---|---|
| `MetadataForm` | 1 | 0 |
| `SinglePageForm` | 1 | 0 |
| `FormSection` | 2 | 0 |
| `FormFieldRenderer` | 10 | 0 |

Only the edited field's `Controller` re-renders.

https://github.com/user-attachments/assets/b18778c5-0c11-43c5-adba-15adeb1445da

### Keeping conditions reactive

Section and step `conditions` are evaluated during render from `context.values`. With the form no longer re-rendering on input, and the sub-components memoized behind stable props, they would have silently stopped reacting. Each layout now subscribes with `useWatch` to only the fields its own conditions reference, so typing in an unrelated field still costs nothing.

Without this the `Forms/Examples > Conditional Sections` story breaks: selecting a mid/senior experience level never reveals the "Work Experience" section. Nothing in the existing suite covered it, so two regression tests were added and confirmed to fail with the fix reverted.

### Also included

- `defaultValues` seeded from `initialData`, so fields paint with their values on first render
- a select whose `options` go away now clears instead of keeping a stale list
- `FormStateViewer` subscribes instead of calling `watch()`, which set `watchAll` on the shared control and re-rendered the `useForm` owner once per keystroke (measured: 3 keystrokes → 3 parent renders, now 0). Its test mock gained `getValues` to match.
- a test pinning the ordering guarantee that plugins see the current value in `context.values`, not the previous one

### Verification

1457 tests, typecheck, and build pass. Biome reports the same three pre-existing findings as `main`, none new. `Conditional Sections`, `Tabbed Steps`, `Multi Step Wizard`, `Cascading Dropdowns`, `With State Viewer`, and `Section Variants` verified in the browser.

### Known gaps, unchanged by this PR

- Wizard mode still ignores section-level `conditions` entirely, unlike the single-page and tabbed layouts
- A wizard whose remaining steps are all hidden renders nothing, with no navigation back. A test now documents it rather than leaving it implicit.
- Consumers that round-trip every value into `schema.initialData` (the Form Builder's field inspector) invalidate the deep-equality check each keystroke and so see none of this benefit

### DIFF
--- a/packages/apollo-wind/src/components/forms/form-state-viewer.tsx
+++ b/packages/apollo-wind/src/components/forms/form-state-viewer.tsx
@@ -1,10 +1,10 @@
+import { AlertCircle, CheckCircle2, Database, Eye, FileEdit, XCircle } from 'lucide-react';
 import { useState } from 'react';
-import type { UseFormReturn, FieldValues } from 'react-hook-form';
+import { type FieldValues, type UseFormReturn, useWatch } from 'react-hook-form';
 import { Badge } from '@/components/ui/badge';
 import { Card } from '@/components/ui/card';
-import { Tabs, TabsContent, TabsList, TabsTrigger } from '@/components/ui/tabs';
 import { ScrollArea } from '@/components/ui/scroll-area';
-import { CheckCircle2, XCircle, AlertCircle, FileEdit, Eye, Database } from 'lucide-react';
+import { Tabs, TabsContent, TabsList, TabsTrigger } from '@/components/ui/tabs';
 
 /**
  * FormStateViewer Component
@@ -27,8 +27,14 @@ export function FormStateViewer({
   compact = false,
 }: FormStateViewerProps) {
   const [activeTab, setActiveTab] = useState('values');
-  const { formState, watch } = form;
-  const values = watch();
+  const { formState, control, getValues } = form;
+
+  // useWatch re-renders only this component. A bare watch() would instead set
+  // watchAll on the shared control, re-rendering the useForm owner (e.g.
+  // MetadataForm) once per keystroke. Values are read fresh on every render so
+  // fields that register after mount are included.
+  useWatch({ control });
+  const values = getValues() as Record<string, unknown>;
 
   const stats = {
     isValid: formState.isValid,
--- a/packages/apollo-wind/src/components/forms/metadata-form.tsx
+++ b/packages/apollo-wind/src/components/forms/metadata-form.tsx
@@ -1,6 +1,6 @@
 import { standardSchemaResolver } from '@hookform/resolvers/standard-schema';
-import React, { useEffect, useMemo, useRef, useState } from 'react';
-import { type FieldValues, FormProvider, useForm, useFormState } from 'react-hook-form';
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { type FieldValues, FormProvider, useForm, useFormState, useWatch } from 'react-hook-form';
 import { z } from 'zod/v4';
 import {
   Accordion,
@@ -10,6 +10,7 @@ import {
 } from '@/components/ui/accordion';
 import { Button } from '@/components/ui/button';
 import { ScrollableTabsList, Tabs, TabsContent, TabsTrigger } from '@/components/ui/tabs';
+import { deepEqual, get } from '@/lib';
 
 import { DataFetcher } from './data-fetcher';
 import { FormFieldRenderer } from './field-renderer';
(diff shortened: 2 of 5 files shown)


packages/apollo-wind/src/components/forms/metadata-form.tsx
```diff
@@ -712,15 +789,19 @@ function FormSection({
...
   onReset: () => void;
 }
 
-function FormActions({ schema, context, onReset }: FormActionsProps) {
+const FormActions = React.memo(function FormActions({
+  schema,
+  context,
+  onReset,
+}: FormActionsProps) {
   const actions = schema.actions || [
```
david-rios-uipath, line 815: will this component update properly on form state change? 

e.g. `isSubmitting` or on a form value change

packages/apollo-wind/src/components/forms/form-state-viewer.tsx
```diff
@@ -28,7 +28,16 @@ export function FormStateViewer({
 }: FormStateViewerProps) {
   const [activeTab, setActiveTab] = useState('values');
   const { formState, watch } = form;
-  const values = watch();
+
+  // A bare watch() sets watchAll on the shared control, re-rendering the useForm
```
david-rios-uipath: seeing that initial state in the stories is not taking empty fields into account by default: 

Image

Is this intentional?

david-rios-uipath (approved):
thanks for optimizing this!

packages/apollo-wind/src/components/forms/metadata-form.tsx
snuziale: Good catch, both were broken. Confirmed with a test before fixing:

```
A isSubmitting      -> FAIL (label="Save" disabled=false)
B action conditions -> FAIL (conditional action never appeared)
```

Memoizing `FormActions` cut it off from both. It reads `context.isSubmitting` and evaluates action `conditions`, but `context` is a stable reference, so neither reached it. The submit button never disabled during an async submit (double-submit was possible) and conditional actions froze.

It now subscribes to what it reads:

```tsx
const { isSubmitting } = useFormState({ control: context.form.control });
useConditionDependencies(context, conditionFields); // action conditions
```

Two tests added, both verified to fail with the subscriptions removed. This was the same class of bug as the section/step conditions, and I'd missed that actions evaluate conditions too.

packages/apollo-wind/src/components/forms/form-state-viewer.tsx
snuziale: Not intentional, and it was a regression from this PR. I A/B'd the story against main:

| | values |
|---|---|
| main | `{ subscribe, age, name: "", email: "" }` |
| this PR (before fix) | `{ subscribe, age }` |

main called `watch()` on every render, so it self-healed as inputs registered. I'd replaced that with a single `useState` snapshot taken before registration, which then never refreshed.

Fixed by subscribing without holding a snapshot:

```tsx
useWatch({ control });
const values = getValues() as Record<string, unknown>;
```

Values are read fresh each render, so late-registering fields are included, and it matches main's output exactly now. Measured that it keeps the perf win too, unlike a bare `watch()`:

```
bare watch()  owner 2->5 (one re-render per keystroke)
useWatch      owner 2->2, child 2->5
```

(`role` is still absent until touched on both main and this branch, since a Controller with no default writes nothing to form state.)

packages/apollo-wind/src/components/forms/form-state-viewer.tsx
```diff
@@ -28,7 +28,16 @@ export function FormStateViewer({
...
+  // owner (MetadataForm) once per keystroke. Subscribing does not.
+  const [values, setValues] = useState<Record<string, unknown>>(() => form.getValues());
+  useEffect(() => {
+    const subscription = watch((newValues) => {
+      setValues(newValues as Record<string, unknown>);
+    });
+    return () => subscription.unsubscribe();
+  }, [watch]);
```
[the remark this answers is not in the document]
snuziale: Correct, fixed. The snapshot was seeded once at mount and never re-read, so a new `form` instance kept showing the previous one's values. It now calls `useWatch({ control })` and reads `getValues()` during render, so there's no snapshot to go stale. This also fixed a related regression david spotted: fields registering after mount were missing from the initial values.

packages/apollo-wind/src/components/forms/metadata-form.tsx
```diff
@@ -146,13 +161,31 @@ export function MetadataForm({
...
+  // valuesRef is written before the plugin fan-out: context.values must be current.
+  useEffect(() => {
+    const subscription = watch((value, { name }) => {
+      valuesRef.current = value as Record<string, unknown>;
+
+      if (!name || !isInitializedRef.current) return;
+
+      pluginsRef.current.forEach((plugin) => {
+        plugin.onValueChange?.(name, value[name], contextRef.current);
+      });
```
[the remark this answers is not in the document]
snuziale: Valid bug, fixed via `get(value, name)`. Worth noting it predates this PR: `value[name]` is unchanged from the original code, which this PR only moved. It's real though, since this codebase does use dot-paths (`inputs.url`, `display.label` in the tabbed-steps schema), and `get` was already exported from `@/lib`. Added a test asserting a plugin receives `['inputs.url', 'a']`, verified to fail with the bare index.

packages/apollo-wind/src/components/forms/metadata-form.tsx
```diff
@@ -177,62 +210,49 @@ export function MetadataForm({
     initializeForm();
   }, [isInitialized]);
```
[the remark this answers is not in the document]
snuziale: Declining this one: the dependency list, the `biome-ignore`, and the init-once design are all unchanged from main. The only edit this PR makes inside that effect is `schema.initialData` -> `stableSchema.initialData`:

```diff
-      if (schema.initialData) {
-        const data = await loadInitialData(schema.initialData, contextRef.current);
+      if (stableSchema.initialData) {
+        const data = await loadInitialData(stableSchema.initialData, contextRef.current);
```

`stableSchema` is deep-equality-stabilized, so if anything this narrows the window rather than widening it.

On the stated mechanism: an effect runs with the values from the render being committed, so it can't observe something "stale between first render and effect execution". The real behavior is the documented one, which the existing `biome-ignore` spells out: the form initializes once and a schema swap needs a `key` prop to remount. That's a pre-existing design constraint worth revisiting, but not a defect introduced here, so I'd rather not change initialization semantics inside a perf PR.

david-rios-uipath (approved):

PR #1144, opened 2026-09-09
fix(apollo-vertex): namespace the registry's own item references

kittyyueli:
## Summary

44 `registryDependencies` across 15 items were bare names (`"button"`, `"sheet"`, `"skeleton"`…). The shadcn CLI resolves a bare name against its **default registry**, so installing an Apollo item silently pulls shadcn's component instead of this one's.

Prefixes all 44 with `@uipath/`, matching the 82 references that already were.

## How it surfaced

A consumer ran `shadcn add @uipath/sidebar @uipath/collapsible`. The sidebar files arrived correctly, but the CLI also resolved `sidebar`'s bare dependencies from shadcn and overwrote five files the app already had:

| File | What the consumer got |
| --- | --- |
| `button.tsx` | shadcn's button — lost the `ai`, `success`, `info`, `warning` and `destructive-outline` variants. Their app renders `variant="ai"` |
| `separator.tsx`, `sheet.tsx`, `skeleton.tsx` | shadcn versions, plus `skeleton` restyled from `bg-muted` to `bg-accent` |
| `use-mobile.ts` | shadcn's, which passes a redundant `undefined` their linter rejects |

All five imported `cn` from a `cn` **npm package** rather than `@/lib/utils`, and pulled `cn` and `radix-ui` into `package.json`. Without `twMerge`, a caller's `className` stops overriding — and their whole app failed to render, taking every E2E test with it.

## Why this is drift, not intent

- **82 of 126 references were already prefixed** — `charts-core → @uipath/spinner`, `line-chart → @uipath/card`, and so on.
- The docs tell consumers to configure this registry as `@uipath` (`app/mcp/page.mdx`, `app/patterns/*/page.mdx`).
- `metric-card` had it both ways inside a single array: `["@uipath/card", "@uipath/badge", "skeleton"]`.

## Items changed

`apollo-vertex-theme` · `button-group` · `combobox` · `command` · `confidence-signal` · `date-picker` · `field` · `form` · `input-group` · `item` · `metric-card` · `onboarding-tour-joyride` · `shell` · `sidebar` · `toggle-group`

Every bare name resolved to an item that exists in this registry — verified before rewriting — so nothing changes which component is intended, only which registry it comes from.

## Impact on consumers

Nothing here breaks a build, but two consequences are worth naming before this merges.

[...]

PR #1145, opened 2026-09-09
fix(apollo-vertex): make useIsMobile correct on first render, harden sidebar cookie

kittyyueli:
## Summary

Two findings from consuming `@uipath/sidebar` in a Vite SPA. Both are in registry source, so they cannot be fixed downstream — `components/ui/` is verbatim vendored output there.

## 1. `useIsMobile()` reports desktop on the first render

```ts
const [isMobile, setIsMobile] = React.useState<boolean | undefined>();
// …effect sets the real value…
return !!isMobile;
```

State starts `undefined` and is coerced with `!!`, so the first render always says desktop and the effect corrects it after paint. `Sidebar` branches on exactly that value, so on a phone it renders the desktop variant once and then swaps to the `Sheet`. A tap landing in that window toggles the desktop collapse state instead of opening the sheet.

Replaced with `useSyncExternalStore`, which reads the media query directly:

```ts
export function useIsMobile() {
  return React.useSyncExternalStore(subscribe, getSnapshot, getServerSnapshot);
}
```

**Why not just initialise from `matchMedia().matches`** — the obvious one-line fix. This file has no `"use client"`, so it is importable from a server render, and a `useState` initialiser would run there and throw on `window`. `getServerSnapshot` returns `false`, which keeps the module server-safe and preserves the current server-side behaviour exactly; the client then corrects on hydration. The `MediaQueryList` is resolved lazily for the same reason, and cached so `getSnapshot` does not allocate on every render.

## 2. `SidebarProvider` and the `sidebar_state` cookie

- Now written with `SameSite=Lax`.
- Documented that the provider **only writes** it.

That second half is the substantive part. The provider writes the cookie but never reads it — by design, since a Next.js consumer reads it in a server component and passes it back as `defaultOpen`. That contract was implicit, and an app with no server render gets no persistence at all: `defaultOpen` always wins and the collapsed state does not survive a reload. It looks like it works, because the cookie is visibly being written.

**Deliberately not changed: the initial state.** Reading the cookie inside `useState` here would desynchronise any SSR consumer that does not also pass `defaultOpen` — the server would render one value and the client hydrate with another. Documenting the contract fixes the surprise without putting a hydration mismatch into every Next.js app. Happy to revisit if you would rather the provider read it and SSR consumers opt out.

## Testing

| Check | Result |
| --- | --- |
| Lines over the 100-char width | none added (`sidebar-provider.tsx` keeps the 4 it already had, all templates/comments Biome cannot break) |
| Files touched outside `registry/` | none |
| `registry.json` | unchanged — no new files, no new dependencies |

I could not run `pnpm build` / `lint` / `format` in this session, so those are for CI to confirm. `Apollo Vertex Registry Check` installs both items into a fresh app, which exercises the changed files directly.

## Related

- [UiPath/apollo-ui#1144](https://github.com/UiPath/apollo-ui/pull/1144) — the `registryDependencies` namespace fix, from the same downstream install. Independent of this one.

https://claude.ai/code/session_01SnmiT3edq81s2qTDqoE8zP

---
_Generated by [Claude Code](https://claude.ai/code/session_01SnmiT3edq81s2qTDqoE8zP)_

### DIFF
--- a/apps/apollo-vertex/registry/sidebar/sidebar-provider.tsx
+++ b/apps/apollo-vertex/registry/sidebar/sidebar-provider.tsx
@@ -52,6 +52,12 @@ export function useSidebar() {
   return context;
 }
 
+/**
+ * This provider only *writes* `sidebar_state`. Reading it back is the consumer's
+ * job: a server component passes the cookie in as `defaultOpen`, and an app with
+ * no server render has to read it on the client and do the same. Without that,
+ * `defaultOpen` always wins and the collapsed state does not survive a reload.
+ */
 export function SidebarProvider({
   defaultOpen = true,
   open: openProp,
@@ -81,7 +87,7 @@ export function SidebarProvider({
     }
 
     // eslint-disable-next-line unicorn/no-document-cookie -- Standard shadcn pattern for persisting sidebar state
-    document.cookie = `${SIDEBAR_COOKIE_NAME}=${openState}; path=/; max-age=${SIDEBAR_COOKIE_MAX_AGE}`;
+    document.cookie = `${SIDEBAR_COOKIE_NAME}=${openState}; path=/; max-age=${SIDEBAR_COOKIE_MAX_AGE}; samesite=lax`;
   }
 
   // Resolve CSS width values to pixels, accounting for style overrides from consumers
--- a/apps/apollo-vertex/registry/use-mobile/use-mobile.ts
+++ b/apps/apollo-vertex/registry/use-mobile/use-mobile.ts
@@ -1,19 +1,32 @@
 import * as React from "react";
 
 const MOBILE_BREAKPOINT = 768;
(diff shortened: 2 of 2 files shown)


kittyyueli:
`enable-auto-merge` is red, and it is not this PR's:

```
GraphQL: Resource not accessible by integration (enablePullRequestAutoMerge)
```

The app token the workflow mints cannot call `enablePullRequestAutoMerge`, so `gh pr merge --auto` fails before it looks at the branch at all. Same failure and same message as on #1144, and on #1131 across runs 1120–1132 — it is a repository/app permission gap, not something a PR can fix.

Not re-running it: the error is a deterministic authorization refusal, so a second attempt returns the same thing and would only burn a run. Everything else on this PR stands on its own.

---

pieman1313 (approved):

frankkluijtmans (approved):

PR #1147, opened 2026-09-10
feat(apollo-react): add-guardrail palette [AL-576]

andreizdrali-uipath:
AL-576, the add-guardrail palette: the picker both products open from their guardrails list.
Third member of the shared family in
`packages/apollo-react/src/canvas/components/Guardrails/`.

One commit over #1139, which is one commit over `main`. Review the last commit only.

> The Jira description is stale. It still says apollo-wind and component-scoped locales, both
> superseded. Read it as apollo-react family, strings as lingui ids in the shared canvas catalog.

## What ships

| File | Exports |
| --- | --- |
| `guardrail-palette.tsx` | `GuardrailPalette`, `GuardrailPaletteProps` |
| `guardrail-palette-utils.ts` | `groupGuardrailsForPalette`, `getGuardrailPaletteItemId` |
| `palette-types.ts` | `GuardrailPaletteDefinition`, `GuardrailPaletteGroup` |
| `components/guardrail-palette-item.tsx` | `GuardrailPaletteItem` (not in `index.ts`: the narrow surface is the palette) |
| `i18n.ts` (extended) | `GuardrailPaletteLabels`, `GUARDRAIL_PALETTE_EN_LABELS`, `GUARDRAIL_PALETTE_EN_MESSAGES`, `resolveGuardrailPaletteLabels`, `useGuardrailPaletteLabels` |

Grouped definitions, an optional create-custom entry, and the loading, failed and empty states.
`GuardrailStatusChip` and `GUARDRAIL_CHIP_GEOMETRY` are on #1139, where all three leaves inherit
them from one place.

## Contract

The host filters, the palette offers. Definitions arrive pre-filtered (flags, entitlements,
`FeatureDisabled` / `Disabled`, scope) and both callbacks are intents: the builder that opens
next, the default name it starts with and the telemetry stay with the product. No flag, product
type or telemetry crosses the boundary.

## Design decisions

- **A list of buttons, not a `Select` or a `Command`.** It composes into a dialog, an inline
  overlay or a whole sidebar, and shows descriptions without a popover. Flow's `SelectContent`
  already fights the primitive to get there (multi-line items, disabled items, sticky headers,
  `h-auto min-h-9` trigger, two `!important` viewport overrides). cmdk is a combobox modelling a
  selection, while palette entries are one-shot actions, and its navigation skips
  `aria-disabled` items by construction, which this palette needs to keep reachable.
  **QA-visible for Flow**: the type dropdown becomes a one-step list.
- **One tab stop, roving focus inside it.** Tab enters and leaves the palette instead of walking
  a user through a create-custom entry, six validators and every BYO group. Arrow keys step
  across group boundaries, Home and End jump to the ends, movement clamps rather than wraps.
  The focused entry comes off the keyboard event target, not `document.activeElement`, which
  retargets to the shadow host inside Agents' shadow root; a test renders into an open shadow
  root and arrows down.
- **`Unauthorised` is offered, chipped and `aria-disabled`, never `disabled`**, so a keyboard
  user reaches the chip that says why it cannot be picked. Flow's behaviour. **QA-visible for
  Agents**, where the entry currently opens a builder that then refuses to save.
- **Definitions are generic** over the eight fields the palette reads.
  `EnrichedGuardrailDefinition` and Agents' `OutOfTheBoxGuardrailDefinition` both satisfy it, and
  `onSelectOotb` hands the host's own object back with `parameters` intact, so there is nothing
  to look up again.
- **Only the picker ships.** Flow opens a 500px dialog *and* an inline overlay that renders the
  chosen builder underneath; Agents takes over the sidebar with a back button and its own create
  affordance. Composing either is a handful of lines (`InAHostSidebar` is 24, `InADialog` 37), so
  a wrapper modelling both would be a worse contract than none.
- **`onCreateCustom` is opt-in by presence.** Flow offers custom guardrails for `scope === 'Tool'`
  only and passes the prop conditionally; Agents keeps the affordance in its own header, omits
  the prop and gets no duplicate entry.
- **An empty catalog produces no groups**, unlike both products' `groupGuardrailsForPalette`,
  which returns one empty group. That is what makes the empty state reachable: Flow guards its
  own empty message on `groups.length === 0`, which its implementation can never return.
- **Entry identity is `validator`, or `byoValidatorName:byoGuardrailConnectionId` for BYO.** Flow
  keys BYO entries on the name alone and collides when two connections expose one validator.
- **Chip tones are harvested.** The BYO connector chip is `success`, the green both products
  already give it; the "Preview" chip is `info`, the blue both give that one.
- **Groups name themselves with `aria-labelledby`** pointing at the visible header rather than
  repeating it in `aria-label`, and the loading line is an `<output>` for its implicit
  `role="status"`.
- **Both sorts are `localeCompare()` with no locale argument**, matching both products today.
  Recorded in the JSDoc rather than changed, because changing it is a product decision.

## Strings

Nine `guardrails.palette.*` ids in one builder function, the `definitions-copy.ts` idiom. All
nine are harvested from the products' own catalogs, so the English is what they already ship:
eight from Flow's `addGuardrailPalette_*` keys, `list-aria-label` from Agents'
`guardrails.palette.list_aria_label`.

English only, like the rest of the package: `chore(l10n): sync from Localization` owns the other
thirteen catalogs. `i18n.test.ts` asserts English parity with the source and that no catalog
keeps a `guardrails.palette.*` id the source has dropped.

One item for the loc pass: the two products translate the word itself differently in German,
"Leitplanke" in Flow against "Leitlinien" in Agents.

## Also in this commit

Wind's `AlertDescription` is unconditionally `mt-1`, which assumes an `AlertTitle` above it,
while `alertVariants` pins the icon at `[&>svg]:top-3.5`. With no title the text lands 4px below
the icon, visible in the **Load Failed** story. `mt-0` at the call site, the same hunk as #1140
and #1161 so the branches merge without a conflict. The real fix is one line in wind's
`alertVariants` (`[&>[data-slot=alert-description]:first-child]:mt-0`), which would let all three
workarounds go; separate PR.

## Verification

Guardrails suite 427 passing (23 files), jest-axe on both rendering suites. `tsc --noEmit`,
biome lint and format clean. Stories and tests typechecked separately with a throwaway tsconfig,
since the package tsconfig excludes both and CI therefore does not cover them: no errors in this
PR's files.

## Open questions

1. **The `Unauthorized` chip is amber** (`tone="warning"`). Flow renders it neutral
   (`bg-muted text-muted-foreground`) and Agents does not render it at all. Neutral matches the
   one product that ships it; amber carries more meaning. Reviewer's call.
2. **Should `FeatureDisabled` / `Disabled` definitions also render as disabled entries?** They do
   not today: both products filter them out before rendering, so `Unauthorised` is the only
   non-`Available` status that reaches a correctly filtered palette. Say the word and the
   status-to-chip mapping generalizes.


andreizdrali-uipath:
Published previews, both green:

| PR | apollo-react | apollo-wind |
| --- | --- | --- |
| this PR (#1147) | `6.43.0-pr1147.95dda0a` | `2.48.0-pr1147.95dda0a` |
| preview build (#1148), palette + list | `6.43.0-pr1148.f612ad1` | `2.48.0-pr1148.f612ad1` |

Host sessions want #1148's pair: this PR's preview has no list, and #1140's has no palette.
Note the minor bump from the #1140 pair (`main` moved to 6.43.0 / 2.48.0).

CI: Test Coverage, Typecheck, Lint, Format and Build green. The two red `Audit * Dependencies`
checks are the repo's Dependabot backlog on `main`, identical on #1138 and #1140; no dependency
is added here and `Detect Lockfile Changes` is green.

apetraru-uipath (approved):

apetraru-uipath (approved):

PR #1154, opened 2026-09-11
feat(apollo-vertex): add timeline registry component

KokoMilev:
Moves the activity-timeline marker and row layout out of the invoice-processing client into apollo-vertex as a shared `timeline` registry component, so any vertical solution can pull it. The marker knows only a plain `variant` string (nine values, no invoice concepts) and keeps the client's exact visuals.

Step 2 of 3: step 1 decoupled the marker from the invoice status enum ([invoice-processing#1730](https://github.com/UiPath/vertical-solution-invoice-processing/pull/1730)); step 3 has the invoice client pull this and render its own rows.


apps/apollo-vertex/registry/timeline/timeline-marker.tsx
```diff
@@ -0,0 +1,157 @@
...
+          <Check className="size-3" />
+        </MarkerCircle>
+      ) : (
+        <MarkerCircle
+          compact={false}
+          className="text-white shadow-sm [background-image:var(--ai-gradient-strong)]"
+        >
+          <AiMark size={markSize} variant="solid" />
+        </MarkerCircle>
+      );
```
ruudandriessen, line 155: why use a ternary here but not in the one above?

apps/apollo-vertex/registry/timeline/timeline-row-layout.tsx
```diff
@@ -0,0 +1,37 @@
...
+        ) : isLast ? (
+          <span
+            className="mt-1 min-h-12 w-px flex-1 bg-gradient-to-b from-border to-transparent"
+            aria-hidden="true"
+          />
+        ) : (
+          <span className="my-1 w-px flex-1 bg-border" aria-hidden="true" />
+        )}
+      </div>
+      {children}
```
ruudandriessen: too many nested ternaries, split out into a var above

ruudandriessen (changes requested):

apps/apollo-vertex/registry/timeline/timeline-marker.tsx
KokoMilev: Image
Each case has its own change. 
Some of them change the icon's type based on if it is compact or not, other only change the icons size.

apps/apollo-vertex/registry/timeline/timeline-row-layout.tsx
KokoMilev: I split the nested ternaries.

ruudandriessen (approved):

PR #1156, opened 2026-09-12
feat(apollo-react): swap start point adornment to a lightning bolt [MST-11943]

bryan-uipath:
## What

Replaces the blue-dot start point adornment on `BaseNode` with a lightning bolt, per the design reference on MST-11943.

- Filled bolt = default trigger
- Outline bolt = secondary trigger

Both use lucide `zap` via the existing `CanvasIcon` registry, so the two variants share identical geometry. Apollo's own icon set only ships filled bolts (`Flash`, `Trigger1`), with no outline twin.

## Before

Both rows render the same blue dot. There is no filled/outline distinction to make.

Before: blue dot adornment

## After

Top row filled, bottom row outline, across all three node shapes.

After: lightning bolt adornment, filled and outline

## Not in this PR

- **Positioning.** The bolt stays in the existing bottom-left badge slot. The reference puts it at bottom-center, which needs a new `BaseBadgeSlot` position; splitting that out so the icon can be evaluated on its own first.
- **Data model.** `filled` is a component prop with a `true` default. The execution API only carries the `isExecutionStartPoint` boolean, so distinguishing default from secondary needs a schema addition. The story drives the outline variant through `BaseNodeOverrideConfigProvider`.
- **Single-trigger suppression.** MST-11943 asks that the mark be omitted when a flow has only one trigger. That is a canvas-level decision, since the resolver cannot see sibling nodes.

## Verify

Storybook: `Apollo React Canvas / Components / Nodes / BaseNode / Adornments`. Rows "Default trigger (filled)" and "Secondary trigger (outline)".

### DIFF
--- a/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.stories.tsx
+++ b/packages/apollo-react/src/canvas/components/BaseNode/BaseNode.stories.tsx
@@ -31,6 +31,7 @@ import {
 } from '../../storybook-utils';
 import { DefaultCanvasTranslations } from '../../types';
 import type { ValidationErrorSeverity } from '../../types/validation';
+import { ExecutionStartPointIndicator } from '../../utils/adornment-resolver';
 import { CanvasIcon } from '../../utils/icon-registry';
 import { BaseCanvas } from '../BaseCanvas';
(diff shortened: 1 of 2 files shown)


bryan-uipath:
Will hold off; not a clear cut UX win

PR #1170, opened 2026-09-15
fix(apollo-react): add data-testId to node label [MST-14013]

ayush578:
## Problem

`BaseTextContainer` — the block holding a node's name, description, and the editors that replace them while renaming — carries no attribute a consumer can key off. A canvas host that intercepts double-click on nodes therefore cannot tell "this click belongs to the label" from "this click belongs to the node", and has to reconstruct the block by enumerating its children (`node-label`, `node-sublabel`, `empty-label-placeholder`, the two editors).

That enumeration can't be made correct. The block is deliberately wider than the text it holds — `w-[150%]` on round nodes, the body's text column on rectangles — so a double-click a few pixels off the text lands on the container itself, which matches none of the children. In Maestro Flow that fell through to the node and opened the properties panel mid-rename, discarding the edit (MST-14013).

## Demo

https://github.com/user-attachments/assets/d37f076a-2687-4bb9-b227-01f67864e306

## Change

Stamp `data-node-part="label"` on `BaseTextContainer`, on both shape branches. Purely additive — no behaviour, styling, or DOM structure changes.

Hosts can then recognise the whole label block with one selector instead of enumerating its children, and the match keeps working across the display → edit swap.

### Why `data-node-part` and not `data-testid`

A first version of this used `data-testid`. That was wrong: a test id is a testing hook, and wiring production behaviour to it means a build that strips test attributes silently changes runtime behaviour. Nothing strips them today, but the failure would be invisible if that ever changed.

`data-node-part` is a plain production attribute, in the same family as the state attributes `BaseContainer` already carries (`data-execution-status`, `data-interaction-state`, `data-validation-status`, `data-suggestion-type`, `data-stacked`). The `part` naming leaves room for other structural markers later without adding a new attribute each time.

Deliberately **not** exporting a `NODE_LABEL_SELECTOR` constant alongside it. That would give consumers a compile-time contract, but `apollo-react` exports no selector or attribute-name constants anywhere today, and a one-line fix is the wrong place to introduce a new category of public API.

## Consumer

Maestro Flow's canvas uses it to exclude the label block from its capture-phase double-click handler: UiPath/flow-workbench#4018.


packages/apollo-react/src/canvas/components/BaseNode/NodeLabel.tsx
```diff
@@ -16,10 +16,18 @@ export const BaseTextContainer = ({
   children,
 }: BaseTextContainerProps) => {
   if (shape === 'rectangle') {
-    return <div className="flex flex-1 min-w-0 flex-col items-start text-left">{children}</div>;
+    return (
+      <div
+        data-testid="node-label-container"
```
KodudulaAshishUiPath: You could use `data-slot` instead of `data-testid` to mark this

KodudulaAshishUiPath (approved):

PR #1178, opened 2026-09-15
fix(apollo-vertex): import tw-animate-css from the theme item

kittyyueli:
## Summary

`apollo-vertex-theme` declares `tw-animate-css` as a dependency but never declares the matching `@import`, so a consumer installs the package and no stylesheet ever loads it. Its `css` field is `null`, and no item in the registry declares an `@import` at all.

Tailwind v4 does not ship `animate-in`, `fade-in-0`, `zoom-in-95` or `slide-in-from-*`. **Twelve registry components use them:**

```
alert-dialog   context-menu  dialog    drawer
dropdown-menu  hover-card    menubar   navigation-menu
popover        select        sheet     tooltip
```

Every one of those opens and closes with its transitions silently inert, everywhere Apollo Vertex is installed. Nothing errors — the utilities simply do not exist, so the classes are inert strings.

One line:

```json
"css": { "@import \"tw-animate-css\"": {} },
```

## Why the CLI does not do this for us

It does, but only for fonts. `et()` in `shadcn@4.4.0` synthesizes an `@import` from a **font item's** `dependency`:

```js
const dep = font.dependency ?? `@fontsource-variable/${short}`;
dependencies.push(dep);
css[`@import "${dep}"`] = {};
```

A plain `dependencies` entry has no equivalent path. A CSS-only package has to declare its own import, which is what this PR adds. `tw-animate-css` is the only CSS-only package in the registry — every other dependency is a JS module a component imports directly.

## Verified end to end, with a control

Built items served over `localhost`, installed with the real `shadcn@4.4.0` CLI into a minimal **Vite** project, then compiled with `tailwindcss` v4 against a probe component using the utilities. Run twice — once against this branch, once against `registry.json` before #1167 and this PR — changing nothing else.

| | before | after |
| --- | --- | --- |
| `@import "tw-animate-css"` in `index.css` | absent | **present** |
| `.animate-in` compiles to CSS | MISS | **OK** |
| `.fade-in-0` | MISS | **OK** |
| `.zoom-in-95` | MISS | **OK** |
| `.slide-in-from-top-2` | MISS | **OK** |

Resulting `index.css`, first three lines:

```css
@import "tailwindcss";
@import "tw-animate-css";
@import "@fontsource-variable/inter";
```

## Why CI could not have caught this

Worth knowing, since it is why three defects of the same shape accumulated:

1. **The registry test installs into a Next.js app.** `apollo-vertex-registry-check.yml` runs `shadcn@4.4.0 init --preset a0 --template next`. The non-Next path — what a Vite consumer gets — is never exercised.
2. **The workflow copies the locales in by hand** (`cp -r apps/apollo-vertex/locales .../minimal-app/locales`), a workaround for the `shell` finding below, so the test cannot fail on it.
3. **Nothing asserts the symptoms.** No check mentions `animate-in`, `font-face`, `Inter Variable` or `tw-animate`. `check-css-vars.ts` only validates variables components actually reference, so the malformed `----` keys passed with zero usages.

The A/B above is small enough to become a CI step. Happy to open that separately; out of scope here.

## Registry sweep

Since this is the third defect of the "declared but not wired" shape, I swept the whole registry for the rest of the class:

| Check | Result |
| --- | --- |
| Bare, non-namespaced `registryDependencies` | none — the namespacing fix held |
| `cssVars` keys carrying a `--` prefix | none |
| `registryDependencies` naming an item that does not exist | none |
| Duplicate item names | none |
| Duplicate JSON keys (silent last-wins) | none |
| Declared `files` missing from disk | none — all present |
| Packages imported but undeclared in the item's dependency closure | none |
| `@/registry/<item>` imports without a matching `registryDependency` | none |
| Font item family vs. what its dependency registers | matches |

## Out of scope, needs an owner decision

**1. `shell` ships i18n code without the catalogs it imports.** `lib/i18n.ts` dynamic-imports all fourteen `../locales/*.json`, and `lib/react-i18next.d.ts` does `import type en from "../locales/en.json"`, but the item declares **no** locale files. So `shadcn add @uipath/shell` writes both into a consumer's project with unresolvable imports — a `tsc` failure from the `.d.ts`, and a build/runtime failure from the fourteen dynamic imports.

Two incompatible fixes, and it is not my call which is right:

- Declare the fourteen locale files, so the install is complete — but that writes fourteen catalogs into the consumer's tree.
- Drop `lib/i18n.ts` and `lib/react-i18next.d.ts` from the item, so the shell stops clobbering a consumer's own i18n setup.

At least one downstream consumer already treats the second as the desired behaviour and has stopped re-installing the shell for exactly this reason.

**2. Inconsistent version pins for the same package across items.** Installing two of these together asks for conflicting ranges:

| Package | Specs in the registry |
| --- | --- |
| `@mantine/hooks` | unpinned, `^9.0.0` |
| `@radix-ui/react-tooltip` | unpinned, `^1.2.8` |
| `@tanstack/react-db` | unpinned, `^0.1.86` |
| `@tanstack/react-query` | unpinned, `^5.90.0` |
| `@uipath/uipath-typescript` | `^1.3.1`, `^1.7.1` |

Also worth a look, though not a wiring bug: `font-serif` and `font-mono` name `IBM Plex Serif` / `IBM Plex Mono`, but nothing in the registry provides either face, so both fall through to the generics.

## Testing

| Check | Result |
| --- | --- |
| `registry.json` parses | yes |
| `biome check apps/apollo-vertex/registry.json` | clean |
| `css` field shape | matches what the CLI itself constructs in `et()`; `registryItemCssSchema` is an open record |
| Real install into a Vite consumer, before/after | table above |
| `Apollo Vertex Registry Check` | passing (note: Next-only, so it does not exercise this path — see above) |
| Files changed | `registry.json` only, 1 insertion |
| Lockfile | untouched |

https://claude.ai/code/session_01SnmiT3edq81s2qTDqoE8zP


kittyyueli:
`enable-auto-merge` is red, and it is not this diff.

`Apollo Vertex Auto Merge` fails at `gh pr merge --auto --rebase` with `GraphQL: Resource not accessible by integration (enablePullRequestAutoMerge)` — the Releaser App token lacks the permission, or auto-merge is not enabled on the repo. It is repo-wide, not PR-specific: the same job failed on [#1167](https://github.com/UiPath/apollo-ui/pull/1167) and twice on [#1145](https://github.com/UiPath/apollo-ui/pull/1145), both of which merged normally.

Nothing to port — the fix is an app or repo setting, not a code change. Not re-running either, since a permission error reproduces deterministically.

---

kittyyueli:
## Verified end to end, with a control

CI passing does not demonstrate this fix works (see the next section for why), so I reproduced a real consumer install locally: the built `apollo-vertex-theme` and `font-inter` items served over `localhost`, installed with the actual `shadcn@4.4.0` CLI into a minimal **Vite** project, then compiled with `tailwindcss` v4 against a probe component using the utilities.

Run twice — once against this branch, once against `registry.json` as it stood before #1167 and this PR — changing nothing else.

| | before | after |
| --- | --- | --- |
| `@import "tw-animate-css"` in `index.css` | absent | **present** |
| `.animate-in` compiles to CSS | MISS | **OK** |
| `.fade-in-0` | MISS | **OK** |
| `.zoom-in-95` | MISS | **OK** |
| `.slide-in-from-top-2` | MISS | **OK** |
| `@theme` `--font-sans` | `Inter` | `'Inter Variable', Inter, ui-sans-serif, sans-serif, system-ui` |
| `@font-face` blocks declaring `Inter Variable` | 7 | 10 |
| …referenced from `--font-sans` | **0** | **2** |
| `var(----destructive-fg)` and friends | 3 occurrences | **0** |

The font row is the clearest statement of that bug: before, seven `@font-face` blocks were emitted and **nothing referenced them** — the woff2 files downloaded and went unused.

Generated `index.css` after, first three lines:

```css
@import "tailwindcss";
@import "tw-animate-css";
@import "@fontsource-variable/inter";
```

and the aliases now resolve:

```css
--color-destructive-fg: var(--destructive-fg);   /* was var(----destructive-fg) */
```

## Why CI could not have caught any of this

Not a criticism of the suite so much as a gap worth knowing about, since it is why three defects of the same shape accumulated:

1. **The registry test installs into a Next.js app.** `apollo-vertex-registry-check.yml` runs `shadcn@4.4.0 init --preset a0 --template next`. In `et()`, the Next branch writes `cssVars.theme[variable] = var(--font-sans)` and never reads `font.family`, so the Inter mismatch was invisible by construction. The non-Next path — what a Vite consumer gets — is never exercised.

2. **The workflow copies the locales in by hand:**
   ```yaml
   - name: Copy locales into base app
     run: cp -r apps/apollo-vertex/locales .../minimal-app/locales
   ```
   That is a workaround for the `shell` finding in this PR's description: the item declares no locale files, and the harness supplies them, so the test cannot fail on it.

3. **Nothing asserts the symptoms.** No check in `.github/scripts/test-registry/` or the workflow mentions `animate-in`, `font-face`, `Inter Variable` or `tw-animate`. `check-css-vars.ts` only validates variables that components actually reference, and the three `*-fg` variables had zero usages, so the malformed `----` keys passed. It never checks key shape.

If it would be useful, the A/B above is small enough to become a CI step — install the theme into a Vite app and assert `animate-in` compiles and `--font-sans` matches a declared `@font-face`. Happy to open that separately; it is out of scope here.

---

frankkluijtmans (approved):

ruudandriessen (approved):

PR #1186, opened 2026-09-16
fix(apollo-react): drop disallowed aria-required from ModelPicker trigger

abegu:
## Summary

`ModelPicker`'s trigger (`packages/apollo-react/src/material/components/ap-model-picker/primitives/PickerTrigger.tsx`) renders as an MUI `ButtonBase` — implicit `role="button"` — but was also setting `aria-required="true"` on it. WAI-ARIA does not permit `aria-required` on `role="button"`; it's only valid on form/input-like roles (`textbox`, `combobox`, `listbox`, etc.). axe-core's `aria-allowed-attr` rule (critical impact) flags this.

**Downstream impact:** `@uipath/apollo-react` is consumed by `UiPath/Agents`' `frontend-sw`, whose `ModelPicker` wrapper (`src/components/common/ModelPicker/ModelPicker.tsx`) forwards `required` straight into this component and adds no markup of its own — so this bug is entirely inside `apollo-react`, and no downstream consumer can fix it in their own repo. It's been flagging red on every one of that repo's Layer 3 accessibility CI scans since 2026-08-18 (73+ runs), and affects every other product that consumes this shared `ModelPicker`. Tracked as [PC-5049](https://uipath.atlassian.net/browse/PC-5049).

## Why this fix, not the alternative

The trigger does control an associated `role="listbox"` (via `aria-haspopup="listbox"` + `aria-controls`), so moving `aria-required` onto the listbox instead looked like the "more spec-correct" option at first glance. But the listbox is rendered through an MUI `Popper` with no `keepMounted`, so it's unmounted whenever the popup is closed — which is the trigger's normal resting state. Putting `aria-required` there would mean a screen-reader user tabbing through the form and landing on the (closed) trigger gets no "required" announcement at all — worse than the current bug in practice, even though it would be axe-clean.

Instead this follows an established convention already in this exact package for the identical problem: `canvas/components/Guardrails` hit the same `aria-allowed-attr` violation on `role="group"` (which also doesn't support `aria-required`) and solved it by keeping the visual asterisk `aria-hidden`/decorative and adding a visually-hidden "required" span via `RequiredIndicator`'s `srLabel` prop (`@uipath/apollo-wind`), so the requirement reaches assistive tech through the field's accessible name instead.

`ModelPicker`'s label is a pure-MUI component (no Tailwind, so it can't use `RequiredIndicator` directly — it's explicitly designed to render identically in MUI apps, Angular shells, and bare web-component hosts without any external CSS dependency). This PR applies the same *pattern* natively: the field's `<label>` (already associated with the trigger via `htmlFor`/`id`) keeps its decorative `aria-hidden` asterisk and gains a visually-hidden (clip-based, not `aria-hidden`) "required" span. That text becomes part of the trigger's accessible name via the label association, so screen readers announce required-ness whether the popup is open or closed.

## Changes

- `PickerTrigger.tsx`: no longer renders `aria-required`. The `required` prop stays on the type for source compatibility (it's a primitive exported for teams composing their own pickers) but is documented `@deprecated`/unused.
- `ModelPicker.tsx`: the label's required marker now also renders a visually-hidden "required" span (new i18n string `modelPicker.label.requiredSrLabel`) alongside the existing decorative asterisk.
- `README.md`: updated the accessibility section and prop table to reflect the new behavior.
- `locales/en.json`: added the new source string. (`pnpm i18n:extract` currently fails repo-wide on an unrelated pre-existing conflict — duplicate id `autopilot-chat.error.multiple-files` with two different default messages in `ap-chat` — so the other 13 locale catalogs weren't touched; they'll pick up the new key on the next successful extract + translation pass. Until then, `useSafeLingui`'s fallback renders the English default, same as every other `_({id, message})` call in this file.)
- `ModelPicker.test.tsx`: added two tests — the trigger has no `aria-required` attribute, and its accessible name includes "required" when `required` is set (and doesn't when it isn't).

## Test plan

- [x] `biome lint` / `biome format --check` clean on the touched files
- [x] `tsc --noEmit` shows no new errors (34 pre-existing errors elsewhere in the package, all unrelated — missing/renamed `@uipath/apollo-wind` exports in `canvas/` components, nothing in `ap-model-picker`)
- [x] Full `ap-model-picker` test suite: 107/107 passing, including the 2 new tests
- [ ] Reviewer: confirm in a screen reader (NVDA/VoiceOver) that the trigger announces "Model, required, button" (or equivalent) when `required` is set

[PC-5049]: https://uipath.atlassian.net/browse/PC-5049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


abegu:
Re: the review comment on \`PickerTrigger.tsx:69\` (the retained \`required\` prop being a no-op for direct primitive consumers):

Confirmed via a repo-wide search — \`PickerTrigger\` has zero current internal consumers outside \`ModelPicker.tsx\` itself, so the practical blast radius today is zero. It's a public export of the package though, so the concern about external consumers is legitimate in principle.

I considered giving the primitive its own real accessible-required mechanism (e.g. an optional \`requiredLabel\`/\`aria-describedby\` prop), but that's a bigger API surface decision for this design system than I want to make unilaterally in a bugfix PR — it affects how every future primitive-level consumer is expected to signal required-ness, and I don't have visibility into whether there's already a preferred convention for that at the primitive level (as opposed to the composed-component level, where the \`RequiredIndicator\`/\`srLabel\` pattern already exists). Left it as a documented \`@deprecated\` no-op for now rather than guess at a design; happy to follow up in a separate PR if the team has a preferred shape for it.

Also pushed a fix for the other real finding: the visually-hidden required span used \`sx={{ width: 1, height: 1 }}\` — confirmed via \`@mui/system\`'s \`sizingTransform\` that any numeric value ≤1 there resolves to a percentage (\`1\` → \`'100%'\`), not a pixel value. Changed to explicit \`'1px'\` strings.

PR #1188, opened 2026-09-16
fix(apollo-react): remove hover recolor from the Data Fabric node icon

uipath-utkarsh:
## Summary

The Data Fabric "Read Entity" canvas node recolored its icon's arrow on hover, which reads as an unintended state change on a non-interactive glyph (MST-15298). The hover rule shipped in #915 as an inline `<style>` inside the SVG, and it resolved `var(--foreground-accent)`, so in Maestro's theme the arrow turned blue on hover. This removes the hover recolor entirely, so the whole glyph stays on `currentColor` in every state.

This was never consumer-configurable. `LayersArrowUpRightIcon` only accepts `w`/`h`, and downstream (`@flow/flow-workbench`) just selects it by registry key via `display.icon`, so there was no prop to turn it off.

## Demo
Screenshot 2026-09-16 at 9 18 44 PM

## Changes

- `canvas/icons/LayersArrowUpRightIcon.tsx` — dropped the inline `<style>` rule (`.layers-arrow-up-right-icon:hover .arrow { stroke: var(--foreground-accent) }`) and the now-unused `.arrow` class on the two arrow sub-paths. Updated the doc comment to match.
- `canvas/utils/icon-registry.test.tsx` — the test asserted on `path.arrow`, which existed only as the hover hook. It now asserts the 4 paths render.

The `svg.layers-arrow-up-right-icon` class is intentionally kept, since the registry test and downstream targeting rely on it.

## Testing

- [x] `pnpm lint` passes (scoped to `@uipath/apollo-react`)
- [x] `pnpm typecheck` passes (forced, no cache)
- [x] `pnpm test` passes — 143 files / 2385 tests; `icon-registry.test.tsx` re-verified after rebase (8/8)
- [x] Storybook runs locally on :6007 (`Apollo React/Canvas/Theme/Icons → All Icons`); the served module confirmed to contain 4 plain paths with no `<style>` element and no `.arrow` class. Visual hover confirmation by the reporter still pending.
- [ ] `pnpm test:visual` — not run locally (`apollo-react` has no `test:visual` script); runs in CI

Note: `pnpm audit --prod` reports 24 pre-existing vulnerabilities in `apps/apollo-docs` / `apps/apollo-vertex` nextra transitive deps. This PR changes no `package.json` or lockfile, so none are introduced here.

Resolves MST-15298.


SreedharAvvari (approved):

packages/apollo-react/src/canvas/utils/icon-registry.test.tsx
snuziale, line 63: delete

packages/apollo-react/src/canvas/icons/LayersArrowUpRightIcon.tsx
snuziale, line 15: delete

uipath-utkarsh: Done in #1189. This PR merged before I got to your comments, so the deletions are in a follow-up.

Removed. You were right that it was dead: the class existed only as the selector for the hover rule this PR deleted, and nothing in the repo targets it (no CSS, no other code). I justified keeping it as "downstream targeting" in the description, which was wrong. Only 2 of 53 canvas icons set a className, so it is the exception, not a convention.

packages/apollo-react/src/canvas/utils/icon-registry.test.tsx
uipath-utkarsh: Done in #1189.

Deleted. Worth noting the two comments are coupled: this test asserted via `svg.layers-arrow-up-right-icon`, and that SVG has no `id` to fall back on (unlike `case-management`), so once the class goes there is no selector left to assert on. It does leave the `layers-arrow-up-right` registry entry without coverage, which matches most other entries. Flagged in the follow-up so it is a deliberate call.

PR #1195, opened 2026-09-19
feat(apollo-react): add a medium size to ApBadge

venkateshcherukuri:
## Summary

`ApBadge` has two sizes whose type steps skip the middle of the scale: `small` renders `FontXsSize` (10px) and `large` renders `FontMSize` (14px). There is no way to put a badge next to 12px body text, which is the most common body size in a dense panel, so a consumer either accepts a visible mismatch or overrides the component's internals.

This adds `BadgeSize.MEDIUM`, filling that gap with the next regular step on Apollo's own token ladder.

| size | height | font size | line height |
|---|---|---|---|
| `SMALL` | `SpacingBase` 16px | `FontXsSize` 10px | `FontSLineHeight` 16px |
| **`MEDIUM`** (new) | **`SpacingM` 20px** | **`FontSSize` 12px** | **`FontSLineHeight` 16px** |
| `LARGE` | `SpacingL` 24px | `FontMSize` 14px | `FontMLineHeight` 20px |

Every value is an existing token and each column is the next step in sequence, so nothing new is introduced to the scale.

## Changes

- `BadgeSize` gains `MEDIUM`.
- The size-dependent styles move from chained ternaries to a `SIZE_METRICS` record keyed by `BadgeSize`. With three sizes the ternaries would have nested, and the record makes each size's full metric set readable in one place. Behaviour for `SMALL` and `LARGE` is unchanged, value for value.
- The Sizes story gains a Medium row.

## Compatibility

Additive. `size` still defaults to `SMALL`, and the two existing values produce identical output.

## Motivation

Found while building the Maestro Patterns Library picker, where a row of count badges sits directly above a row of 12px variable-name chips. At `small` the badge text was 10px and read as too faint next to the names; at `large` it was 14px semibold and overpowered them. With no size in between, we fell back to a hand-rolled chip. `MEDIUM` lets that panel use the design system instead.


venkateshcherukuri:
Closing this: it turned out to be unnecessary, and aimed at the wrong package.

`apollo-react/material` is labelled **Maintenance Only** (`apps/storybook/.storybook/main.ts:58`), so adding a size and a prop there is feature work on a surface that is deliberately frozen. I missed that before writing the patch.

More to the point, `apollo-wind`'s `Badge` already does everything we needed, with no Apollo change at all:

- `text-xs` is 12px, the size that was missing from `ApBadge`;
- it takes children rather than a `label` prop, so a leading icon is just a child and needs no `icon` plumbing;
- `inline-flex items-center` composes icon plus text for free;
- `info` and `secondary` variants covered both of our rows.

We have switched the Maestro Patterns Library picker to it and verified it in the product: 22px badges, 12px labels, 16px icons, both rows sharing one component.

Leaving the branch in place in case the `MEDIUM` size is ever wanted for material consumers who cannot move to apollo-wind, but we do not need it.

